PR #369, opened 2015-01-05
Publish schemas to content store

tommyp:
As we want to remove Finder API from the stack we needed to find some other way of publishing Finders to the Content Store. Placing them into Specialist Publisher seemed like the most natural approach so this PR takes the steps required for this to happen. The following is contained within this PR:

* Move the schemas and metadata from Finder API to a `/finders` folder and remove the old schemas
* Change the file paths in the Wiring
* Treat old style single select values as multi select
* Add presenters for Finder Content Items
* Add a `publishing_api_finder_publisher` class 
* Add a rake task to publish the Finders
* Explain new finders and presenters folders in `README`


tommyp:
Amended based on @evilstreak's comments. We now use the file modified time as the timestamp for the content item.

PR #371, opened 2015-01-07
Refactor finder publishing

evilstreak:
Clean up the separation between the `PublishingApiFinderPublisher` and the two presenters it uses to format data.

Also fixes a bug where email signup pages wouldn't be published.

This work was done to prepare for modifying the publisher to also index the pages in Rummager.


tommyp:
LGTM

PR #373, opened 2015-01-13
Update RAIB slug for permission checks

evilstreak:
RAIB was originally created with an incorrect slug in signon. This has since been updated and RAIB editors can no longer access their documents. This commit updates the slug to match the correct version currently in signon.

We should probably republish all RAIB documents after deploying this because they're all associated with a non-existent org

### DIFF
--- a/app/exporters/formatters/raib_report_artefact_formatter.rb
+++ b/app/exporters/formatters/raib_report_artefact_formatter.rb
@@ -15,6 +15,6 @@ class RaibReportArtefactFormatter < AbstractArtefactFormatter
   end
 
   def organisation_slugs
-    ["rail-accidents-investigation-branch"]
+    ["rail-accident-investigation-branch"]
   end
 end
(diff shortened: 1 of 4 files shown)


elliotcm:
Are you missing some?

https://github.com/alphagov/specialist-publisher/search?utf8=%E2%9C%93&q=accidents

tommyp:
@elliotcm It's only RAIB that is singular so I'm merging this.

elliotcm:
:+1:

evilstreak:
Just to follow up to that – MAIB is also singular, and your search showed at least one inconsistency there, so we'll get that fixed but I'm happy for that to be a follow up as none of their stuff is published yet, and they aren't blocked from editing content

jamiecobbett:
Should specialist-publisher be added to these instructions? https://github.com/alphagov/wiki/wiki/Changing-GOV.UK-URLs

evilstreak:
Yes. It's on my todo list, but if someone else gets there first I'd be happy!

PR #380, opened 2015-01-16
State the alert type in email alert subject line

james:
Currently it is not clear to the user whether an alert is relevant to a
drug or a medical device. This commit adds this to the email subject.

### DIFF
--- a/app/exporters/formatters/medical_safety_alert_publication_alert_formatter.rb
+++ b/app/exporters/formatters/medical_safety_alert_publication_alert_formatter.rb
@@ -14,6 +14,16 @@ class MedicalSafetyAlertPublicationAlertFormatter < AbstractDocumentPublicationA
     )
   end
 
+  def subject
+    if document.alert_type == "drugs"
+      "Drug alert: #{document.title}"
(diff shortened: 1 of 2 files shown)


james:
I'm adding yet more hard coded copy in this commit, but specialist publisher doesn't currently seem to provide a way of mapping "drugs" to "Drug Alert". Unless I missed something.

tommyp:
Specialist Publisher has access to the schemas, so you can do it there.

james:
The alternative is:

``` 
alert_type_value = SpecialistPublisherWiring.get(:medical_safety_alert_finder_schema).humanized_facet_value(:alert_type, document.alert_type).first
    "#{alert_type_value}: #{document.title}"
```

not sure which one is better/worse.

james:
I'd vote for the former, in support of my "be more stupid" mantra.

evilstreak:
Hard-coding is definitely clearer to follow.

evilstreak:
We should probably write a migration to follow up on this, stripping the alert type prefix from all of the document titles.

PR #390, opened 2015-01-21
Add change notes to alert emails

james:
When a major change is made to a document, include the change note in the alert email.

Should not effect email alerts for new documents.

Includes MHRA, who have a different email template.

### DIFF
--- a/app/views/email_alerts/medical_safety_alerts/publication.html.erb
+++ b/app/views/email_alerts/medical_safety_alerts/publication.html.erb
@@ -15,6 +15,12 @@
         <%= document_summary %>
       </p>
 
+      <% if document_change_note.present? %>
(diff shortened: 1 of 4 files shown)


app/views/email_alerts/medical_safety_alerts/publication.html.erb
```diff
@@ -15,6 +15,12 @@
         <%= document_summary %>
       </p>
 
+      <% if updated_or_published == "updated" && document_change_note.present? %>
```
evilstreak: It feels like it would be cleaner to just not populate `document_change_note` in the formatter if it shouldn't be shown, rather than putting the logic into the view

james: Hmm, I think I was just being over defensive. I don't believe there is a situation where there would be a document_change_note and we didn't want to see it.

james:
@evilstreak updated.

PR #391, opened 2015-01-21
Catch multiple publishing

james:
If a user is trying to publish a document that has already been
published (e.g. they double click publish, or have the same document
open in multiple tabs), then do nothing.


tommyp:
Does this stop the double "First published" update?

james:
Has when testing on my dev machine. This will probably need some proper testing on preview though.

PR #392, opened 2015-01-21
Reduce duplication in finders

tommyp:
This PR contains the following commits, which are necessary to reduce specificity in Finder Frontend:

- **Specify if a value should be returned as metadata** - This commit allows us to define on a per-facet basis if we want that facet returned as document metadata.
- **Add labels for certain facets** - Certain facets,  (such as `date_issued`) don't just use the name of the Filter for the metadata label. This commit defines it for all existing instances of this.
- **Allow summary definition in metadata** - This commits sets the summaries flag for certain formats and send that to the content store as part of the details hash.

The rest of the work in Finder Frontend to use these changes will come later.

### DIFF
--- a/app/lib/finder_schema.rb
+++ b/app/lib/finder_schema.rb
@@ -20,7 +20,7 @@ class FinderSchema
   end
 
   def humanized_facet_value(facet_key, value, &block)
-    if facet_data_for(facet_key).fetch("type", nil) == "multi-select"
+    if facet_data_for(facet_key).fetch("type", nil) == "text"
       Array(value).map do |v|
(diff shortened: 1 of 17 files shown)


evilstreak:
Added a couple of comments, haven't managed to review it all yet. To be continued…

evilstreak:
:+1: This looks good.

If this is merged and deployed will both finder-frontend and specialist-frontend continue to work without any changes?

evilstreak:
Ie are these changes backwards compatible? If not, are there changes ready for the affected apps which make them backwards compatible?

PR #394, opened 2015-01-22
Migration to remove unwanted draft MAIB reports

evilstreak:
There's no way to delete these unwanted drafts from Specialist Publisher, so here's a migration that does it.

Trello ticket: https://trello.com/c/Jl9tHAyn/520-deleting-items-in-draft-from-prod-maib-finder


evilstreak:
The data in the spreadsheet just changed :(

PR #396, opened 2015-01-26
Make manual sections reorderable

evilstreak:
## Overview

This PR provides editors with a drag-drop interface for reordering the sections of a manual.

reorder-manual-sections

We've added a link to the manual show page to allow reordering, and moved the existing link for adding a section next to it.

screen shot 2015-01-26 at 12 10 03

This is part of the firebreak work for iterating manuals.

## Change history

We don't modify the change history for a manual when the sections are rearranged. Whether the new version of the manual is a major or minor change continues to be determined by whether there are any major changes on any of the sections being published.

Reordering isn't necessarily a substantive change in itself, and where it requires commentary we expect it will be as part of a larger piece of work restructuring whole sections, which will provide a place to add change notes.

## Javascript

Javascript is required for this to work. This is already true in many of our publishing apps, but this is the first feature to require javascript in Specialist Publisher.

Non-javascript versions of this would require a hard-to-use interface (for example, a list of document IDs in a textarea to be cut-and-pasted into the correct order, or an input for each section with an index number) or a lot of intermediate round trips (for example, an up and down button for each section which moves it one position). We decided a javascript-only approach would be appropriate for this MVP.

## Error feedback

The only guard against unexpected behaviour is checking the list of document IDs submitted for reordering against the current list of documents. If the IDs aren't a perfect match (probably because someone has added or removed a section in between the editor opening the reordering form and submitting it) then it will raise an `ArgumentError` which will result in a 500 error.

This should be a rare situation, but it would be good to show a proper validation error to the user instead of a stock 500.


tommyp:
LGTM :+1: 

Especially the PR description.

PR #398, opened 2015-01-27
Only publish Finder if it has a content id

tommyp:
As we work on some Finders, we may not want them to be public. This commit changes the Finder Publisher to only publish a Finder (and it's matching signup page) if it has a content id for the Finder present in the metadata file.


jamiecobbett:
No tests? Also, will the user get any feedback that it hasn't been published?

tommyp:
Chatted with @KushalP offline and doing it now. The User doesn't get any feedback as this all happens in a post deploy hook. I'll update the rake task to say what finder it's published as it happens.

PR #402, opened 2015-01-28
Fix sections in Support for Publishers manual

evilstreak:
Until now it's not been possible for editors to reorder sections in a manual. It's still not possible for editors to remove sections from a manual.

To work around these limitations, editors have been copy-pasting content around, reusing old sections to get the correct ordering. Unfortunately, the slug of a section is never updated after it's been published, so many of the slugs are now very wrong. The version history for these sections is also wrong because it includes content from different sections.

This migration updates the data but does not republish the manual. That needs to be done by an editor as a manual task.

### DIFF
--- a/db/migrate/20150122124937_reslug_some_manual_sections.rb
+++ b/db/migrate/20150122124937_reslug_some_manual_sections.rb
@@ -1,0 +1,252 @@
+# The history
+# ===========
+#
+# Until now it's not been possible for editors to reorder sections in a manual.
+# It's still not possible for editors to remove sections from a manual.
+#
+# To work around these limitations, editors have been copy-pasting content
+# around, reusing old sections to get the correct ordering. Unfortunately, the
+# slug of a section is never updated after it's been published, so many of the
+# slugs are now very wrong. The version history for these sections is also
+# wrong because it includes content from different sections.
+#
+# This migration addresses those issues in the "Support for government
+# publishers" manual.
+#
+# Incidentally, the slug for the manual is now a bit inappropriate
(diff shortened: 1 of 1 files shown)


db/migrate/20150122124937_reslug_some_manual_sections.rb
```diff
@@ -0,0 +1,248 @@
...
+    manual_record = ManualRecord.where(
+      manual_id: manual_id,
+    ).first
+
+    latest_manual_edition = manual_record.editions.last
+
+    current_document_order = latest_manual_edition.document_ids
+    new_document_order = current_document_order
+      .-([support_processes_id, emergency_publishing_id, feedback_id])
+      .+([emergency_publishing_id, content_quality_id, feedback_id])
```
elliotcm: Is this a verbose invocation of the array diff/merge methods, or some Mongoid weirdness?

evilstreak: Yep, diff/merge methods. This is basically `[1, 2, 3, 4] - [2, 3, 4] + [3, 2, 5]`, just making it explicit that `-` and `+` are method calls for easier line wrapping.

elliotcm: Any chance of switching to an explicit set order rather than diffing?  Makes it harder to keep track of what's happening, plus keeping track of the operator precedence.

For instance, if they do bind the way I think they do, the new IDs will be at the end of the list, is that what you're after?

evilstreak: Updated to use an explicit list of IDs

PR #403, opened 2015-01-29
Add facets for CSG

tommyp:
Grant type, Land use, Tiers or standalone items and funding per unit per year. Also adding the inputs to the form and the indexable formatter. [Ticket](https://trello.com/c/gZoWid5B/523-finder-countryside-stewardship-grants).

### DIFF
--- a/app/exporters/formatters/countryside_stewardship_grant_indexable_formatter.rb
+++ b/app/exporters/formatters/countryside_stewardship_grant_indexable_formatter.rb
@@ -7,7 +7,12 @@ class CountrysideStewardshipGrantIndexableFormatter < AbstractSpecialistDocument
 
   private
   def extra_attributes
-    {}
+    {
+      grant_type: entity.grant_type,
+      land_use: entity.land_use,
+      tiers_or_standalone_items: entity.tiers_or_standalone_items,
+      funding_amount: entity.funding_amount,
+    }
   end
 
   def organisation_slugs
--- a/finders/schemas/countryside-stewardship-grants.json
+++ b/finders/schemas/countryside-stewardship-grants.json
@@ -1,5 +1,67 @@
 {
   "slug": "countryside-stewardship-grants",
   "document_noun": "grant",
-  "facets": []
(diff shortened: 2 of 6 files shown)


app/exporters/formatters/countryside_stewardship_grant_indexable_formatter.rb
```diff
@@ -7,7 +7,12 @@ def type
 
   private
   def extra_attributes
-    {}
+    {
+      grant_type: entity.grant_type,
+      land_use: entity.land_use,
+      tiers_or_standalone_items: entity.tiers_or_standalone_items,
+      funding_per_unit_per_year: entity.funding_per_unit_per_year,
```
evilstreak: Could we call this something a bit less of a mouthful, like `funding_amount`? The per unit per year thing feels like a display concern, so something to just put into the label strings.

tommyp: I was thinking that myself but I thought I'd put in the full label for the sake of discussing it. I'll change it now.

finders/schemas/countryside-stewardship-grants.json
```diff
@@ -1,5 +1,67 @@
...
+    },
+    {
+      "key": "tiers_or_standalone_items",
+      "name": "Tiers or standalone items",
+      "type": "multi-select",
+      "preposition": "for",
+      "allowed_values": [
+        {"label": "Higher Tier", "value": "higher-tier"},
+        {"label": "Mid Tier", "value": "mid-tier"},
+        {"label": "Standalone capital items", "value": "standalone-capital-items"}
```
evilstreak: Have we found out whether it's only possible to have certain combinations of these allowed values and the allowed values for grant type? I think that's something Holly was asking about.

evilstreak:
Other than the funding amount key, this is all fine. I'd like to speak to Holly about the tiers/standalone items facet before we merge this though.

jackscotti:
:+1:

PR #408, opened 2015-02-04
Rename slug to base_path

tommyp:
We don't call them slugs when it comes to the content_items. Also this was duplicated in the schemas and wasn't used.


evilstreak:
:+1:

PR #409, opened 2015-02-04
Specify filter mechanism in metadata

tommyp:
Now that we have Finders which may not Filter by one document type, it makes sense to move this into the metadata for each Finder. This is the format for all the existing Finders.

### DIFF
--- a/app/presenters/finder_content_item_presenter.rb
+++ b/app/presenters/finder_content_item_presenter.rb
@@ -45,10 +45,11 @@ private
       document_noun: schema.fetch("document_noun"),
       document_type: metadata.fetch("format"),
       email_signup_enabled: metadata.fetch("signup_enabled", false),
-      format_name: metadata.fetch("format_name"),
+      filter: metadata.fetch("filter", {}),
+      format_name: metadata.fetch("format_name", nil),
       signup_link: metadata.fetch("signup_link", nil),
       show_summaries: metadata.fetch("show_summaries", false),
-      facets: schema.fetch("facets"),
+      facets: schema.fetch("facets", []),
     }
   end
 
--- a/finders/metadata/aaib-reports.json
+++ b/finders/metadata/aaib-reports.json
@@ -6,6 +6,9 @@
   "name": "Air Accidents Investigation Branch reports",
   "beta": true,
   "beta_message": "Until early 2015, the <a href='http://www.aaib.gov.uk/home/index.cfm'>AAIB website</a> is the main source for AAIB reports",
+  "filter": {
+    "filter_document_type": "aaib_report"
+  },
   "signup_content_id": "1ad5918c-19ff-4431-b20b-1bab890cd84b",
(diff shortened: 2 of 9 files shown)


finders/metadata/aaib-reports.json
```diff
@@ -6,6 +6,9 @@
   "name": "Air Accidents Investigation Branch reports",
   "beta": true,
   "beta_message": "Until early 2015, the <a href='http://www.aaib.gov.uk/home/index.cfm'>AAIB website</a> is the main source for AAIB reports",
+  "filter": {
+    "filter_document_type": "aaib_report",
```
evilstreak: Given that [Finder Frontend prefixes filter params with `filter_`](https://github.com/alphagov/finder-frontend/blob/master/lib/finder_frontend.rb#L105), should this just be `document_type`?

tommyp: This is a straight copy of https://github.com/alphagov/finder-frontend/blob/master/lib/finder_frontend.rb#L109. I could change it to be used in the `#filter_params` method though?

evilstreak: Ah, yeah, that makes sense. Go with it like this then!

jamiecobbett:
For Contacts, would we want to specify two filters: the organisation and the document type?

app/presenters/finder_content_item_presenter.rb
```diff
@@ -45,10 +45,11 @@ def details
       document_noun: schema.fetch("document_noun"),
       document_type: metadata.fetch("format"),
       email_signup_enabled: metadata.fetch("signup_enabled", false),
-      format_name: metadata.fetch("format_name"),
+      filter: metadata.fetch("filter", {}),
+      format_name: metadata.fetch("format_name", nil),
       signup_link: metadata.fetch("signup_link", nil),
       show_summaries: metadata.fetch("show_summaries", false),
-      facets: schema.fetch("facets"),
+      facets: schema.fetch("facets", []),
```
jamiecobbett: This change doesn't seem to relate to the commit message.

app/presenters/finder_content_item_presenter.rb
```diff
@@ -45,10 +45,11 @@ def details
       document_noun: schema.fetch("document_noun"),
       document_type: metadata.fetch("format"),
       email_signup_enabled: metadata.fetch("signup_enabled", false),
-      format_name: metadata.fetch("format_name"),
+      filter: metadata.fetch("filter", {}),
+      format_name: metadata.fetch("format_name", nil),
```
jamiecobbett: This change (adding the `nil` fallback) doesn't seem to relate to the commit message.

app/presenters/finder_content_item_presenter.rb
tommyp: Ah yeah. I noticed these were missing some decent defaults so I added them in. Forgot to say why.

PR #410, opened 2015-02-09
Remove format from Finder content item

tommyp:
Now that we have the filter mechanism in the content item, we don't need the format to be present.


jackscotti:
LGTM

PR #413, opened 2015-02-10
Add facets to ESI funds

jackscotti:
fund_state, fund_type, location, funding_source, closing_date

ticket:https://www.pivotaltracker.com/story/show/87188568

### DIFF
--- a/app/exporters/formatters/esi_fund_indexable_formatter.rb
+++ b/app/exporters/formatters/esi_fund_indexable_formatter.rb
@@ -7,6 +7,16 @@ class EsiFundIndexableFormatter < AbstractSpecialistDocumentIndexableFormatter
   end
 
 private
+  def extra_attributes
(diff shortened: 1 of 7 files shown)


app/presenters/finder_content_item_presenter.rb
```diff
@@ -48,6 +48,7 @@ def details
       format_name: metadata.fetch("format_name", nil),
       signup_link: metadata.fetch("signup_link", nil),
       show_summaries: metadata.fetch("show_summaries", false),
+      summary: metadata.fetch("summary", nil),
```
tommyp: Can you do this in a separate commit? Maybe even in a separate PR? It's not related to the ESI funds specifically.

jackscotti: :+1:

PR #423, opened 2015-02-23
Tidy up observer wiring

evilstreak:
A small set of changes to observer registries and the wiring file that cuts the number of lines in the wiring by a bit over one third.

Best reviewed one commit at a time. Perhaps the best person to review this is @tommyp.

### DIFF
--- a/app/observers/aaib_report_observers_registry.rb
+++ b/app/observers/aaib_report_observers_registry.rb
@@ -1,4 +1,3 @@
-require "email_alert_exporter"
 require "formatters/aaib_report_publication_alert_formatter"
 
 class AaibReportObserversRegistry < AbstractSpecialistDocumentObserversRegistry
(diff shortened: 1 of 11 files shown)


jamiecobbett:
I'll leave it to others to review, but it is awesome to see this simplified :+1:

app/observers/aaib_report_observers_registry.rb
```diff
@@ -1,5 +1,7 @@
-require "email_alert_exporter"
 require "formatters/aaib_report_publication_alert_formatter"
+require "formatters/aaib_report_indexable_formatter"
+require "markdown_attachment_processor"
```
tommyp: Since these are in them all, can we move this and the next line to the `AbstractSpecialistDocumentObserversRegistry`?

evilstreak: Good catch. I'll move the require for `rummager_indexer` into the superclass, but I'll leave the require for `markdown_attachment_processer` in the subclasses as that's where the class is referenced.

evilstreak: I've moved the requires and fixed up the original commit.

PR #424, opened 2015-02-23
Include summaries in MAIB finder search results

russellthorn:
Some finders already show the summary field
(eg https://www.gov.uk/international-development-funding)

Marine Accident Investigation Branch (MAIB) has completed work on
their summaries and are now ready to have these show in:
https://www.gov.uk/maib-reports

Finder is configured in a JSON file in Specialist Publisher:
https://github.com/alphagov/specialist-publisher/blob/master/finders/metadata/maib-reports.json
simply include '"show_summaries": true,'
The following rake task `bundle exec rake publishing_api:publish_finders`.
is run automatically on deploy to preview, staging and production.
Which publishes changes made to those files

https://www.pivotaltracker.com/story/show/88531162

Before -
screen shot 2015-02-23 at 14 04 54

After -
screen shot 2015-02-23 at 14 03 43

### DIFF
--- a/finders/metadata/maib-reports.json
+++ b/finders/metadata/maib-reports.json
@@ -10,6 +10,7 @@
   },
   "signup_content_id": "56cb57e2-7e7f-4f67-b2f6-39c9a55385dc",
   "signup_copy": "You'll get an email each time a report is updated or a new report is published.",
+  "show_summaries": true,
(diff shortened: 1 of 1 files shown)


tommyp:
This is fine, but what's with the file being added in the commit?

russellthorn:
Morning @tommyp I have removed that file now, thanks.

PR #429, opened 2015-02-26
Bump content models for CSG

tommyp:
This commit bumps govuk_content_models to 28.5.0 to allow Specialist Publisher to publish countryside stewardship grants.

Also bumping gds-sso to 10.0.0 as this is required by govuk_content_models 28.5.0


evilstreak:
To follow up on a conversation in Slack – I'm pretty sure this is unnecessary. I need to dig a bit more to understand why publishing countryside stewardship grant documents is failing

PR #431, opened 2015-02-26
Add artefact formatters to observers registries

russellthorn:
This commit/pull request adds requires for artefact formatters to enable
publishing in dev environments, this was refactored and works in
tests and works in preview but didn't work locally

I also reordered the requires alphabetically where possible.


evilstreak:
Thanks for sorting this out.

For reference, this is a follow up to #423

PR #439, opened 2015-03-09
Allow DWP and Defra to contribute to ESI finder

jennyd:
- add unit tests for `PermissionChecker`
- add DWP and Defra as owning organisations for the ESI finder

### DIFF
--- a/app/lib/permission_checker.rb
+++ b/app/lib/permission_checker.rb
@@ -34,27 +34,31 @@ private
   end
 
   def user_organisation_owns_format?(format)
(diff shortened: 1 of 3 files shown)


tadast:
The code looks good to me. Perhaps adding some tests with DWP would give more confidence?

evilstreak:
:+1: Looks good.

I wonder whether it would be better to be more explicit about the interface we expect `user` to have (`has_permission?` and `organisation_slug`) by using doubles instead of factories, but that has a tradeoff in that we wouldn't necessarily have tests guarding against the interface of `user` changing. Something to pick up in the future.

PR #442, opened 2015-03-10
Persist changes to CMA case document

mikejustdoit:
It looks like we were only persisting the document during its initial creation, and not after editing its `body` or attachments.

The approach I've taken seems equivalent the other document importers. The only difference is that both valid and non-valid CMA cases are persisted, as described in previous commits (https://github.com/alphagov/specialist-publisher/pull/434).

Part of this ticket: https://www.pivotaltracker.com/story/show/87001102

### DIFF
--- a/app/importers/cma_import.rb
+++ b/app/importers/cma_import.rb
@@ -33,9 +33,9 @@ private
 
   def attachment_attacher
     CmaImportAttachmentAttacher.new(
-      attribute_mapper,
-      cma_cases_repository,
-      data_files_dir,
+      create_document_service: attribute_mapper,
+      document_repository: cma_cases_repository,
+      assets_directory: data_files_dir,
     )
   end
 
--- a/app/importers/cma_import.rb
+++ b/app/importers/cma_import.rb
@@ -32,7 +32,11 @@ private
   end
 
   def attachment_attacher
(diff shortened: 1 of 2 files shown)


app/importers/cma_import/attachment_attacher.rb
```diff
@@ -1,8 +1,9 @@
 class CmaImportAttachmentAttacher
   include ::DocumentImport::AttachmentHelpers
 
-  def initialize(create_document_service, assets_directory)
+  def initialize(create_document_service, repo, assets_directory)
```
evilstreak: It would be good to be explicit about what kind of repo we're expecting here: `document_repository`

app/importers/cma_import.rb
```diff
@@ -32,7 +32,11 @@ def import_job_builder
   end
 
   def attachment_attacher
-    CmaImportAttachmentAttacher.new(attribute_mapper, data_files_dir)
+    CmaImportAttachmentAttacher.new(
+      attribute_mapper,
```
evilstreak: This would probably benefit from being made to use keyword arguments now, but that's not something that should stop us getting this merged

evilstreak:
:+1: Good catch

mikejustdoit:
Thanks @evilstreak. I've made those changes

PR #443, opened 2015-03-10
Import cma cases with missing bodies

mikejustdoit:
Generates a body based on the default body copy "Full text of the decision" and links to the document's attachments.

Suggestions for a better approach than CmaImport#document_creator particularly welcome

### DIFF
--- a/app/importers/cma_import.rb
+++ b/app/importers/cma_import.rb
@@ -1,6 +1,7 @@
 require "builders/specialist_document_builder"
 require "cma_import/mapper"
 require "cma_import/attachment_attacher"
+require "cma_import/missing_body_generator"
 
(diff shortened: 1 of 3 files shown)


mikejustdoit:
Oh, here's a small problem. Of the 345 cases that have no body, 34 have no attachments, so end up with just the body copy and no link. Any ideas?

app/importers/cma_import.rb
```diff
@@ -98,6 +106,10 @@ def parse_json_file(filename)
     })
   end
 
+  def body_defined?(data)
```
evilstreak: This method isn't used any more

evilstreak:
:+1: Nice work

PR #444, opened 2015-03-10
Improve test coverage of authorisation

jennyd:
- test that requests to publish or withdraw specialist documents and manuals are denied, rather than only checking that buttons aren't on the page
- add missing authorisation check when publishing to the `ManualsController`


evilstreak:
:+1: Good work

PR #445, opened 2015-03-11
Add publisher URL to CMA import output

evilstreak:
So that when we provide the output to the CMA for review they can click straight into the right page in Specialist Publisher.


mikejustdoit:
This looks excellent

PR #446, opened 2015-03-13
Amend the text of some of the Countryside Stewardship Finder filters

issyl0:


evilstreak:
This will need a corresponding change to the doctype schema in Rummager: https://github.com/alphagov/rummager/blob/master/config/schema/default/doctypes/countryside_stewardship_grant.json

finders/schemas/countryside-stewardship-grants.json
```diff
@@ -19,16 +20,16 @@
       "preposition": "for",
       "filterable": true,
       "allowed_values": [
-        {"label": "Access", "value": "access"},
```
evilstreak: When removing keys we need to be careful to not break existing documents. In general we should make sure we're making backwards-compatible changes, and removing a key used by a document will break something somewhere (specialist publisher, specialist frontend, finder frontend and rummager all use this data in some way).

I've had a quick look at the data on staging and it doesn't seem like any of the six existing documents is using either of these two keys, so we should be fine to merge this as is.

PR #447, opened 2015-03-17
Contract testing for finders

jamiecobbett:
I had to tweak the output slightly and some of the dummy data in the test to make
it valid. ~~I've made a change in govuk-content-schemas to cover the
additional fields in specialist-publisher finders, hence the build failure. Once this is merged, it should pass: https://github.com/alphagov/govuk-content-schemas/pull/32~~


jamiecobbett:
Tests are now passing

PR #450, opened 2015-03-18
Withdrawing a manual

mikejustdoit:
Service, *Exporter* and /bin-script for withdrawing a published manual.

The script doesn't warn or throw any errors other than `Manual Not Found` (e.g. it won't complain that a manual is "draft" and therefore cannot be withdrawn).

Paired with @evilstreak a load on much of this


app/models/manual.rb
```diff
@@ -66,4 +66,8 @@ def publication_state
   def published?
     state == "published"
   end
+
+  def withdraw!
```
jennyd: I'd be inclined to make this a non-`!` method, for consistency with the `publish` and `update` methods on this model (even though `SpecialistDocument` has `withdraw!`). It's probably also a good idea to return `self` from this method, again for consistency with the two other methods which modify state here.

app/observers/manual_observers_registry.rb
```diff
@@ -19,6 +23,14 @@ def creation
     []
   end
 
+  def withdrawal
+    [
+      publishing_api_withdrawer,
+      rummager_withdrawer,
+      panopticon_exporter,
+    ]
```
jennyd: Should we be creating a `PublicationLog` record when withdrawing? It seems like an important change which would be useful to log.

evilstreak: Logging this would be a good thing to do, but will require changing `PublicationLog` so it can distinguish between publication and withdrawal events. That would be a good time to revisit the responsibilities of that class – it's currently doubling up as an event log and a way to build change history.

If we want to tackle it, I think it would be best done as part of the story but not as part of this PR. The change here provides a useful improvement by itself, as it automates a lot of what 2nd line have to do by hand when withdrawing a manual, so I think we should merge as is.

app/models/manual.rb
evilstreak: :+1:

app/services/withdraw_manual_service.rb
```diff
@@ -0,0 +1,38 @@
...
+  def withdraw
+    manual.withdraw!
+  end
+
+  def persist
+    manual_repository.store(manual)
+  end
+
+  def notify_listeners
+    listeners.each { |l| l.call(manual) }
```
jennyd: It's worth being aware that our Ruby [style guide](https://github.com/alphagov/styleguides/blob/9884028a035afb693e4c9111394e0a9f57a685e0/ruby.md#syntax) says:

> Use single-letter variables with extreme caution. These are sometimes ok in single-line blocks when the line itself explains what the variable is for, but never in multi-line blocks or where the context isn't entirely clear. If in any doubt, give your variables descriptive names.

I'm totally happy with it in this case - it's very clear to me what's happening here.

features/withdrawing-a-manual.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Script to withdraw published manuals
+  As a DevOps specialist
```
jennyd: DevOps!

features/withdrawing-a-manual.feature
```diff
@@ -0,0 +1,10 @@
...
+  I want to withdraw a manual
+  So that it is not accessible to the public
+
+  Scenario:
+    Given I am logged in as a "CMA" editor
+    And a published manual with at least two sections exists
+    When I withdraw the manual
```
jennyd: I think the name of this step is a bit misleading - the manual is actually being withdrawn by the DevOps specialist, not the CMA editor, and users of the app can't do that yet. If we add that functionality later, we may end up accidentally using this step when we intend to test something different, so I think it's worth being more accurate here to avoid future confusion.

app/services/withdraw_manual_service.rb
```diff
@@ -0,0 +1,38 @@
...
+  def initialize(manual_repository:, listeners: [], manual_id:)
+    @manual_repository = manual_repository
+    @listeners = listeners
+    @manual_id = manual_id
+  end
+
+  def call
+    withdraw
+    persist
+    notify_listeners
```
jennyd: If you try to withdraw a draft manual, then its state won't be updated but the listeners will still be called, so a 'gone' content item will be put into the content store and so a 'gone' page will be served instead of a 404.

app/observers/manual_observers_registry.rb
jennyd: OK, let's talk about it and not tackle it as part of this PR then.

mikejustdoit:
Thanks @jennyd, I've addressed those issues

features/support/manual_helpers.rb
```diff
@@ -227,4 +215,70 @@ def check_for_manual_body_preview
...
+
+    manual_services = ManualServiceRegistry.new
+    manual_services.withdraw(manual_id).call
+  end
+
+  def get_id_for_manual(manual_title)
+    visit manuals_path
+    link = page.find_link(manual_title)
+    # TODO this is pretty gross, consider making it less so
+    link.native.attribute("href").value.match(%r{\A/manuals/(.*?)(\?|\z)})[1]
```
jennyd: I agree, this isn't nice, but @evilstreak and I think that it's best tackled as part of a separate story to make all test setup use the services directly, rather than submitting forms to do it. The services would then return objects, which would make it much simpler to make assertions about changes to those objects. Let's write up a separate story to do that, and leave this as it is for now.

PR #451, opened 2015-03-18
Fix already initialized constant

mikejustdoit:
These commits are very small, but basically...

SpecialistPublisherWiring is a regular constant variable dressed as a class name. The way it was being assigned was causing `already initialized constant` warnings which really added a lot of noise to a script I've written in another PR.

This does involve handling Rubocop inline, which seems a bit ugly and it'd be good to hear some thoughts


evilstreak:
Although the Rubocop thing is a bit gross, one of the medium term aims for this codebase is to get rid of the wiring file entirely, so it doesn't seem like a big deal.

PR #454, opened 2015-03-25
Fix incomplete bodies in cma import

evilstreak:
Ticket: https://www.pivotaltracker.com/story/show/90459058

Some of the data that was scraped as part of the work to import legacy CMA cases has `markup_sections` which aren't included in the `body` field.

In some cases the missing content are the links to many of the supporting attachments. That is the case where we are missing the `final-report-and-appendices-glossary` section and no other sections.

This code checks if that's the case, and appends it to the body. It also adds some import notes for the log output to list the missing sections and whether it was fixed automatically.

### DIFF
--- a/app/importers/cma_import.rb
+++ b/app/importers/cma_import.rb
@@ -2,6 +2,7 @@ require "builders/specialist_document_builder"
 require "cma_import/mapper"
 require "cma_import/attachment_attacher"
 require "cma_import/missing_body_generator"
+require "cma_import/body_fixer"
 
 class CmaImport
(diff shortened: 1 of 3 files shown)


jackscotti:
:+1:

mikejustdoit:
Otherwise, :+1: this all looks excellent

evilstreak:
@mikejustdoit I extracted a `down_level_headings` method to make it clearer what it's for, and stopped using a reference to the whole match in the replacement. That change has been rebased into the original commit.

PR #459, opened 2015-03-27
Search slug and title fields, ignore case

fofr:
Make it easier to find documents by:
* automatically searching both title and slug
* making search case insensitive

Also:
* cleans up the search form design
* removes `Reform` gem dependency

https://www.pivotaltracker.com/story/show/90518840

## Before
screen shot 2015-03-27 at 15 33 12

## After
screen shot 2015-03-27 at 15 32 50


evilstreak:
:+1: Looks good.

Since I paired on some of this I'll leave someone else to merge.

app/controllers/abstract_documents_controller.rb
```diff
@@ -121,18 +120,15 @@ def truncate_and_warn(documents)
   end
 
   def searching?
-    search_adapter.term.present?
+    search_query.present?
   end
 
   def search_adapter
```
benilovj: it doesn't feel like `search_adapter` is the right name for this method. Maybe `search_details`?

benilovj: Thinking about it, it's probably fine.

PR #461, opened 2015-03-31
Make attachments collapsible

evilstreak:
Some documents have a lot of attachments. An example of an extreme case at the moment is the "Local bus services market investigation" from the CMA, which has **460 attachments**.

Since the attachments are all displayed on the show page and the edit, publish and withdraw buttons are all right at the bottom of the page, that means editors are having to do way more scrolling than is reasonable.

This is probably best reviewed with whitespace ignored: https://github.com/alphagov/specialist-publisher/pull/461/files?w=1

### Before

screen shot 2015-03-31 at 13 44 07

### After

screen shot 2015-03-31 at 13 50 30

### Document without attachments

screen shot 2015-03-31 at 13 51 42

### DIFF
--- a/app/views/specialist_documents/show.html.erb
+++ b/app/views/specialist_documents/show.html.erb
@@ -48,23 +48,30 @@
 </div>
 <div class="row">
   <div class=" col-md-8">
-    <h2>Attachments</h2>
     <% if document.attachments.any? %>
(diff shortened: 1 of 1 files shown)


fofr:
Closing in favour of #464

PR #462, opened 2015-04-01
Raib finder  - remove betabranding and enable email subscriptions

binaryberry:
RAIB reports (https://www.gov.uk/raib-reports) soon won't be beta any more, and will need email subscription to be turned on, in the same way this happened for the MAIB finder. 

So I removed beta branding in the finder, and enabled subscriptions.

Before:
before raib finder email subscription and beta branding removal

After:
after raib finder email subscription and beta branding removal


evilstreak:
:+1:

PR #464, opened 2015-04-02
Make UI more like other admin apps, prevent accidental publishing

fofr:
* Use Bootstrap form styles rather than custom ones (and extract common form parts into partials)
* Use admin template pattern for attachments table, and truncate large tables with a JS toggle (some documents have over 100 attachments) – instead of https://github.com/alphagov/specialist-publisher/pull/461
* Separate publish, edit and withdraw actions and add confirm dialogues to irreversible actions
  * Completely hide publish button when unavailable, a disabled button is
confusing
* Correctly space sections on the `show` pages with a consistent vertical rhythm
* Use `lead` styles to make summary more readable
* Remove collapsible body function, instead restrict height, and don’t show at all unless there is a body
* Reduce dependencies on `govuk_frontend_toolkit` – currently Bootstrap and the toolkit don’t play well together

https://www.pivotaltracker.com/story/show/90529348

# Editing a document

## Before
screen shot 2015-04-02 at 11 55 02

## After
screen shot 2015-04-02 at 11 55 21

# Viewing a document

## Before
screen shot 2015-04-02 at 11 53 32

Actions:
screen shot 2015-04-02 at 11 53 46

## After
screen shot 2015-04-02 at 11 53 58

Actions:
screen shot 2015-04-02 at 11 54 13

# Truncated attachments table
screen shot 2015-04-02 at 12 09 22

# Warnings
screen shot 2015-04-02 at 11 54 33
screen shot 2015-04-02 at 11 54 22


russellthorn:
This looks good to me, but probably worth having someone else look over it as well.

russellthorn:
Looks good to me. It's probably worth getting a ruby dev to check the ruby, but otherwise I'm happy.

evilstreak:
It looks like you've swapped a collapsing body for a fixed height display. How easy is it to scan through the content when it's shown in a small slice? Are editors looking at that content at all? If they're not, we could probably drop it. If they are, I worry that showing such a narrow slice will make scanning through long documents pretty impossible.

:+1: for the truncated attachments table.

app/views/specialist_documents/_publish_form.html.erb
```diff
@@ -1,14 +1,18 @@
-<div class="panel panel-danger">
-  <% if warning %>
+<% unless disabled %>
```
evilstreak: It seems weird to allow this partial to be included with a parameter that results in it being a no-op. Would it be better to just not include the partial?

evilstreak:
In a couple of commits you hide the publish button altogether when the action is unavailable because "disabled buttons have proved to be confusing in the past". The advantage of showing a disabled button is that it distinguishes between the user not having permissions to publish, and the content item not being a suitable state to be published. Is that a valuable distinction to maintain? If so, is there another way to solve it without using disabled buttons?

fofr:
@evilstreak The fixed body height is a technique used on other publishing tools on textareas when entering body text. If it's not quite right we can iterate. This is preferable to clicking open on an expand link and seeing an empty box.

fofr:
To summarise the hangout I've had with @evilstreak, for this PR we are going to:
* Move the withdraw button to a different place, because its actions are more like those of publish and it's not a safe action
* Re-instate the warning messages for when something can't be published, and check the appropriateness of those messages

For a later PR and after some research:
* Consider either remove the body excerpt altogether, or making it easier to expand and see all of the body copy if it's being used

fofr:
Closing this PR until the first two points are finished.

PR #466, opened 2015-04-10
Make UI more like other admin apps, prevent accidental publishing

fofr:
Supercedes #464 (force-pushed after master rebase so couldn't re-open old PR)

## Differences with #464:
* Move the withdraw button to a different place, because its actions are more like those of publish and it's not a safe action
* Re-instate the warning messages for when something can't be published, and check the appropriateness of those messages

## Original PR
* Use Bootstrap form styles rather than custom ones (and extract common form parts into partials)
* Use admin template pattern for attachments table, and truncate large tables with a JS toggle (some documents have over 100 attachments) – instead of https://github.com/alphagov/specialist-publisher/pull/461
* Separate publish, edit and withdraw actions and add confirm dialogues to irreversible actions
  * Completely hide publish button when unavailable, a disabled button is
confusing
* Correctly space sections on the `show` pages with a consistent vertical rhythm
* Use `lead` styles to make summary more readable
* Remove collapsible body function, instead restrict height, and don’t show at all unless there is a body
* Reduce dependencies on `govuk_frontend_toolkit` – currently Bootstrap and the toolkit don’t play well together

https://www.pivotaltracker.com/story/show/90529348

# Editing a document

## Before
screen shot 2015-04-02 at 11 55 02

## After
screen shot 2015-04-02 at 11 55 21

# Viewing a document

## Before
screen shot 2015-04-02 at 11 53 32

Actions:
screen shot 2015-04-02 at 11 53 46

## After
screen shot 2015-04-02 at 11 53 58

Actions:
screen shot 2015-04-09 at 16 59 57

# Truncated attachments table
screen shot 2015-04-02 at 12 09 22

# Warnings
screen shot 2015-04-02 at 11 54 33


evilstreak:
:+1: This is great

PR #467, opened 2015-04-10
Fix body fixer

mikejustdoit:
Just a couple of small fixes to CmaImport's BodyFixer decorator.

It transpires that this decorator was only originally developed for and tested against a subset of the legacy CMA cases. Without these defaults in place, a few hundred cases outside of that subset complain


boffbowsh:
Looks good to me :+1:

PR #471, opened 2015-04-16
Correct the MAIB organisation slug

jamiecobbett:
This was added in 2d0a48c4d8b5ae9524ad5b40c25d28d4678ed1c4 and seems to have
just been a typo.

~~I don't think just changing this is enough to fix the existing [bad metadata on search results](https://www.gov.uk/search?q=atlantic) - presumably all of the reports need to be resent to Panopticon/Rummager somehow?~~

### DIFF
--- a/app/exporters/formatters/maib_report_indexable_formatter.rb
+++ b/app/exporters/formatters/maib_report_indexable_formatter.rb
@@ -15,6 +15,6 @@ private
   end
 
   def organisation_slugs
-    ["marine-accidents-investigation-branch"]
+    ["marine-accident-investigation-branch"]
   end
 end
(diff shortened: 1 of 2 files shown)


tommyp:
MAIB's slug and name were mismatched when the MAIB finder was created so this was correct at that time. I think in order to correct it they all need to be republished.

jamiecobbett:
After testing in development (and adding support for them in the last commit), I will run:
```
govuk_setenv specialist-publisher bundle exec ruby bin/republish_documents maib_report
``` 
To reindex the documents. Hat tip to @evilstreak for being awesome.

evilstreak:
:+1: As a follow up, we'll move organisation slugs into constants to make this kind of mistake impossible and make managing slug changes a bit less painful

PR #472, opened 2015-04-20
Add script to change document slugs

jamiecobbett:
Based on instructions from the opsmanual: https://github.gds/pages/gds/opsmanual/2nd-line/changing-specialist-publisher-document-slug.html

### DIFF
--- a/bin/document_slug_updater
+++ b/bin/document_slug_updater
@@ -1,0 +1,29 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require "specialist_publisher"
+
+def strip_leading_slash(path)
+  path.sub(%r{\A/}, "")
+end
+
+slug     = strip_leading_slash(ARGV.fetch(0))
+new_slug = strip_leading_slash(ARGV.fetch(1))
+
+edition = SpecialistDocumentEdition.published.where(slug: slug).last
+raise(ArgumentError, "No published document for #{slug}") unless edition
+
+doc_type = edition.document_type
+document_id = edition.document_id
+services = SpecialistPublisher.document_services(doc_type)
+
+puts "Changing #{slug} to #{new_slug}"
+
+puts "Withdrawing #{slug}"
+services.withdraw(document_id).call
+
+puts "Creating a new draft to republish"
+edition.update_attribute(:slug, new_slug)
+
+puts "Republishing as #{new_slug}"
+services.publish(document_id).call


bin/document_slug_updater
```diff
@@ -0,0 +1,33 @@
...
+require './app/lib/specialist_publisher.rb'
+
+def strip_leading_slash(path)
+  path.sub(%r{^/}, '')
+end
+
+slug     = strip_leading_slash(ARGV.fetch(0))
+new_slug = strip_leading_slash(ARGV.fetch(1))
+
+edition = SpecialistDocumentEdition.published.where(slug: slug).last
```
jamiecobbett: We added the `.published` scope as it seemed sensible - it wasn't in the Opsmanual instructions.

bin/document_slug_updater
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require "./app/lib/specialist_publisher.rb"
```
evilstreak: I think this can just be `require "specialist_publisher"` – the path hoop-jumping is only necessary in the Rails console

bin/document_slug_updater
```diff
@@ -0,0 +1,32 @@
...
+
+def strip_leading_slash(path)
+  path.sub(%r{^/}, "")
```
evilstreak: It shouldn't make a difference here, but you probably mean `\A` to match the start of the input rather than `^` which matches the start of any line

bin/document_slug_updater
```diff
@@ -0,0 +1,32 @@
...
+services.withdraw(document_id).call
+
+puts "Creating a new draft to republish"
+edition.update_attribute(:slug, new_slug)
+
+# Without reloading the record, the edition would be left in the "archived" state
+edition.reload
+
+puts "Republishing as #{new_slug}"
+services.publish(document_id).call
```
evilstreak: We shouldn't be using the `publish` service here. I renamed the old `republish` service to `bulk_publish` and slightly modified the behaviour, so the call documented in the ops manual isn't possible any more.

I'll add a proper `republish` service that we can use here.

bin/document_slug_updater
```diff
@@ -0,0 +1,29 @@
...
+puts "Changing #{slug} to #{new_slug}"
+
+puts "Withdrawing #{slug}"
+services.withdraw(document_id).call
+
+puts "Creating a new draft to republish"
+edition.update_attribute(:slug, new_slug)
+
+puts "Republishing as #{new_slug}"
+services.publish(document_id).call
```
jamiecobbett: So after speaking to @evilstreak, `publish` isn't the right thing to do as it will trigger email alerts. `bulk_publish` is also wrong for some reason. `republish` was the right thing to do, but it was removed recently.

dsingleton:
I'm not familiar enough with spec-pub to review or merge this, but is this still in a shape that could be picked up/eventually merged?

jamiecobbett:
This Pull Request automates the instructions in the opsmanual. The problem is that the way they both deal with the problem is to withdraw and then publish the document at the new location. There are some problems with this:
* redirects don't get created - you have to know that you should add one, presumably in router-data
* email alerts are triggered because you are republishing

Doing this properly is non-trivial and not urgent, so I'm closing this.

PR #474, opened 2015-04-23
Add logo to finder metadata

jackscotti:
European Structural and Investment funds finder must comply with EU legislations and they therefore needs to display the EU logo.

This pull request adds the possibility to specify a `logo_path` in the finders' metadata.

Quote from the legislation: "the Union emblem and the reference to the Union shall be visible, when landing on the website, inside the viewing area of a digital device, without requiring a user to scroll down the page".

ticket:
https://www.pivotaltracker.com/story/show/92225774
linked to:
https://github.com/alphagov/finder-frontend/pull/189
https://github.com/alphagov/govuk-content-schemas/pull/49

### DIFF
--- a/spec/lib/publishing_api_finder_publisher_spec.rb
+++ b/spec/lib/publishing_api_finder_publisher_spec.rb
@@ -15,6 +15,7 @@ describe PublishingApiFinderPublisher do
             "content_id" => SecureRandom.uuid,
             "format" => "a_report_format",
             "signup_content_id" => SecureRandom.uuid,
+            "logo_path" => "http://example.com/logo.png",
           },
           timestamp: "2015-01-05T10:45:10.000+00:00",
         },
@@ -25,6 +26,7 @@ describe PublishingApiFinderPublisher do
             "format_name" => "second finder things",
             "content_id" => SecureRandom.uuid,
             "format" => "some_case_format",
+            "logo_path" => "http://example.com/logo.png",
           },
           timestamp: "2015-01-05T10:45:10.000+00:00",
         }
--- a/app/presenters/finder_content_item_presenter.rb
+++ b/app/presenters/finder_content_item_presenter.rb
@@ -44,6 +44,7 @@ private
       document_noun: schema.fetch("document_noun"),
       filter: metadata.fetch("filter", {}),
       format_name: metadata.fetch("format_name", nil),
+      logo_path: metadata.fetch("logo_path", nil),
       signup_link: metadata.fetch("signup_link", nil),
       show_summaries: metadata.fetch("show_summaries", false),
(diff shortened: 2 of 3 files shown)


tommyp:
The tests are failing because you need to update the schema in https://github.com/alphagov/govuk-content-schemas. Give me a shout if you want a hand with it.

jackscotti:
Yep, sorry I forgot to add the 'discuss' tag.

I wanted to get a confirmation on whether this approach to the problem is correct. Can/should I add the image name using another property instead adding a new one (logo)?

tommyp:
Adding a new one seems like the most sensible option since none of the other existing bits in the content item have the same semantic meaning or fulfil even a vaguely similar function.

edds:
I worry that we are referencing an asset in another repository. I would prefer the publisher to just state if it should if this document is part of the EU structural investment fund, then the frontend application can toggle a logo or other description on or off based on that setting. This would allow us to iterate the fontend application without having to update the publisher. It would also give us confidence in removing the asset in future without having to check all the entries in the content store.

Hope that makes sense.

tommyp:
Had a chat with @edds IRL and I agree with him that the best (and simplest) thing to do is to just add this image into the Finder Frontend repo and display the image if the `base_path` matches. Also, will this need to be displayed in Specialist Frontend for the associated documents?

jackscotti:
Okay, thank you for the good feedback!

Anyway, just had a chat with @evilstreak and after going through the EU guidance we decided that the logo is going to be a component. 
The logo will be made of:
- EU logo as an image
- A string

Therefore, `logo` in this PR won't be referring anymore to an asset in Finder Frontend but it will instead be the string that will be rendered in the text part of the component.

jackscotti:
@tommyp @edds, had another chat with @evilstreak, PMs, DM and Mark (Designer) and we decided to not proceed with making it a component because images won't follow a pattern. 

The image will be uploaded through our asset manager and `logo` will indeed specify a path to the asset.

spec/lib/publishing_api_finder_publisher_spec.rb
```diff
@@ -15,6 +15,7 @@
             "content_id" => SecureRandom.uuid,
             "format" => "a_report_format",
             "signup_content_id" => SecureRandom.uuid,
+            "logo_path" => "https://assets.digital.cabinet-office.gov.uk/media/5540ab8aed915d15d8000030/esi_funds_eu_logo.gif",
```
tommyp: I'm not sure this change is needed for this spec.

boffbowsh: Updated with a more generic path. It still needs to be present otherwise it fails validation

PR #475, opened 2015-04-24
Remove email_alert_signup_enabled

tommyp:
As all current Finders that want to show their email_alert_signup page display the link, we can remove the logic for showing and hiding this and do this based on the presence of the email_alert_signup in the links array. In this PR I've removed the line in each metadata file for this and removed the entry in the details hash of the `FinderContentItemPresenter`.


tommyp:
Don't merge before https://github.com/alphagov/specialist-publisher/pull/477 is deployed.

PR #477, opened 2015-04-28
Remove email signup for AAIB

tommyp:
AAIB don't want to have their signup page be public and as we are
removing the ability to hide the link this commit removes all the email
signup information from the AAIB metadata.


evilstreak:
:+1:

PR #478, opened 2015-04-28
Add a script that publishes a single asset via asset-manager

jackscotti:
At the moment there is no easy way to publish assets that aren't attachments
via asset-manager. This script will allow us to do so.

Uploading the European Structural and Investment funds logo is exactly the
case that required us to write this script.
EU logo PT ticket: https://www.pivotaltracker.com/story/show/92225774

All the merit to the fantastic @jennyd!

NB: 
We made this changes in this codebase and not inside asset-manager because it would have been weird to have the asset-manager to send request to itself and to have an API token to access itself.
Ideally though, a script that uploads assets via asset-manager shouldn't be tied with any repository. Therefore, we might consider rewring this script inside asset-manager and making it using the internals instead of having it to send requests to itself.
This is a temporary workaround because at the moment we have no way to easily manage assets on GOV.UK

### DIFF
--- a/bin/create_asset_on_asset-manager
+++ b/bin/create_asset_on_asset-manager
@@ -1,0 +1,31 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+
+filename = ARGV.any? ? ARGV.fetch(0) : nil
+unless filename
+  puts "You need to provide a filename as argument when running this script"
+  abort
+end
+file = File.new(filename)
+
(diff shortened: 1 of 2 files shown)


jackscotti:
Deploy this PR on preview and tested the script, it works!

screen shot 2015-04-28 at 15 17 17

jackscotti:
Screenshot of the script giving a positive outcome:
screen shot 2015-04-28 at 15 20 01

PR #481, opened 2015-05-06
Validate manuals against schema

jamiecobbett:


jamiecobbett:
Closing while I check the effect of some of the changes.

PR #483, opened 2015-05-07
Script and Service to republish manuals

jamiecobbett:
This enables us to make changes to the format of manuals as sent to the
publishing-api and then republish them so that they are consistent.

Successfully tested in development.

Hat tip to @evilstreak.


evilstreak:
:+1:

PR #484, opened 2015-05-07
Fix some historical mangled manual edition states

jamiecobbett:
Apparently, some manuals were unpublished by hand and the edition states were
mistakenly set to "archived" instead of "withdrawn".

SpecialistDocumentEditions use "archived", but manuals use "withdrawn", so
perhaps that was where the confusion came from.

@evilstreak might be able to give a better description and then I can update the commit message.

### DIFF
--- a/db/migrate/20150507155750_change_manual_edition_states_from_archived_to_withdrawn.rb
+++ b/db/migrate/20150507155750_change_manual_edition_states_from_archived_to_withdrawn.rb
@@ -1,0 +1,13 @@
+class ChangeManualEditionStatesFromArchivedToWithdrawn < Mongoid::Migration
+  def self.up
+    ManualRecord.all.each do |manual_record|
+      manual_record.editions.where(state: "archived").each do |edition|
+        edition.update_attribute(:state, "withdrawn")
+      end
+    end
+  end
+
+  def self.down
+    raise IrreversibleMigration
+  end
+end


jamiecobbett:
Rebased with a fix for the rubocop failure.

evilstreak:
Having had a chance to look over the code, I think I may have given you some misleading advice previously. Sorry!

Using `archived` for the state in the database is consistent with what we do for `SpecialistDocumentEdition`. For those, we have some logic to [work out whether the document is withdrawn](https://github.com/alphagov/specialist-publisher/blob/master/app/models/specialist_document.rb#L118-L120) and [expose an appropriate state](https://github.com/alphagov/specialist-publisher/blob/master/app/models/specialist_document.rb#L126-L134).

It would probably make more sense to do something along those lines with manuals as well, rather than updating the data to use `withdrawn` directly. Alternatively, if we need a quick fix for this (which I don't think we do, because we aren't planning to export withdrawn manuals to Panopticon), we could [override `#state_mapping`](https://github.com/alphagov/specialist-publisher/blob/master/app/exporters/formatters/abstract_artefact_formatter.rb#L54-L60) in [`ManualArtefactFormatter`](https://github.com/alphagov/specialist-publisher/blob/master/app/exporters/formatters/manual_artefact_formatter.rb).

evilstreak:
Follow up: since I don't this data is wrong, and since the Panopticon integration will go away soon anyway, I think we can probably just close this PR and do nothing.

jamiecobbett:
Thank you for correction and the detailed explanation. We won't be able to get rid of the Panopticon integration until Topic tagging/collections-publisher/collections (frontend) has moved to the publishing pipeline because the [GDS content manuals are tagged/presented that way](https://www.gov.uk/government-digital-guidance/content-publishing).

chrisroos:
This doesn't appear to have been touched since 11 May 15 so I'm deleting the branch.

PR #485, opened 2015-05-08
Stop sending manual change notes to Panopticon

jamiecobbett:
We want to reduce our dependency on Panopticon, to eventually remove our
dependency on it.

Before this change, we were relying on Panopticon to register the route for the
change notes/updates with router-api. By sending it on the content item,
publishing-api/content-store will do it for us.

There will be some cleanup work to remove the existing change notes from
Panopticon's database and the ManualChangeHistory model from
govuk_content_models and where it is used in govuk_content_api.


evilstreak:
:+1:

PR #486, opened 2015-05-11
Add content ids to manuals

jamiecobbett:
In theory, every content item in the publishing-api/content-store should have a content_id so that it can be linked to. In practice, it isn't required, although the format of it is validated when it is sent.

We will need to republish all manuals after this is deployed, which can be done with `bin/republish_manuals`.

### DIFF
--- a/app/exporters/manual_publishing_api_exporter.rb
+++ b/app/exporters/manual_publishing_api_exporter.rb
@@ -32,6 +32,7 @@ private
 
   def exportable_attributes
     {
+      content_id: manual.id,
       format: "manual",
       title: rendered_manual_attributes.fetch(:title),
       description: rendered_manual_attributes.fetch(:summary),
--- a/app/exporters/manual_section_publishing_api_exporter.rb
+++ b/app/exporters/manual_section_publishing_api_exporter.rb
@@ -22,6 +22,7 @@ private
 
   def exportable_attributes
     {
+      content_id: document.id,
       format: "manual_section",
       title: rendered_document_attributes.fetch(:title),
       description: rendered_document_attributes.fetch(:summary),
--- a/spec/manual_publishing_api_exporter_spec.rb
+++ b/spec/manual_publishing_api_exporter_spec.rb
@@ -26,6 +26,7 @@ describe ManualPublishingAPIExporter do
   let(:manual) {
     double(
       :manual,
+      id: "52ab9439-95c8-4d39-9b83-0a2050a0978b",
       attributes: manual_attributes,
       documents: documents,
     )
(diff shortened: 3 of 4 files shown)


rboulton:
Does each manual section get the same ID as its parent, or different ones?

I can see arguments both ways, and I'm not sure I've seen a written down decision.

app/exporters/manual_publishing_api_exporter.rb
```diff
@@ -32,6 +32,7 @@ def updates_path
 
   def exportable_attributes
     {
+      content_id: rendered_manual_attributes.fetch(:manual_id),
```
evilstreak: This should be `.fetch(:id)`

app/exporters/manual_section_publishing_api_exporter.rb
```diff
@@ -22,6 +22,7 @@ def base_path
 
   def exportable_attributes
     {
+      content_id: document.document_id,
```
evilstreak: This should be `document.id`

spec/manual_publishing_api_exporter_spec.rb
```diff
@@ -61,6 +61,7 @@
 
   let(:manual_attributes) {
     {
+      manual_id: "52ab9439-95c8-4d39-9b83-0a2050a0978b",
```
evilstreak: This should be `id:`

spec/manual_section_publishing_api_exporter_spec.rb
```diff
@@ -46,6 +46,7 @@
   let(:document) {
     double(
       :document,
+      document_id: "c19ffb7d-448c-4cc8-bece-022662ef9611",
```
evilstreak: This should be `id:`

evilstreak:
Several comments which can be summarised by:

The domain objects `Manual` and `SpecialistDocument` both have `id` attributes which are uuids generated when we create them. When being persisted we map those fields to `manual_id` and `document_id` respectively, because Mongoid uses a BSON `id` field to do its business. The only parts of the system which ought to know or care that the fields are called `manual_id` and `document_id` are those parts that deal with persistence. Everything else can happily use `id` and it'll do the right thing.

evilstreak:
Do we need to consider what this will do to content items that have already been published?

Their content ID will change the next time they are published – is anything using those content IDs?

jamiecobbett:
> Do we need to consider what this will do to content items that have already been published?
> Their content ID will change the next time they are published – is anything using those content IDs?

Currently their content_id is nil, and it will become populated. It seems unlikely that there will be a collision as the content_ids are generated by `SecureRandom.uuid`. In as much as collisions might happen, this is already a problem with our architecture.

jamiecobbett:
> Does each manual section get the same ID as its parent, or different ones?
> I can see arguments both ways, and I'm not sure I've seen a written down decision.

@rboulton I think they should have different IDs. Several reasons come to mind:
* they are not equivalent pieces of content (which is one use case we have for sharing content_id)
* I can well imagine wanting to link to individual sections of a manual

rboulton:
Thanks, @jamiecobbett, that makes sense.

jamiecobbett:
Is this good to merge?

PR #487, opened 2015-05-15
Improve republish documents

jamiecobbett:
Equivalent to: https://github.com/alphagov/specialist-publisher/pull/483

Hat tip to @heathd for helping with this.

Successfully tested in development.


app/services/abstract_document_service_registry.rb
```diff
@@ -72,6 +72,20 @@ def publish(document_id)
...
 
+  def republish(document_id)
+    RepublishDocumentService.new(
+      document_repository: document_repository,
+      listeners: observers.republication,
+      document_id: document_id,
+    )
+  end
+
+  def republish_all
```
evilstreak: It seems odd to have this method perform the action when every other public method in this class returns a callable.

app/services/republish_document_service.rb
```diff
@@ -0,0 +1,30 @@
...
+    listeners.each { |l| l.call(document) }
+  end
+
+  def document
+    @document ||= document_repository.fetch(document_id)
+  rescue KeyError => error
+    raise DocumentNotFoundError.new(error)
+  end
+
+  class ManualNotFoundError < StandardError; end
```
evilstreak: Should this be `DocumentNotFoundError`?

jamiecobbett: Fixed and rebased.

app/services/abstract_document_service_registry.rb
jamiecobbett: Good point. We could make it return a callable?
```

evilstreak: Sounds good, though I'd probably use a lambda for consistency with the rest of this code base: `-> { … }`

jamiecobbett: Fixed and rebased.

evilstreak:
:+1:

PR #491, opened 2015-06-03
Add logging to manuals republishing

heathd:
just to give some feedback about the republishing process

### DIFF
--- a/bin/republish_manuals
+++ b/bin/republish_manuals
@@ -2,9 +2,20 @@
 
 require File.expand_path("../../config/environment", __FILE__)
 require "manual_service_registry"
+require "logger"
+
+logger = Logger.new(STDOUT)
+logger.formatter = Logger::Formatter.new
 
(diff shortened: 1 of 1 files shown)


bin/republish_manuals
```diff
@@ -2,9 +2,20 @@
...
 
 repository = SpecialistPublisherWiring.get(:repository_registry).manual_repository
 
-repository.all.each do |manual|
+count = repository.all.count
+
+logger.info "Republishing #{count} manuals..."
+
+repository.all.each.with_index do |manual, i|
+  logger.info("[% 3d/% 3d] id=%s slug=%s" % [i + 1, count, manual.id, manual.slug])
```
jamiecobbett: Simple interpolation would be easier to read I think.

heathd: I wanted to left pad the numbers so that they stayed vertically lined up :-)

PR #497, opened 2015-06-05
Publish a legacy report to the MHRA finder

issyl0:
- This was left behind during the import when they transitioned to
  GOV.UK. Hence, set the publish date to be 2013-07-18.
- Do this using a bin script, Specialist Publisher's equivalent to
  data migrations.

Thanks to @evilstreak for helping with this!

Potentially a next TODO item is to make this script generic so it takes three arguments and can be used for any document that needs publishing at a specific date. Won't do this right now though as the important thing is to get this one document published.

### DIFF
--- a/bin/publish_legacy_report_to_mhra_finder
+++ b/bin/publish_legacy_report_to_mhra_finder
@@ -1,0 +1,15 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require "specialist_publisher"
+
+repository = SpecialistPublisherWiring.get(:repository_registry).for_type(:medical_safety_alert)
+document = repository.fetch("46147971-ed51-4d1a-88f5-f9ae06ef23f1")
+document.update(bulk_published: true)
+document.update(public_updated_at: DateTime.new(2013, 07, 18, 14, 00, 00))
+document.publish!
+repository.store(document)
+
+SpecialistPublisher.document_services("medical_safety_alert").republish("46147971-ed51-4d1a-88f5-f9ae06ef23f1").call
+
+puts "Timestamps updated."


bin/publish_legacy_report_to_mhra_finder
```diff
@@ -0,0 +1,15 @@
...
+repository = SpecialistPublisherWiring.get(:repository_registry).for_type(:medical_safety_alert)
+document = repository.fetch("46147971-ed51-4d1a-88f5-f9ae06ef23f1")
+document.update(bulk_published: true)
+document.update(public_updated_at: DateTime.new(2013, 07, 18, 14, 00, 00))
+document.publish!
+repository.store(document)
+
+SpecialistPublisher.document_services("medical_safety_alert").republish("46147971-ed51-4d1a-88f5-f9ae06ef23f1").call
+
+puts "Timestamps updated."
```
tommyp: If you're going to use this code to make a more generic script, you could change this line to be `puts "#{document.title} bulk published with a timestamp of #{%k:%M %d/%m/%Y}"`

Not a blocker for merging though.

:+1:

jennyd:
@evilstreak said a while ago (documented in the story) that `created_at` on the `PublicationLog` record for the document would also need to be updated, I think because it shows up on the page somewhere. Is that still necessary?

issyl0:
@jennyd: I tried updating that field on `PublicationLog` and it couldn't find anything for the document in question. Then, according to mine and @evilstreak's pairing on this, the instructions in the story were wrong as `PublicationLog` is only for another type of document (I can't immediately remember which one).

tommyp:
`PublicationLog` is used for Manuals. Change notes for Specialist Documents are saved directly to the Document.

jennyd:
Ah ok :+1:

bin/publish_legacy_report_to_mhra_finder
issyl0: I don't know how often this type of thing has to happen, but if it has to happen in future then I'll (or someone else could) make it generic with this change, as it's apparently not a blocker for merging. Thanks!

PR #500, opened 2015-06-16
Feature helpers that dont use the UI

mikejustdoit:
Helpers for use in Cucumber features that:

- create a manual,
- create a manual section and
- publish a manual

...without logging in as a user and using the UI.

For better or for worse there are features that test services and/or scripts,
and having to log in and use the UI to satisfy prerequisites like,

  `'Given a published manual exists'`

doesn't seem to fit these features' requirements.

As part of this work, we've moved the Withdrawing a Manual feature to these (methods.

(Paired with @binaryberry - thanks!)

### DIFF
--- a/app/services/create_manual_document_service.rb
+++ b/app/services/create_manual_document_service.rb
@@ -33,6 +33,6 @@ class CreateManualDocumentService
   end
 
   def document_params
-    context.params.fetch(:document)
+    context.params.fetch("document")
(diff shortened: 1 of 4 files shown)


boffbowsh:
This makes sense overall, ordinarily I'd say it needs more use of Factory Girl, but I don't think that's really relevant to this repo.

PR #502, opened 2015-06-17
Rename file to prevent exception during db:setup

robmckinnon:
* Change file name to match class name to fix this loading error during `rake db:setup`:
```
Failed to determine model from app/models/validators/vehicle_recalls_and_faults_validator.rb:
            NameError:uninitialized constant VehicleRecallsAndFaultsValidator
```


jennyd:
Thanks for fixing this @robmckinnon!

PR #503, opened 2015-06-22
Removing a manual section

mikejustdoit:
I'd suggest reviewing commit-by-commit - I think we've managed to keep the commits in this pull request small and where necessary documented our design decisions in the commit messages.

Overview of changes:
1. Allow a Manual to 'remove' a draft section by moving it from its `#documents` to its `#removed_documents`.
2. Add a `RemoveManualSectionService` to handle the details of that.
2. Add a /bin script to use the service.
3. Make changes to the `ManualSectionServiceRegistry` to better accommodate the script's needs.

There is some terrible Cucumber step naming here, to avoid collisions with existing steps that weren't implemented in a way that we could make use of. (Some steps end with "without the UI"). Fixing this would require continuing the work started in [pull request 500](https://github.com/alphagov/specialist-publisher/pull/500).

(Again, thanks @binaryberry for some great pairing!)


jackscotti:
LGTM but might want to get a 2nd review by a pair of eyes a bit more experienced than mine!

boffbowsh:
This all *looks* ok from a dumb Ruby reviewer point of view, but I have very little context about the app.

evilstreak:
:+1: Good work.

I'd like to talk about the service registries this adds and changes a little bit. I think we can find a cleaner solution without much work.

mikejustdoit:
@jackscotti and @evilstreak, points taken and issues addressed

PR #504, opened 2015-06-23
Order ESI Funds calls in reverse chronological order of closing date

issyl0:
- Users had to scroll down past recently updated closed funds in order to see the open cases unless they used the filters. Now the results are in reverse chronological order of closing date, so the open ones are at the top.
- Add the default_order attribute to the list that is sent to finder-frontend, else the default of "-public_timestamp" would always be used even if we set a different one here (https://github.com/alphagov/finder-frontend/blob/master/app/lib/search_query_builder.rb#L73).
- Make sure it's not passed if it's nil, so that GOV.UK Content Schemas don't complain (they always expect a string - https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/finder/publisher/schema.json#L123).

This requires an update to Rummager's `ALLOWED_SORT_FIELDS`, which I'm about to open a PR for.

Paired with @mikejustdoit on this. Thanks also to @evilstreak for his initial help.

### DIFF
--- a/app/presenters/finder_content_item_presenter.rb
+++ b/app/presenters/finder_content_item_presenter.rb
@@ -49,7 +49,8 @@ private
       show_summaries: metadata.fetch("show_summaries", false),
       summary: metadata.fetch("summary", nil),
       facets: schema.fetch("facets", []),
-    }
+      default_order: metadata.fetch("default_order", nil),
+    }.reject {|_, value| value.nil?}
   end
 
   def format
--- a/finders/metadata/esi-funds.json
+++ b/finders/metadata/esi-funds.json
@@ -9,5 +9,6 @@
   "filter": {
     "document_type": "european_structural_investment_fund"
   },
+  "default_order": "-closing_date",
   "organisations": []
 }


app/presenters/finder_content_item_presenter.rb
```diff
@@ -50,6 +50,10 @@ def details
       summary: metadata.fetch("summary", nil),
       facets: schema.fetch("facets", []),
     }
+    unless metadata["default_order"].nil?
+      details_hash[:default_order] = metadata["default_order"]
```
boffbowsh:
```ruby
details_hash[:default_order] = metadata["default_order"] if metadata["default_order"]
```
would be cleaner

issyl0: That takes the line to 89 characters, not our suggested 80. Does that matter in this case?

mikejustdoit: It matters as much as it does anywhere else.

If the objection is to the use of `unless metadata["default_order"].nil?`, sure it's verbose and could be changed to `if metadata["default_order"]`, but it read clearly already.

boffbowsh: Ah sorry, didn't realise. I'm too used to 100 still :/

mikejustdoit: If anything'd be preferable, for me it'd be a `compact`-style `reject` to only have details with non-`nil` values returned.

I don't know https://github.com/alphagov/finder-frontend well enough to know if that'd be a good move, though.

app/presenters/finder_content_item_presenter.rb
```diff
@@ -50,6 +50,8 @@ def details
       summary: metadata.fetch("summary", nil),
       facets: schema.fetch("facets", []),
     }
+    details_hash[:default_order] = metadata["default_order"] if metadata["default_order"]
```
rboulton: I'd be tempted to put this into the main body of the hash (ie, as `default_order: metadata["default_order"]`) and then filter out anything with a nil value in the hash by calling something like `reject {|_, value| value.nil?}` on the hash.

If you have more than one such optional value, this is going to get hard to read quickly.

boffbowsh: I like this idea :+1:

issyl0: Thanks. I'll look into doing it like that. Quick question: what does the underscore mean in `|_, value|`?

boffbowsh: Basically disregards it. http://po-ru.com/diary/rubys-magic-underscore/

issyl0: I just remembered something we said yesterday when Mike thought about filtering out all nil values. A lot of other things are explicitly set to nil in that hash, and we were wondering if filtering them out would change behaviour of other things in other apps that rely or handle those values being nil? See https://github.com/alphagov/specialist-publisher/pull/504#discussion_r33061850.

issyl0: That appears to do the right thing for our purposes here. Everything I can see still works.

issyl0:
I think I've addressed all the comments here.

PR #507, opened 2015-06-26
Add script for republishing withdrawn documents

evilstreak:
Restores documents to a published state and republishes them silently, ie without triggering email alerts.


mikejustdoit:
:+1:

PR #508, opened 2015-06-30
Add script to update publication timestamp

binaryberry:
RAIB asked to update the publication date of their RAIB reports, because
the date they were originally published precedes the date they were
published on GOV.UK.

I created a bin script that allows to update publication timestamp
of any document by specifying document type and content_id.

The Trello story mentions updating the PublicationLog but that is not 
relevant here, as the PublicationLog is for Manuals only.

cc @benilovj 

Trello story: https://trello.com/c/A8h4UxSj/1-correct-publication-date-on-2-raib-investigation-reports-medium

### DIFF
--- a/bin/update_timestamp
+++ b/bin/update_timestamp
@@ -1,0 +1,50 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require "specialist_publisher"
+
+def usage
+  $stderr.puts %Q{
+USAGE: #{File.basename(__FILE__)} (format ...)(document_id) (date)
+
+1. available formats:
+
+  #{SpecialistPublisher.document_types.sort.join("\n  ")}
+
+2. document_id:
+  To find document_id, login to specialist-publisher in production, go to document page and copy document_id from url.
+
+3. date:
+  Format: yyyy/mm/dd or to specify publishing time: yyyy/mm/dd hh:mm
+  }
+
+  exit(1)
+end
+
+usage unless ARGV.any?
+document_type = ARGV[0]
+document_id = ARGV[1]
+
+begin
+  date = DateTime.parse(ARGV[2])
+rescue ArgumentError
+  puts "date format invalid, format should be: yyyy/mm/dd or to specify publishing time: yyyy/mm/dd hh:mm"
+  usage
+end
+
+unless SpecialistPublisher.document_types.include?(document_type)
+  raise ArgumentError, "please select document type among available formats provided:\n\n  #{SpecialistPublisher.document_types.sort.join("\n  ")}"
+end
+unless document_id.length > 30 && document_id.length < 40
+  raise ArgumentError, "document_id invalid"
+end
+
+repository = SpecialistPublisherWiring.get(:repository_registry).for_type(document_type.to_sym)
+document = repository.fetch(document_id)
+document.update(public_updated_at: date)
+document.publish!
+repository.store(document)
+
+SpecialistPublisher.document_services(document_type).republish(document_id).call
+
+puts "#{document_type} item timestamp updated to #{date}."


bin/update_timestamp
```diff
@@ -0,0 +1,41 @@
...
+def update
+  $stderr.puts %Q{
+USAGE: #{File.basename(__FILE__)} (all|[format ...])(content_id) (date)
+
+1. available formats:
+
+  #{SpecialistPublisher.document_types.sort.join("\n  ")}
+
+2. content_id:
+	To find content_id, login to specialist-publisher	in production, go to content item page and copy content_id from url.
```
mikejustdoit: It's probably best to replace any tabs with spaces

mikejustdoit: I'm not sure if this is the right place for the advice on how to find the `content_id`. It might be, but previously that kind of information's been written up in the Ops Manual's Specialist Publisher section (https://github.gds/pages/gds/opsmanual/2nd-line/applications/specialist-publisher.html)

bin/update_timestamp
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require "specialist_publisher"
+
+def update
```
mikejustdoit: Should this be called something like `usage`?

bin/update_timestamp
```diff
@@ -0,0 +1,41 @@
...
+
+  exit(1)
+end
+
+update unless ARGV.any?
+document_type = ARGV[0]
+content_id = ARGV[1]
+date = DateTime.parse(ARGV[2])
+
+raise ArgumentError, "please select document type among available formats provided" unless SpecialistPublisher.document_types.include?(document_type)
```
mikejustdoit: Some of these `ArgumentError` lines're quite long. Maybe it'd be better to break them over multiple lines like

bin/update_timestamp
```diff
@@ -0,0 +1,41 @@
...
+raise ArgumentError, "content_id invalid, see point 2. to retrieve it" unless content_id.length > 30 && content_id.length < 40
+raise ArgumentError, "date format invalid, see point 3 for correct format" unless date.is_a?(DateTime)
```
mikejustdoit: If there's an `ArgumentError` raised, are the "point 2" and "point 3" actually displayed for the user to refer to?

bin/update_timestamp
```diff
@@ -0,0 +1,41 @@
...
+
+repository = SpecialistPublisherWiring.get(:repository_registry).for_type(document_type.to_sym)
+document = repository.fetch(content_id.to_s)
```
mikejustdoit: If the `content_id` needs `#to_s`ing, maybe it'd be best to do it just once when it's pulled out of `ARGV`? It might be that it doesn't need it anyway?

bin/update_timestamp
tommyp: This isn't the `content_id` of the Document, as Specialist Documents don't live in the Content Store yet. This is better described as the `document_id`.

bin/update_timestamp
```diff
@@ -0,0 +1,41 @@
...
+  exit(1)
+end
+
+update unless ARGV.any?
+document_type = ARGV[0]
+content_id = ARGV[1]
+date = DateTime.parse(ARGV[2])
+
+raise ArgumentError, "please select document type among available formats provided" unless SpecialistPublisher.document_types.include?(document_type)
+raise ArgumentError, "content_id invalid, see point 2. to retrieve it" unless content_id.length > 30 && content_id.length < 40
```
tommyp: Same point as above about `content_id`.

bin/update_timestamp
binaryberry: Thanks both! Updated now with spaces and document_id. 
Mike, regarding finding the document id: once I've written up the Ops Manual section for running this script, maybe I can delete the explanation and tell people to refer to it?

bin/update_timestamp
binaryberry: yes, it very much should have. Corrected now!

bin/update_timestamp
binaryberry: Done now, not the prettiest code in the world but lines are a more manageable length.

bin/update_timestamp
binaryberry: changed error message to stop referring to point 2 and 3 - also changed the way Date error was raised, as my error message wasn't being displayed because the DateTime parsing error was happening before.

bin/update_timestamp
binaryberry: changed to document_id, thanks!

binaryberry:
@mikejustdoit are you happy with my updates? Happy to discuss anything / Hangout if necessary.

mikejustdoit:
Good work :+1:

PR #510, opened 2015-07-01
GDS Editors should have permission to access all manuals

mikejustdoit:
Access is normally limited to Manuals owned by the current user's organisation. So that GDS Editors can support other departmental content editors, without having to change their own organisation temporarily in Signon, we treat a GDS Editor as a special case with access to all Manuals.

These changes do not allow a GDS Editor to create a Manual on behalf of another organisation, only to access existing Manuals.

The test coverage for this feature is minimal but since it mirrors the existing access control tests for standard Editors, it's presumably sufficient.

It might make more sense to abstract the explicit use of `current_user_is_gds_editor?` here into a new or existing dedicated object/config for permissions logic. Suggestions would be very welcome.

I think the changes made to the various ServiceRegistries here really draw attention to the increasing duplication and need for refactoring. Again, suggestions would be excellent there as well.

### DIFF
--- a/app/controllers/manual_documents_controller.rb
+++ b/app/controllers/manual_documents_controller.rb
@@ -94,6 +94,18 @@ class ManualDocumentsController < ApplicationController
 
 private
   def services
+    if current_user_is_gds_editor?
+      gds_editor_services
+    else
+      organisational_services
+    end
+  end
+
+  def gds_editor_services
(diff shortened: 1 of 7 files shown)


mikejustdoit:
I think I opened this too soon. I'm gonna close and have a rethink

PR #511, opened 2015-07-02
Allow GDS Editors to see and edit other organisations' manuals

mikejustdoit:
Access is normally limited to Manuals owned by the current user's organisation. So that GDS Editors can support other departmental content editors, without having to change their own organisation temporarily in Signon, we treat a GDS Editor as a special case with access to all Manuals.

These changes do not allow a GDS Editor to create a Manual on behalf of another organisation, only to access existing Manuals and to create GDS Manuals.

It might make more sense to abstract the explicit use of current_user_is_gds_editor? here into a new or existing dedicated object/config for permissions logic. Suggestions would be very welcome.

I think the changes made to the various ServiceRegistries here really draw attention to the increasing duplication and need for refactoring. Again, suggestions would be excellent there as well.


features/creating-a-manual-as-a-gds-editor.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Creating and editing manuals
+  As a GDS editor
+  I want to create and edit manuals of my own and other organisations
```
jennyd: This isn't quite accurate: GDS editors can't create new manuals which belong to other organisations, they can only view and edit existing ones.

jamiecobbett:
Maybe I'm picking nits, but the PR title keeps confusing me - would you mind rewording it to be proper English and stuff?

features/support/manual_helpers.rb
```diff
@@ -333,4 +335,22 @@ def check_manual_is_withdrawn_from_panopticon(manual_slug)
         )
     end
   end
+
+  def get_id_for_manual(manual_title)
+    visit manuals_path
+    link = page.find_link(manual_title)
+    # TODO this is pretty gross, consider making it less so
+    link.native.attribute("href").value.match(%r{\A/manuals/(.*?)(\?|\z)})[1]
```
jennyd: We agree that this is pretty gross ([as before](https://github.com/alphagov/specialist-publisher/pull/450#discussion_r26955293)) but can't see another way of doing this right now. We think that showing the organisation's title on the manual page is a good idea, not just to make testing this easier but also to make the ownership clear to GDS editors, and to help with debugging potential issues where users think they should be able to see a manual but can't. Adding the organisation to the page is outside the scope of this PR, but this test can be simplified when we do that.

mikejustdoit:
@jennyd and @issyl0, I've made the changes you suggested, thanks.

I've also refactored the Manual and ManualDocument service registries to remove the duplication that was upsetting everyone. (And tried to explain my design decisions in those commit messages.)

jennyd:
Thanks for dealing with those comments and other things we discussed. Excellent commit messages too :+1:

PR #513, opened 2015-07-07
Gracefully handle manuals that can't be found

jamiecobbett:
Pattern copied from app/repositories/specialist_document_repository.rb and
app/controllers/abstract_documents_controller.rb.

If you navigate to the path for a manual that doesn't exist in the database, you
currently see a generic Rails error page. This happens routinely when a manual
is created in our Preview environment, gets lost when the production => preview
data sync happens and then a user tries to access it the following day. By
raising a more specific error, we can then catch it and handle it gracefully.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,6 +11,10 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :exception
 
+  rescue_from("ManualRepository::NotFoundError") do
+    redirect_to(manuals_path, flash: { error: "Manual not found" })
+  end
+
   def current_finder
     finders.fetch(request.path.split("/")[1], nil)
   end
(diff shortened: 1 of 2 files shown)


mikejustdoit:
This makes sense.

We could also display something more specific for "Manual not found"?

jamiecobbett:
> We could also display something more specific for "Manual not found"?

I've corrected the copy pasta of "Document not found" to "Manual not found". Is that what you meant?

edds:
Out of curiosity how would a user on our production-like stacks be able to navigate to a manual that doesn't exist? Are we registering a prefix path for them rather than registering them individually with the router? Or does this handle if you are trying to navigate to a page in the manual that doesn't exist?

bradleywright:
@edds this is for the publisher, not the frontend.

edds:
@bradleywright well that would certainly explain that. As you all were, never mind me.

jamiecobbett:
Bump.

PR #518, opened 2015-07-08
Tweak update timestamp script

jamiecobbett:


mikejustdoit:
Thanks, @jamiecobbett. Well spotted.

mikejustdoit:
Another time it'd be excellent to talk about designing nicer interfaces for the tasks this (and some other scripts in this repository) are performing but having to work around/outside of existing service objects to do.

PR #521, opened 2015-07-16
Show Org for each manual

tommyp:
Since GDS can now see Manuals from all Orgs we should have some way of displaying which Org each one is from. This commit modifies `show` and `index` to say which Organisation owns that Manual. This is only visible to GDS editors.

This PR also modifies a test which checked that when a GDS editor modified a manual created by a user from another Org, it kept the previous Orgs slug. This test now visits the Manual page and checks that the slug is rendered on the show page. [Ticket](https://trello.com/c/Vdgu4gLb/48-specialist-publisher-should-show-which-organisation-a-manual-belongs-to).

Screenshots:
## `manuals/show.html.erb`:
screen shot 2015-07-16 at 17 09 23

## `manuals/index.html.erb`:
screen shot 2015-07-17 at 10 45 23


features/support/manual_helpers.rb
```diff
@@ -346,11 +346,8 @@ def get_id_for_manual(manual_title)
   def check_manual_has_organisation_slug(attributes, organisation_slug)
     # there's currently no mention of the manual's organisation on its page
```
mikejustdoit: This comment can go now! :dancers:

features/support/manual_helpers.rb
```diff
@@ -346,11 +346,8 @@ def get_id_for_manual(manual_title)
...
     manual_id = get_id_for_manual(attributes.fetch(:title))
-    manual_services = ManualServiceRegistry.new
-
-    manual, _ = manual_services.show(manual_id).call
+    visit manual_path(manual_id)
```
mikejustdoit: I think we might be able to just use `go_to_manual_page(attributes.fetch(:title))` now... And remove the method `get_id_for_manual` while we're at it, too, cause this is the only remaining use of it

tommyp: Good call. On this and the comment. Fixed up the commits.

tommyp:
Updated the design in `index.html.erb` based on design input.

config/initializers/content_domain.rb
```diff
@@ -7,3 +7,11 @@
...
                        "https://www.gov.uk"
                      end
+
+ORGANISATION_URL = if domain.include?  "dev"
+                     "http://whitehall-admin.dev.gov.uk"
+                   elsif domain.include? "preview"
+                     "http://www.preview.alphagov.co.uk"
+                   else
+                     "https://www.gov.uk"
+                   end
```
jennyd: This logic seems more complex than it should need to be. `Plek.current.website_root` uses the `GOVUK_WEBSITE_ROOT` envvar, which is set from Puppet:

- development: http://www.dev.gov.uk
- preview: https://www.preview.alphagov.co.uk
- staging: https://www-origin.production.alphagov.co.uk
- production: https://www.gov.uk

It may be preferable to override that in dev, where it's useful to use the frontend app rather than assume we have the router running by using `http://www.dev.gov.uk`. The logic here uses `http` for preview, and points to production from staging, which could lead to confusion when deploying.

Although this is following the pattern of `MANUAL_CONTENT_URL`, could we simplify both here?

tommyp: So we're happy to remove this file and just call `Plek.current.website_root` from the helpers?

jennyd:
:+1: :snail: :no_entry_sign: :shell:

PR #522, opened 2015-07-20
Update README to describe converting finder schemas

heathd:
The process for adding a new specialist document format should also
include updating govuk-content-schemas appropriately to include the new
format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,12 +75,6 @@ $ bundle exec rake
 2. Add missing field definitions in `config/schema/field_definitions.json`.
 3. Add the new document type in `config/schema/indexes/mainstream.json`.
 
-### [govuk_content_models](https://github.com/alphagov/govuk_content_models)
-
-1. Add the new artefact type in `app/models/artefact.rb`.
-2. Bump the gem version.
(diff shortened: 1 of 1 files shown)


tommyp:
Can we add a section to the `README.md` about how to change a format with these changes? It should also probably specify that changing the `value` for any of the facet options is a bad idea because:

1. Links to pre-filtered Finders won't have that option selected
2. The old value won't be an option in the facet which means;
3. Any documents with the old `value` won't be filterable on that `value`
4. Any new documents published will have the new `value`, which will be different from the old one

The last 2 issues are fixable by republishing all the documents, but this isn't advisable as some of the specialist document formats have more than 10,000 documents.

Changing the label is fine though. :+1:

heathd:
@tommyp I agree that some step-by-step instructions about modifying existing finders would be valuable but I think it's outside the scope of this pull request.

I'd be happy to sit with you and draft something (for a separate PR).

danielroseman:
We also need to remove the section about govuk-content-models from this doc; there's no longer any need to do anything there.

heathd:
Thanks @danielroseman, done

PR #523, opened 2015-07-21
Specialist document publishing api formatter

heathd:
The formatter takes a specialist document and generates a hash suitable for sending to the publishing API.

## Testing approach

The spec for the formatter takes a slightly different approach than some other specs in specialist publisher. Rather than stubbing out all of its collaborators, it constructs a real mongoid `SpecialistDocumentEdition` and fetches the real collaborators using the `SpecialistPublisherWiring`. This means that we are testing more objects together at the same time. 

Unit level test coverage of each individual object is quite low, so the heavily stubbed approach gives little confidence that the overall system works.

## Why are we only testing one specialist document type (rather than all of them)?

We considered adding tests which would iterate over every document type, construct an example document and present it.

However, this would give a false confidence because there would be no guarantee that our sample documents would match the document types which the editing interface would be capable of producing. We do not have a set of curated fixtures/factories in specialist publisher, and even if we did, there would again be no guarantee that they would match up with what the UI could produce.

So we felt it would be better to leave these out altogether.

Instead when we implement the exporter itself, we should amend the cucumber features for each specialist document type to include a check that a valid JSON document was sent to the publishing API.

## Contract tests

Specialist-publisher already has contract tests, we've just included the formatter spec in the list of specs run in `jenkins-schema.sh`.

### DIFF
--- a/spec/exporters/formatters/specialist_document_publishing_api_formatter_spec.rb
+++ b/spec/exporters/formatters/specialist_document_publishing_api_formatter_spec.rb
@@ -11,12 +11,22 @@ RSpec.describe SpecialistDocumentPublishingApiFormatter do
   }
   let(:formatter) {
     described_class.new(
-        document, specialist_document_renderer: specialist_document_renderer
+      document,
+      specialist_document_renderer: specialist_document_renderer,
+      publication_logs: publication_logs
     )
   }
 
-  let(:document) {
-    SpecialistDocument.new(nil, edition.id, [edition])
+  let(:publication_logs) { class_double("PublicationLog", change_notes_for: [publication_log]) }
+
+  let(:publication_log) {
+    instance_double("PublicationLog",
+      slug: document.slug,
+      title: document.title,
+      version_number: document.version_number,
+      change_note: "My change note",
+      published_at: 1.day.ago
+    )
   }
 
   let(:document) {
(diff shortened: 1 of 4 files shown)


spec/exporters/formatters/specialist_document_publishing_api_formatter_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+      let(:body) { "## Heading 2\n\nParagraph" }
+
+      it { should be_valid_against_schema("specialist_document") }
+
+      it "should convert the body from govspeak to html" do
+        expect(presented["details"]["body"]).to eq(%{<h2 id="heading-2">Heading 2</h2>\n\n<p>Paragraph</p>\n})
+      end
+    end
+
+    context "with a body containing a govspeak header" do
```
elliotcm: I'd like to see a test for a body containing no headers, since the code explicitly leaves out the "headers" attribute in that case rather than leaving it blank, and it would be good to concretize that behaviour.

spec/models/publication_log_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+            change_note: "Duplicate note",
+            version_number: 2,
+          )
+        end
+
+        it "removes duplicates" do
+          expect(PublicationLog.change_notes_for(slug)).to eq(change_notes_for_first_doc)
+        end
+      end
+    end
```
elliotcm: Could we have a test for what happens if there are no publication logs?

elliotcm:
Feature: agree
Implementation: broadly agree
Specifics: could do with a couple more tests as per inline comments

danielroseman:
Extra tests added.

PR #524, opened 2015-07-24
Send correct format for ESIF documents to email-alert-api

jamiecobbett:
We were sending "esi_fund" to email-alert-api because that's the string used as
the "document_type" internal to specialist-publisher. All of the other
exporters, however, use "european_structural_investment_fund". This means that
email subscriptions are created for the format
"european_structural_investment_fund", but email alerts are triggered for
"esi_fund", therefore, there aren't any matching subscriptions and so emails
weren't sent.

I couldn't see where and how to write a test for this, suggestions welcome.

### DIFF
--- a/app/exporters/formatters/abstract_document_publication_alert_formatter.rb
+++ b/app/exporters/formatters/abstract_document_publication_alert_formatter.rb
@@ -14,7 +14,7 @@ class AbstractDocumentPublicationAlertFormatter
 
   def tags
     arrayified_extra_fields.merge(
-      format: [document.document_type]
+      format: [format]
     )
   end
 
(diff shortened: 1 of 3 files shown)


boffbowsh:
Maybe to just prevent this edge case breaking, add a "its format should be xyz" spec here? https://github.com/alphagov/specialist-publisher/blob/send_correct_format_to_email_alert_api/spec/exporters/formatters/esi_fund_indexable_formatter_spec.rb

boffbowsh:
Or rather, its tags?

jamiecobbett:
Rebased with a test - thanks @boffbowsh

PR #525, opened 2015-07-24
Speed up publishing

elliotcm:
The intent of this PR is to get a big enough speed boost during the normal publishing flow to allow us to send to the draft publishing API and other services on every save instead of only on publish, and therefore allow easy previewing.

#### Only call the organisation API once when publishing.

We currently call the organisation API for the
manual and each section, but the organisation slug
used is the one from the manual, thus the response
is guaranteed to always be the same.

#### Avoid sending unchanged manual sections to publishing API.

This avoids a bunch of govspeak rendering and time
waiting on the API to respond.

### Speed gains

The current time taken to publish a manual is around 1.5s per section, give or take.  This leads to documents like the open document format manual taking 32 seconds to publish (measured on the preview environment).

After this change (and after the first time each document is published) the time taken to publish any document is around 3 seconds plus 1 second per changed section.

https://trello.com/c/yP4BSFf3/284-speed-up-publishing-of-manuals

### DIFF
--- a/app/exporters/manual_publishing_api_exporter.rb
+++ b/app/exporters/manual_publishing_api_exporter.rb
@@ -1,8 +1,8 @@
 class ManualPublishingAPIExporter
 
-  def initialize(export_recipent, organisations_api, manual_renderer, publication_logs, manual)
+  def initialize(export_recipent, organisation, manual_renderer, publication_logs, manual)
     @export_recipent = export_recipent
-    @organisations_api = organisations_api
+    @organisation = organisation
     @manual_renderer = manual_renderer
     @publication_logs = publication_logs
     @manual = manual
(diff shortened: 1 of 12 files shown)


danielroseman:
Looks good. I think we need a migration to set the `exported_at` date on all published manuals, so that the first update doesn't incur the cost of exporting everything.

elliotcm:
We can do, though it would be no worse than it already is.

danielroseman:
Well, I was talking about the impact when we enable exporting to content-preview on save. But fair enough, we can do the migration when we make that change.

elliotcm:
With manuals we can just send the relevant bit of the manual to the API since the save button is either on the manual cover page or on each individual section.

PR #529, opened 2015-07-30
Add task to publish finders to rummager.

alext:
At present the finders have been hacked into search by using
recommended-links. This has problems because we can't include their
topic taggings etc.

This therefore adds a task to send the finders to rummager. The finder descriptions have been populated from the data in recommended-links. The finder topics have been populated from the data in contentapi.

### DIFF
--- a/finders/metadata/cma-cases.json
+++ b/finders/metadata/cma-cases.json
@@ -3,6 +3,7 @@
   "base_path": "/cma-cases",
   "format_name": "Competition and Markets Authority case",
   "name": "Competition and Markets Authority cases",
+  "description": "Find reports and updates on current and historical CMA investigations",
   "filter": {
     "document_type": "cma_case"
   },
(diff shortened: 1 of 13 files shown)


finders/metadata/cma-cases.json
```diff
@@ -3,12 +3,20 @@
...
   "format_name": "Competition and Markets Authority case",
   "name": "Competition and Markets Authority cases",
+  "description": "Find reports and updates on current and historical CMA investigations",
   "filter": {
     "document_type": "cma_case"
   },
   "signup_content_id": "43dd2b13-93ec-4ca6-a7a4-e2eb5f5d485a",
   "signup_copy": "You'll get an email each time a case is updated or a new case is published.",
   "organisations": ["957eb4ec-089b-4f71-ba2a-dc69ac8919ea"],
+  "topics": [
```
tommyp: Is it a bit weird to have this listed as topics when it's actually the `specialist_sectors`?

jamiecobbett: I think "topic" is the new name for "specialist_sector", but I may not have that exactly right.

alext: specialist_sectors is the legacy/deprecated name for them.  They're called topics in all the new parts of the stack, so I wanted to keep that naming here.  At the point we send them to publishing-api, they will also be called topics.

PR #530, opened 2015-07-30
[WIP] Send draft manuals to content store

heathd:
This is still work in progress, but just opening a pull request so I can get some feedback on the general approach.


elliotcm:
The general approach looks reasonable to me.

I agree with you regarding the content ID though I haven't thought it through enough to know for sure whether there would be consequences.

app/services/abstract_manual_document_service_registry.rb
```diff
@@ -80,4 +88,46 @@ def manual_document_builder
...
+
+  def manual_document_renderer
+    SpecialistPublisherWiring.get(:manual_document_renderer)
+  end
+
+  def publishing_api
+    SpecialistPublisherWiring.get(:publishing_api)
+  end
+
+  def organisations_api
```
danielroseman: Since this is now referenced from more than one class, it should probably go into the SpecialistPublisherWiring, as we did with publishing_api.

danielroseman:
Just one comment, otherwise :+1:

app/services/abstract_manual_document_service_registry.rb
heathd: :+1: I agree, will do

PR #531, opened 2015-07-31
[WIP] Add script to rename a document slug

tijmenb:
`bin/rename_slug` renames a slug by withdrawing it so that the original page returns a `410 Gone`. It then republishes the document with the new slug.

The process is taken from the ops-manual, which currently has out-of-date code examples (will be updated after this is merged).

https://github.gds/pages/gds/opsmanual/2nd-line/changing-specialist-publisher-document-slug.html

Example:

    bin/rename_slug "50a02899-0fa9-44d0-aff5-30adeb9898d4" "cma-cases/new-slug-for-case"

Notes:

- The script is mostly cargo-culted from existing scripts in the repo. I'm assuming this is how things are supposed to be done in this app.
- specialist-publisher in in the process of moving to use the content-store. This process might need updating.  
- This method is not really safe, if republishing or `update_attributes!` fails, the document will stay `archived`. In that case you'll have to manually republish.

### DIFF
--- a/bin/rename_slug
+++ b/bin/rename_slug
@@ -1,0 +1,44 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require "specialist_publisher"
+
+begin
+  document_id = ARGV.fetch(0)
+  new_slug = ARGV.fetch(1)
+rescue KeyError
+  $stderr.puts "usage: #{File.basename(__FILE__)} document_id new_slug"
+  exit(1)
+end
+
+if new_slug.starts_with?("/")
+  $stderr.puts "Slug should not start with a slash."
+  exit(1)
+end
+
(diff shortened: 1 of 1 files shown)


mikejustdoit:
It might be clearer to call the script `rename_document_slug`, since (unfortunately) Specialist Publisher also handles Manuals

mikejustdoit:
I can't fault the cargo culting, but I've got a few small suggestions. I add them as comments, but if you're in the office tomorrow, we could chat in person?

jamiecobbett:
> It might be clearer to call the script rename_document_slug, since (unfortunately) Specialist Publisher also handles Manuals

Agreed.

tijmenb:
As per discussion on Slack: I'm closing this in favour of a PR on opsmanual. The team will revisit slug renaming at some point in the future.

PR #532, opened 2015-07-31
Fetch tag for manuals when publishing

tommyp:
In order for Manuals to appear on topic pages we need to fetch the topics in Content API and send those to Rummager when publishing. This PR adds a `TagFetcher` class which gets the tags and modifies the Manual before saving and publishing it.

The last commit has some shoddy stubbing at a random step. This looks a bit weird but I couldn't figure out where to stub this anywhere else that would result in my tests passing.

Paired on this with @KushalP. [Ticket](https://trello.com/c/QtMO88Uj/73-manuals-are-currently-topic-sub-topic-tagged-in-panopticon-so-the-details-of-the-tags-cannot-be-sent-to-rummager-when-the-manual).

### DIFF
--- a/app/exporters/formatters/manual_indexable_formatter.rb
+++ b/app/exporters/formatters/manual_indexable_formatter.rb
@@ -6,11 +6,21 @@ class ManualIndexableFormatter < AbstractIndexableFormatter
   end
 
 private
+  def extra_attributes
+    {
+      specialist_sectors: specialist_sectors
+    }
+  end
+
   def indexable_content
     entity.summary # Manuals don't have a body
   end
 
   def organisation_slugs
     [entity.organisation_slug]
+  end
+
+  def specialist_sectors
+    entity.tags.select { |t| t[:type] == "specialist_sector" }.map { |t| t[:slug] }
   end
 end
--- a/app/services/publish_manual_service.rb
+++ b/app/services/publish_manual_service.rb
@@ -1,3 +1,5 @@
+require "tag_fetcher"
+
 class PublishManualService
   def initialize(dependencies)
     @manual_id = dependencies.fetch(:manual_id)
@@ -8,6 +10,7 @@ class PublishManualService
 
   def call
     if versions_match?
+      update_manual_with_tags
       publish
       notify_listeners
       persist
@@ -36,6 +39,19 @@ private
 
   def publish
     manual.publish
+  end
+
+  def tags
+    TagFetcher.new(manual).tags.map { |t|
+      {
+        type: t.details.type,
+        slug: t.slug,
+      }
+    }
+  end
+
+  def update_manual_with_tags
+    manual.update({tags: tags})
   end
 
   def persist
--- a/spec/lib/tag_fetcher_spec.rb
+++ b/spec/lib/tag_fetcher_spec.rb
@@ -1,0 +1,63 @@
+require "spec_helper"
+require "tag_fetcher"
+require "gds_api/test_helpers/content_api"
+require "webmock/rspec"
+
+describe TagFetcher do
+
+  include GdsApi::TestHelpers::ContentApi
+
+  describe "tag fetching" do
+
+    let(:manual) {
+      Manual.new({
+        id: "guidance/style-guide",
+        slug: "guidance/style-guide",
+        title: "Style guide",
+        summary: "A summary",
+        body: "A body",
+        organisation_slug: "cabinet-office",
+        state: "draft",
+      })
+    }
+
+    let(:slug) { "guidance/style-guide" }
+    let(:tag_slug) { "government-digital-guidance/content-publishing" }
+
+    before do
+      WebMock.disable_net_connect!
+
+      artefact = artefact_for_slug(slug).merge(
+        "title" => @title,
+        "format" => "manual",
+        "details" => {
+          "body" => "<p>Body content</p>\n",
+          "summary" => "Summary of document",
+          "updated_at" => "2014-10-24T08:41:18Z",
+          "published_at" => "2014-10-24T08:41:18Z",
+        },
+        "tags" => [
+          {
+            "id" => "https://www.gov.uk/api/tags/specialist_sector/government-digital-guidance%2Fcontent-publishing.json",
+            "slug" => tag_slug,
+            "web_url" => "https://www.gov.uk/topic/government-digital-guidance/content-publishing",
+            "title" => "Content and publishing",
+          }
+        ]
+      )
+
+      content_api_has_an_artefact(
+        slug,
+        artefact
+      )
+    end
+
+    it "fetches tags for given manual" do
+      tags = TagFetcher.new(manual).tags
+
+      expect(tags.map(&:slug)).to eq([tag_slug])
+    end
+
+  end
+
+end
--- a/app/exporters/formatters/abstract_indexable_formatter.rb
+++ b/app/exporters/formatters/abstract_indexable_formatter.rb
@@ -30,7 +30,7 @@ private
   end
 
   def extra_attributes
-    {}
+    raise NotImplementedError
   end
 
   def title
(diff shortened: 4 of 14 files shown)


danielroseman:
I don't know anything about this story, but given we are just finishing up a big piece of work to untangle specialist-publisher from content-api it's a shame this adding another dependency on it. Is there any other way to get this data?

bradleywright:
We have a ticket to remove all connections with Content API from this publisher, which this PR will block. Can we get the topics from anywhere else?

KushalP:
@danielroseman the different options that we looked at for querying this data can be found here: https://gov-uk.atlassian.net/wiki/display/FS/HMRC+manuals+update+-+29+JUL+2015

We opted to query the Content API here because it would be easier to throw away in the near future and make sure that Rummager has the correct data.

bradleywright:
From some conversations I've had, it seems like this is (regrettably) the only viable solution that doesn't create more work later. So let's go with this, but make sure it's documented as tech debt on the Wiki.

bradleywright:
You're going to need to reconcile this PR with https://github.com/alphagov/specialist-publisher/pull/533, though.

KushalP:
> but make sure it's documented as tech debt on the Wiki

Are there any good examples of how we're doing this at the moment?

bradleywright:
> Are there any good examples of how we're doing this at the moment?

See the Whitehall app page.

KushalP:
> You're going to need to reconcile this PR with #533, though.

Yup. @tommyp and I have discussed and we're okay with dealing with the merge conflicts on this PR if #533 is likely to get merged soon. I'll chat to @elliotcm about this.

elliotcm:
We're expecting to merge https://github.com/alphagov/specialist-publisher/pull/533 some time today.

app/services/publish_manual_service.rb
```diff
@@ -8,6 +10,8 @@ def initialize(dependencies)
 
   def call
     if versions_match?
+      fetch_tags
+      update_manual_with_tags
```
mikejustdoit: Could there be an advantage to persisting a manual with its new tags regardless of the outcome of the publishing steps?

mikejustdoit: Are there any additional potential problems with this dependency between publishing and tag-fetching?

Are there situations where we'd currently (and/or in the near future) still be able to (and want to) publish while _some_ services are down? (If you see what I mean..?)

lib/tag_fetcher.rb
```diff
@@ -0,0 +1,29 @@
+require "gds_api/content_api"
+
+class TagFetcher
+
+  def initialize(manual)
+    @manual = manual
+  end
+
+  def fetch_tags
```
mikejustdoit: I think the convention in this app would be to call this either `call` or maybe at a push `tags`?

lib/tag_fetcher.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def artefact
+    content_api.artefact(manual.slug)
+  end
+
+  def content_api
+    @content_api ||= GdsApi::ContentApi.new(Plek.new.find("contentapi"))
+  end
+
+end
```
mikejustdoit: This `TagFetcher` is quite specific to this app, so maybe it would live in `app/lib` instead?

spec/lib/tag_fetcher_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      content_api_has_an_artefact(
+        @slug,
+        artefact
+      )
+    end
+
+    it "fetches tags for given manual" do
+      tags = TagFetcher.new(manual).fetch_tags
+
+      expect(tags.map(&:slug)).to eq([@tag_slug])
```
mikejustdoit: Could this `@tag_slug` be more idiomatic as a `double`?

app/services/publish_manual_service.rb
```diff
@@ -38,6 +42,19 @@ def publish
...
+    @tags = TagFetcher.new(manual).fetch_tags.map { |t|
+      {
+        type: t.details.type,
+        slug: t.slug,
+      }
+    }
+  end
+
+  def update_manual_with_tags
+    manual.update({tags: @tags})
```
mikejustdoit: It seems like we usually access instance variables in these classes via a `private attr_reader`

app/exporters/formatters/manual_indexable_formatter.rb
```diff
@@ -5,6 +5,10 @@ def type
     "manual"
   end
 
+  def indexable_attributes
+    super.merge({specialist_sectors: specialist_sectors})
```
mikejustdoit: ManualIndexableFormatter's superclass already merges an overwritable empty-by-default hash called `extra_attributes`... I _think_ that's where this extra stuff should be defined.

It might also be good to document that superclass feature in a quick separate commit. I'd say do that by either

- Adding a comment to the `extra_attributes` method in the superclass, or
- Replacing the superclass' implementation with a `raise NotImplementedException` and defining it as a method that returns an empty hash on all subclasses that don't currently have their own

tommyp:
Fixed the commits based on @mikejustdoit's comments.

app/services/publish_manual_service.rb
```diff
@@ -8,6 +10,8 @@ def initialize(dependencies)
 
   def call
     if versions_match?
+      tags
```
mikejustdoit: `update_manual_with_tags` already fetches the tags, so we shouldn't need to call `tags` separately here.

mikejustdoit: I'd still be interested to hear your answer to my questions that this doesn't address, too (https://github.com/alphagov/specialist-publisher/pull/532#discussion-diff-36097376)

KushalP: I'm going to answer the above questions out of order, because I think it might help to give more context.

> Are there any additional potential problems with this dependency between publishing and tag-fetching?

In the short-term, it's a problem because it slows down moving to the new publishing workflow. This discussion is documented in the [meeting notes](https://gov-uk.atlassian.net/wiki/display/FS/HMRC+manuals+update+-+29+JUL+2015). This PR unblocks a few things around manuals and collections. Removing this dependency will become something owned by the Finding Things team.

Additionally, it slows down the Publishing Platform team. They're trying to reduce the dependency that many of our publishing applications have on the [Content API](http://github.com/alphagov/govuk_content_api). The wider aim is to move everything over to use the [Content Store](https://github.com/alphagov/content-store) to produce a cleaner publishing workflow.

> Could there be an advantage to persisting a manual with its new tags regardless of the outcome of the publishing steps?

At the moment, the tagging is done in another application. Persisting a manual with its new tags for draft content would only be useful in terms of auditing what the tags for a given piece of content looked like when the draft was saved. As far as I'm aware, this isn't a user need at this time.

There are currently discussions going on about what this tagging implementation might look like and how it interacts with workflow and auditing. It's in the very early stages and this might be better there.

> Are there situations where we'd currently (and/or in the near future) still be able to (and want to) publish while some services are down? (If you see what I mean..?)

I think we want our publishing platform (in the future) to gracefully degrade where possible. However in the short term we have to make this step. It will help us update the data that we hold for manuals in [Rummager](http://github.com/alphagov/rummager) and their related tags. Once we have that in place we can start to use Rummager as one of the sources of truth for tagging content items and pull apart this dependency.

Let me know if there's anything I've missed.

mikejustdoit:
I think I'd prefer if new features like _fetching tags_ lived in their own services, but
1. @KushalP's explained why that isn't necessarily beneficial in this case, and
2. I don't think we've figured out a nice way to use multiple services like that yet

mikejustdoit:
:+1: I'm happy with this, but does it need rebasing or something before merging?

KushalP:
> I think I'd prefer if new features like fetching tags lived in their own services

That's one of the things being discussed in the early tagging discussions I mentioned :rocket:

tommyp:
> I think I'd prefer if new features like fetching tags lived in their own services

Further to @KushalP's reply, this App will go through some _major_ refactoring sooner or later.

PR #533, opened 2015-08-03
Remove panopticon and content api publishing

elliotcm:
This app now relies solely on the publishing API and content store.

https://trello.com/c/XW6wJpbJ/268-stop-publishing-specialist-documents-to-content-api


features/support/esi_fund_helpers.rb
```diff
@@ -32,15 +32,7 @@ def withdraw_esi_fund(*args)
...
-  def check_header_metadata_depth_is_limited(slug, depth: 1)
-    published_document = RenderedSpecialistDocument.find_by_slug(slug)
-
-    headers = published_document.details.fetch("headers")
-
-    expect(find_header_depths(headers).max).to eq(depth)
-  end
-
-  def document_body_with_deeply_nested_headers
+  def document_body
```
danielroseman: Given these are no longer doing anything special with the headers at all, the body no longer need to be defined separately in a helper - they should be inline as with the other document types (aaib, raib etc).

PR #534, opened 2015-08-06
Add link to content preview for draft items.

danielroseman:
**Don't merge until the change has been communicated to departments.**

Modifies the show page for manuals and specialist docs so that the slug
is no longer a link, and add "View on website" and "Preview draft" links
next to it. The former only shows if the item has been published, and
the latter only if there is an unpublished draft.

image

### DIFF
--- a/app/assets/stylesheets/_header.scss
+++ b/app/assets/stylesheets/_header.scss
@@ -7,5 +7,15 @@
     display: block;
     margin-top: 5px;
     font-weight: normal;
+    padding: 0;
+
+    & li {
+      display: inline;
+      list-style: none;
+
+      & + li:before {
+        content: "– ";
+      }
+    }
   }
 }
--- a/app/views/shared/_title.html.erb
+++ b/app/views/shared/_title.html.erb
@@ -1,0 +1,13 @@
+<h1 class="page-header"><%= document.title %>
+  <ul class='document-slug'>
+    <li title='<% if !document.published? %>When published this document will appear at this URL<% end %>'>
+      <%= document.slug %>
+    </li>
+    <% if document.published? %>
+      <li><%= link_to "View on website", public_url %></li>
+    <% end %>
+    <% if document.draft? %>
+      <li><%= link_to "Preview draft", content_preview_url(document) %></li>
+    <% end %>
+  </li>
+</h1>
--- a/app/views/specialist_documents/show.html.erb
+++ b/app/views/specialist_documents/show.html.erb
@@ -3,15 +3,8 @@
   <li class='active'><%= document.title %></li>
 <% end %>
 
-<h1 class="page-header"><%= document.title %>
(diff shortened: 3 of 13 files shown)


elliotcm:
Should we more clearly distinguish between live and draft links? eg
"View on live website - Preview on draft website"

app/views/shared/_title.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1 class="page-header"><%= document.title %>
+  <ul class='document-slug'>
+    <li title='<% if !document.published? %>When published this document will appear at this URL<% end %>'>
```
elliotcm: Do we have a pre-publish state for specialist documents which isn't draft?

tommyp:
What's the language we're using when differentiating the environments to editors? Would _live GOV.UK_ and _draft GOV.UK_ be better?

app/views/specialist_documents/show.html.erb
```diff
@@ -89,6 +82,7 @@
       <%= link_to "Edit document", [:edit, document], class: "btn btn-success" %>
     </div>
 
+
```
elliotcm: Aberrant newline.

elliotcm:
- General concept :+1: 
- Implementation :+1: 
- Specifics - some comments and questions

danielroseman:
Naming is difficult, this is the best that Paula and I could come up with. I wanted to avoid calling it specifically "draft" or "preview"; currently "preview" is the name of an environment which some editors use for scratch and training, but in the long term that will become something like "integration". We're not really using the "draft GOV.UK" name.

app/views/shared/_title.html.erb
danielroseman: No, a document is draft once it's been created. When a document has been published and a new draft created, both `document.published?` and `document.draft?` are true.

app/views/specialist_documents/show.html.erb
danielroseman: Removed.

elliotcm:
OK, we can always change it if it proves confusing to users.

elliotcm:
Did you mention that we shouldn't be merging this yet in standup?  If so could you add a note or checkbox to the description.

elliotcm:
:+1: to go when ready

bradleywright:
Has @fofr looked at this?

fofr:
@bradleywright I spoke with @danielroseman about the change yesterday and it looks good.

Regarding copy, we'll need to see how users interpret this.

app/assets/stylesheets/_header.scss
```diff
@@ -7,5 +7,15 @@
     display: block;
     margin-top: 5px;
     font-weight: normal;
+    padding: 0;
+
+    & li {
```
edds: You shouldn't need the `&` here should you? It is implicit when you are nesting selectors like this.

PR #537, opened 2015-08-13
Add leading slash to slugs when sending to Rummager

jennyd:
This adds the leading slash to the slug for the `id` and `link` fields when formatting manuals, manual sections and specialist documents to send to Rummager (finders already use the full `base_path`) and adds a script to migrate the existing documents in Rummager to the new format.

See the commit messages for the full details.

A [temporary workaround](https://github.com/alphagov/rummager/pull/479) for the broken links due to missing leading slashes has recently been deployed in Rummager, but we need to fix the data at source. This issue is fixed for HMRC manuals in a [PR on the API](https://github.com/alphagov/hmrc-manuals-api/pull/94).

### DIFF
--- a/app/exporters/formatters/manual_section_indexable_formatter.rb
+++ b/app/exporters/formatters/manual_section_indexable_formatter.rb
@@ -31,6 +31,10 @@ private
   end
 
   def link
-    section.slug
+    with_leading_slash(section.slug)
+  end
+
+  def with_leading_slash(slug)
+    slug.start_with?("/") ? slug : "/#{slug}"
   end
 end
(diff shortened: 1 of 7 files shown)


app/exporters/formatters/manual_section_indexable_formatter.rb
```diff
@@ -1,4 +1,6 @@
-class ManualSectionIndexableFormatter
+require "formatters/abstract_indexable_formatter"
+
+class ManualSectionIndexableFormatter < AbstractIndexableFormatter
   def initialize(section, manual)
     @section = section
```
mikejustdoit: I _think_ this could be `@entity = section` plus changes later on to use `entity` and not create a `attr_reader :section`, for more consistency with the other indexable formatters... But I assume there's a reason you chose not to do this already?

jennyd: No good reason, no - I think I noticed the `attr_reader :entity` later on and added the note in the commit message, and didn't consider using it instead of `section`. Done now :)

mikejustdoit:
Excellent pull request

PR #538, opened 2015-08-17
Send finders description to Content Store

jackscotti:
Finders have a description and should be sent to Content Store at the moment of
publishing.

Having the description stored in Content Store means that Finder Frontend will
be able to access it and use it (i.e., when generating a Finder's metadata for
search engines).

Part of: https://trello.com/c/zFUIJFaJ/287-add-meta-description-tags-to-finders

### DIFF
--- a/app/presenters/finder_content_item_presenter.rb
+++ b/app/presenters/finder_content_item_presenter.rb
@@ -34,7 +34,7 @@ private
   end
 
   def description
-    ""
+    metadata.fetch("description", "")
   end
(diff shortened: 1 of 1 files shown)


jackscotti:
PS: I've tested this by:
- republishing locally the finders using the `rake publishing_api:publish_finders` task in Specialist Publisher.
- accessing the finders data from Finder Frontend


screen shot 2015-08-17 at 11 00 56

PR #539, opened 2015-08-18
Show preview link when manual's document is draft

danielroseman:
Previously the preview link would only display when the manual itself
was edited; an edit to a document belonging to that manual does not
change the draft status of the manual itself so the preview link would
not be displayed. Changed the logic so that it also checks the child
documents' statuses.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -18,6 +18,14 @@ module ApplicationHelper
 
   end
 
+  def show_preview?(item)
+    if item.respond_to?(:documents)
+      item.draft? || item.documents.any?(&:draft?)
+    else
+      item.draft?
+    end
+  end
+
   def publication_task_state(task)
     zoned_time = time_with_local_zone(task.updated_at)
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -18,6 +18,14 @@ def state(document)
...
   end
 
+  def show_preview?(item)
+    draft = item.draft?
+    if item.respond_to?(:documents)
+      draft ||= item.documents.any?(&:draft?)
+    end
+    draft
+  end
+
```
elliotcm: This looks like it'll work, but would confuse me less like this:

```
def show_preview?(item)
  return item.draft? unless item.respond_to?(:documents)
  
  item.draft? || item.documents.any?(&:draft?)
end
```
or this:
```
def show_preview?(item)
  if item.respond_to?(:documents)
    item.draft? || item.documents.any?(&:draft?)
  else
    item.draft?
  end
end
```

What do you think?

danielroseman: I think the second one is best, updated.

PR #540, opened 2015-08-20
GDS editors can add attachments to any Manual

tommyp:
When the User comes to add an attachment, if they were a GDS editor editing a non GDS manual then this would come back with a "Manual not found". This commit fixes that by using a new registry which isn't scoped to an organisation for GDS editors.

This approach introduces some duplication. This is intended as it keeps the attachment code similar to the code that deals with manuals. The hope this is that this duplication can be removed more easily at a later date. 

This PR also refactors the background in some feature tests to make them less tied to CMA. 

[Ticket](https://trello.com/c/yo9nTcYO/85-attachments-can-t-be-added-to-manual-sections).

### DIFF
--- a/app/controllers/manual_document_attachments_controller.rb
+++ b/app/controllers/manual_document_attachments_controller.rb
@@ -41,6 +41,18 @@ class ManualDocumentAttachmentsController < ApplicationController
 
 private
   def services
+    if current_user_is_gds_editor?
+      gds_editor_services
+    else
+      organisational_services
+    end
+  end
+
+  def gds_editor_services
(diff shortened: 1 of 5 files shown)


mikejustdoit:
Thanks @tommyp, this is a decent solution. Maybe we can have a chat about it and look at how we're handling this situation elsewhere?

And possibly even look at where the bug was introduced (which I believe can be easily and quickly traced back to a recent pull request of mine :fearful:)

mikejustdoit:
Nice, self-contained commits :+1:

PR #542, opened 2015-08-21
  Add asylum support tribunal decisions finder

robmckinnon:
* Add specialist document finder for asylum support tribunal decisions.
* Code was generated via finder generator.
* Use "tribunal_decision" prefix to distinguish document attributes added for tribunal decisions.
* Tested in VM.

**Dependent on rummager PR:** https://github.com/alphagov/rummager/pull/502

### DIFF
--- a/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
+++ b/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
@@ -1,0 +1,23 @@
+require "formatters/abstract_specialist_document_indexable_formatter"
+
+class AsylumSupportDecisionIndexableFormatter < AbstractSpecialistDocumentIndexableFormatter
+  def type
+    "asylum_support_decision"
+  end
+
+private
+  def extra_attributes
+    {
+      tribunal_decision_decision_date: entity.tribunal_decision_decision_date,
+      tribunal_decision_judges: entity.tribunal_decision_judges,
+      tribunal_decision_category: entity.tribunal_decision_category,
+      tribunal_decision_sub_category: entity.tribunal_decision_sub_category,
+      tribunal_decision_landmark: entity.tribunal_decision_landmark,
+      tribunal_decision_reference_number: entity.tribunal_decision_reference_number,
+    }
+  end
+
+  def organisation_slugs
+    ["first-tier-tribunal-asylum-support"]
+  end
+end
--- a/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
+++ b/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
@@ -14,6 +14,7 @@ private
       tribunal_decision_sub_category: entity.tribunal_decision_sub_category,
       tribunal_decision_landmark: entity.tribunal_decision_landmark,
       tribunal_decision_reference_number: entity.tribunal_decision_reference_number,
+      indexable_content: entity.hidden_indexable_content || entity.body
     }
   end
 
--- a/app/lib/specialist_publisher.rb
+++ b/app/lib/specialist_publisher.rb
@@ -23,8 +23,8 @@ module SpecialistPublisher
     OBSERVER_MAP.keys
   end
 
+private
   OBSERVER_MAP = {
-    "asylum_support_decision" => AsylumSupportDecisionObserversRegistry,
     "aaib_report" => AaibReportObserversRegistry,
     "asylum_support_decision" => AsylumSupportDecisionObserversRegistry,
     "cma_case" => CmaCaseObserversRegistry,
@@ -37,8 +37,6 @@ module SpecialistPublisher
     "raib_report" => RaibReportObserversRegistry,
     "vehicle_recalls_and_faults_alert" => VehicleRecallsAndFaultsAlertObserversRegistry,
   }.freeze
-
-  private
 
   def view_adapters
     SpecialistPublisherWiring.get(:view_adapter_registry)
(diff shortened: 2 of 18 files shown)


app/lib/specialist_publisher.rb
```diff
@@ -25,6 +25,7 @@ def document_types
 
 private
   OBSERVER_MAP = {
+    "asylum_support_decision" => AsylumSupportDecisionObserversRegistry,
```
tommyp: Alphabeticalness.

robmckinnon:
The business are currently iterating document metadata and allowed values for facets. I generated the code in this PR using a [custom rails generator](http://guides.rubyonrails.org/generators.html). The generator allows us to quickly make code updates across the `specialist-publisher` and `rummager` repositories.

I understand there is a convention to alphabetically sort values in hashes and methods in classes in `specialist-publisher`. Using the rails generator helper method [`insert_into_file()`](http://www.rubydoc.info/github/wycats/thor/Thor/Actions#insert_into_file-instance_method) does not allow for this alphabetically ordered placement.

Can we discuss, in person's probably best, whether the alphabetical sorting convention can be relaxed given the circumstances?

Before we have that discussion I'll investigate sorting hash keys and using the [`gsub_file()`](http://www.rubydoc.info/github/wycats/thor/Thor/Actions#gsub_file-instance_method) helper to change a hash. I'll test whether this prevents the ability to revert the generated code using `rails destory finder`.

app/lib/specialist_publisher.rb
```diff
@@ -23,9 +23,10 @@ def document_types
     OBSERVER_MAP.keys
   end
 
-private
   OBSERVER_MAP = {
+    "asylum_support_decision" => AsylumSupportDecisionObserversRegistry,
     "aaib_report" => AaibReportObserversRegistry,
+    "asylum_support_decision" => AsylumSupportDecisionObserversRegistry,
```
tommyp: This is duplicated.

app/lib/specialist_publisher.rb
```diff
@@ -37,6 +38,8 @@ def document_types
     "vehicle_recalls_and_faults_alert" => VehicleRecallsAndFaultsAlertObserversRegistry,
   }.freeze
 
+  private
```
tommyp: [This shouldn't be indented](https://github.com/alphagov/styleguides/blob/master/ruby.md#classes).

app/view_adapters/view_adapter_registry.rb
```diff
@@ -17,6 +17,7 @@ def get(type)
       "medical_safety_alert" => MedicalSafetyAlertViewAdapter,
       "raib_report" => RaibReportViewAdapter,
       "vehicle_recalls_and_faults_alert" => VehicleRecallsAndFaultsAlertViewAdapter,
-    }.fetch(type)
+       "asylum_support_decision" => AsylumSupportDecisionViewAdapter,
+   }.fetch(type)
```
tommyp: Indentation change here?

robmckinnon:
Happy to make changes as suggested.

app/lib/specialist_publisher.rb
robmckinnon: I'll push a fix putting private keyword back in original position, and removing the duplicated mapping.

app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
```diff
@@ -0,0 +1,23 @@
...
+class AsylumSupportDecisionIndexableFormatter < AbstractSpecialistDocumentIndexableFormatter
+  def type
+    "asylum_support_decision"
+  end
+
+private
+  def extra_attributes
+    {
+      tribunal_decision_decision_date: entity.tribunal_decision_decision_date,
+      tribunal_decision_judges: entity.tribunal_decision_judges,
```
tijmenb: @robmckinnon 

The way to avoid having to add the `allowed_values` in rummager (discussed here:  https://github.com/alphagov/rummager/pull/502#discussion-diff-37851721), is to send the full names of the judges to rummager as well. 

The line above sends the parameterised version like `tribunal_decision_judges: ['victoria-walker']`. You would need to add something like `tribunal_decision_judge_names: ["Victoria Walker"]`. Some example voodoo to get the label for a value from the allowed_values is here: https://github.com/alphagov/specialist-publisher/commit/42a62591f474b15f51348109cae73915044d9207. I don't know if that is the "correct" way to do that.

This would enable you to use the full name value in `asylum-support-decisions.json`. Instead of having one entry do both `display_as_result_metadata: true` and `filterable: true`, you separate those out. This is an example of how to do that: https://github.com/alphagov/specialist-publisher/commit/437b89323568324aeb8212f2bd74b34ecf27f425.

I hope this makes sense.

app/view_adapters/view_adapter_registry.rb
robmckinnon: Commiting change that corrects indenting, and makes mapping a constant hash with sorted keys.

app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
rboulton: :+1:

robmckinnon: @tijmenb Is it ok with you if we merge what we have so far, and make the changes you have suggested in a new branch?

tijmenb: Yes, that's fine!

tommyp:
🎉🎉🎉

PR #543, opened 2015-08-21
WIP: Add code generator that creates new specialist document finders

robmckinnon:
Run generator to add code required for new specialist document finder to specialist-publisher and rummager repos.

Usage details in lib/generators/finder/USAGE


robmckinnon:
Off the table for now.

PR #544, opened 2015-08-26
Prevent finder publishing when preview_only set and not in preview env

robmckinnon:
This allows for new finders to be merged onto master for use in preview environment, while preventing them going live on gov.uk. This feature flag was agreed with @jennyd.

It will still publish finder if RAILS_ENV is not production. Set preview_only in the document metadata file when needed.

Also refactored spec to make it easier to reuse metadata and schema fixture setup.

### DIFF
--- a/lib/publishing_api_finder_publisher.rb
+++ b/lib/publishing_api_finder_publisher.rb
@@ -3,24 +3,43 @@ require_relative "../app/presenters/finder_content_item_presenter"
 require_relative "../app/presenters/finder_signup_content_item_presenter"
 
 class PublishingApiFinderPublisher
-  def initialize(metadata, schemae)
+  def initialize(metadata, schemae, log = true)
     @metadata = metadata
     @schemae = schemae
+    @log = log
   end
 
   def call
-    metadata.zip(schemae).map { |metadata, schema|
-      if metadata[:file].has_key?("content_id")
-        export_finder(metadata, schema)
-        export_signup(metadata) if metadata[:file].has_key?("signup_content_id")
+    metadata.zip(schemae).map do |metadata, schema|
+      if metadata[:file].has_key?("content_id") && !preview_only?(metadata)
+        publish metadata, schema
+      elsif preview_only?(metadata)
+        if preview_domain_or_not_production?
+          publish metadata, schema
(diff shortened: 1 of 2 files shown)


lib/publishing_api_finder_publisher.rb
```diff
@@ -3,25 +3,44 @@
...
+    export_finder(metadata, schema)
+    export_signup(metadata) if metadata[:file].has_key?("signup_content_id")
+  end
+
+  def preview_only? metadata
+    metadata[:file]["preview_only"] == true
+  end
+
+  def preview_domain_or_not_production?
+    ENV.fetch("GOVUK_APP_DOMAIN", "")[/preview/] || !Rails.env.production?
```
bradleywright: This is the wrong way to do this as it'll break if we ever change host names. We should be using some config, eg an initializer, which is overridden in preview and development.

Something like `config.publish_preview_finders = true`.

robmckinnon: The requirement is for a preview only finder to be available in the preview environment, but not any other deployed environments. I was told the preview environment has `RAILS_ENV` set to `production`. Your suggestion of a config setting requires separate config environment files for preview, draft, and public environments. Is this the case?

This approach of identifying the preview environment using the `GOVUK_APP_DOMAIN` was recommended by @jennyd. Maybe you could discuss this further with her?

jennyd:
I suggested using an existing environment variable because it was a quicker way for @robmckinnon to get this done now, without him needing to make changes to Puppet to add a new environment variable or dealing with alphagov-deployment. We'll make this more robust soon, but as it is this helps unblock the first MoJ finder.

jennyd:
There's a separate [RummagerFinderPublisher](https://github.com/alphagov/specialist-publisher/blob/9f2d1814c3b7470f1f46bc44a5f099ead69414ea/lib/rummager_finder_publisher.rb#L11-L18) which still uses the same logic. We should refactor so that the code to determine whether or not to publish a finder in a particular environment is in one place.

bradleywright:
Cool. Just as long as we plan to fix it soon - once we move to the service
domain for preview this code won't work anymore.

PR #548, opened 2015-08-27
Fix name mismatch for asylum support decisions

tommyp:
The name of the Finder Metadata and Schema files needs to match or else
Dir.glob doesn't match them up. We should probably move this to be in
the same files.


jamiecobbett:
This should fix the immediate problem and make specialist-publisher deployable again. We need to change the rake task and/or `PublishingApiFinderPublisher` class so that it doesn't assume the glob returns things in the same order. @jennyd and I will make sure that happens as soon as possible.

PR #549, opened 2015-08-28
Revise asylum support decision facets allowed values

robmckinnon:
* Change asylum support facet allowed values for `category` and `sub-category`.
* Change order of facet filters.

### DIFF
--- a/finders/schemas/asylum-support-decisions.json
+++ b/finders/schemas/asylum-support-decisions.json
@@ -2,14 +2,6 @@
   "document_noun": "decision",
   "facets": [
     {
-      "key": "tribunal_decision_decision_date",
-      "name": "Decision date",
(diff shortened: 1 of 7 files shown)


tommyp:
There's lots of unnecessary churn in this PR. The only thing that determines the order of facets is the order in the schema.

finders/schemas/asylum-support-decisions.json
```diff
@@ -226,6 +272,14 @@
...
+    {
+      "key": "tribunal_decision_decision_date",
+      "name": "Decision date",
+      "type": "date",
+      "preposition": null,
+      "display_as_result_metadata": true,
+      "filterable": true
     }
   ]
 }
```
tommyp: Missing newline.

PR #550, opened 2015-08-28
Refactor finder publishing

edds:
To create the finder content items we were previously running a glob
over two different folders and then zip'ing the resultant file arrays
together. This assumed that both sets of files would be read of disk in
the same order.

This changes that so that instead we now read both folders but then only
select the files that exist in both folders. We then load each of those
files by name so that we can be sure that the files are matched together
correctly.

To do this I have refactored the arguments to the publishing publisher
to take a single object which is an array of finders. Each finder
contains both the schema and the metadata to then create the finder
from. Internally I have modified many of the methods to pass the whole
finder object around rather than passing just the metadata or schema as
they are individually needed.

https://trello.com/c/yFKkfBLp/98-specialist-publisher-finder-publishing-can-silently-mix-finder-facets-between-finders

### DIFF
--- a/lib/publishing_api_finder_loader.rb
+++ b/lib/publishing_api_finder_loader.rb
@@ -1,0 +1,47 @@
+require "multi_json"
+
+class PublishingApiFinderLoader
+  def initialize(folder)
+    @folder = folder
+  end
+
+  def finders
+    matching_finder_filenames.map do |filename|
+      metadata_file = File.join(folder, "metadata", filename)
+      schema_file = File.join(folder, "schemas", filename)
+
+      {
+        metadata: MultiJson.load(File.read(metadata_file)),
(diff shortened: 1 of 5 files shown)


mikejustdoit:
Nice one @edds, matching the files definitely feels like progress.

It'd be nice to have the rake task complain about non-matching files..?

In fact, I was wondering if we could have the globbing/file-matching moved to its own object? Then we could also use it to perform checks for non-matching files in rake's default task and on the CI build

edds:
Hey @mikejustdoit, I have split the loading and matching off into it's own lib and added some tests around it (pretty glad I did as it highlighted some typos, a nice reminder of why TDD is a good thing). Willing to leave adding it to the CI build as an exercise to the reader but I think rest should be good if you are happy with it.

PR #552, opened 2015-09-02
Store asylum support facet labels directly in rummager

robmckinnon:
This approach means we no longer need to set `allowed_values` in the `rummager` schema configuration. In `finder-frontend` metadata labels now come directly from `rummager` as separate `_name` fields.

Tested in the VM.

**Requires this rummager PR be merged**: https://github.com/alphagov/rummager/pull/508

### DIFF
--- a/app/exporters/formatters/abstract_specialist_document_indexable_formatter.rb
+++ b/app/exporters/formatters/abstract_specialist_document_indexable_formatter.rb
@@ -2,6 +2,12 @@ require "formatters/abstract_indexable_formatter"
 
 class AbstractSpecialistDocumentIndexableFormatter < AbstractIndexableFormatter
 
+  def expand_value(key)
+    schema = SpecialistPublisherWiring.get("#{type}_finder_schema".to_sym)
+    value = entity.send(key)
+    schema.humanized_facet_value(key, value)
+  end
+
 private
   def public_timestamp
     entity.public_updated_at || entity.updated_at
--- a/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
+++ b/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
@@ -9,9 +9,13 @@ private
   def extra_attributes
     {
       tribunal_decision_judges: entity.tribunal_decision_judges,
+      tribunal_decision_judges_name: expand_value(:tribunal_decision_judges),
       tribunal_decision_category: entity.tribunal_decision_category,
+      tribunal_decision_category_name: expand_value(:tribunal_decision_category),
       tribunal_decision_sub_category: entity.tribunal_decision_sub_category,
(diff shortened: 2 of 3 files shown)


robmckinnon:
This change follows the approach recommended by @tijmenb - as demonstrated in this branch: https://github.com/alphagov/specialist-publisher/compare/index-fields-directly

robmckinnon:
Also used this pull request to set preposition and short_name for asylum support decision_date field - a minor change.

mikejustdoit:
@robmckinnon, the minor change you mention might be fine as part of this pull request, but do you think it would be better off in its own commit?

app/exporters/formatters/abstract_specialist_document_indexable_formatter.rb
```diff
@@ -2,6 +2,12 @@
 
 class AbstractSpecialistDocumentIndexableFormatter < AbstractIndexableFormatter
 
+  def expand_value(key)
```
mikejustdoit: Should this be a private method?

robmckinnon:
@mikejustdoit Frequently on other pull requests comments have been left requesting minor commits be squashed. I'm happy with this minor not being in a separate commit this time.

app/exporters/formatters/abstract_specialist_document_indexable_formatter.rb
robmckinnon: Thanks. Having a look at this now.

robmckinnon:
Closing for now. Adding more specs for `expand_value()`.

mikejustdoit:
@robmckinnon, I think as a rule, logically distinct changes should be in separate commits.

mikejustdoit:
Also, I'd say if you're tempted to explain further details in a pull request's description or comments, it's a good sign that there are more things happening than the commits are letting on.

At the very least that merits adding the additional information to the commit messages themselves.

It's typically considered easier (and more reliable) to find information in commit messages than on pull requests

robmckinnon:
@mikejustdoit ok, understand your reasoning. I'll keep this sort of commit separate in the future. Is it ok if we let this change sit in the same commit this time? Otherwise it'd be very time consuming to create a new git history.

mikejustdoit:
@robmckinnon, it's trivial to make the change from one commit to two. But I understand if you're not used to doing it. One of us would be more than happy to help if you'd like to pair on it!

robmckinnon:
My preference is to leave it as it is for this time. Is that ok with you?

mikejustdoit:
No, @robmckinnon, sorry but when this is discovered by a confused developer sometime in the future, they'll only end up more confused.

As I said, if you'd like to pair on making the git history more usable, let me know when you're available :smiley_cat:

robmckinnon:
Re-opened as a new pull request with git history changes here: https://github.com/alphagov/specialist-publisher/pull/555

PR #553, opened 2015-09-02
Use phase correctly

tommyp:
With the changes in https://github.com/alphagov/finder-frontend/pull/230
we now use the `phase` attribute from the top level of the Content Item
hash. This commit modifies the check to merge this if it is present in
the metadata along with adding a test to test this behaviour.

### DIFF
--- a/spec/lib/publishing_api_finder_publisher_spec.rb
+++ b/spec/lib/publishing_api_finder_publisher_spec.rb
@@ -85,6 +85,17 @@ describe PublishingApiFinderPublisher do
       PublishingApiFinderPublisher.new(finders, false).call
     end
 
+    it "can publish a Finder with a phase" do
+      finders = [
+        make_finder("/finder-with-phase", "phase" => "beta"),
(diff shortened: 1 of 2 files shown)


spec/lib/publishing_api_finder_publisher_spec.rb
```diff
@@ -19,7 +19,7 @@
       }
     end
 
-    def make_metadata base_path, overrides = {}
+    def make_metadata(base_path, overrides = {})
```
mikejustdoit: The change on this line is excellent, but unrelated to the rest of the commit... can we make it a separate one?

tommyp: Ah yeah. Meant to do that in a separate commit but forgot I'd changed it. Added it to an extra commit and pushed it.

mikejustdoit:
Other than the minor comment, looks good!

Nice to see some context in the commit message :+1:

PR #555, opened 2015-09-03
Store asylum support facet labels directly in rummager

robmckinnon:
* No longer set `allowed_values` in the rummager schema configuration.
* In finder-frontend metadata labels now come directly from rummager as separate `_name` fields.
* Tested in the VM.
* Add specs for FinderSchema `humanized_facet_value()` to confirm expected behaviour.
* Minor decision date field config change.

Associated rummager PR is: alphagov/rummager#508

### DIFF
--- a/finders/schemas/asylum-support-decisions.json
+++ b/finders/schemas/asylum-support-decisions.json
@@ -6,7 +6,7 @@
       "name": "Judges",
       "type": "text",
       "preposition": "by judge",
-      "display_as_result_metadata": true,
+      "display_as_result_metadata": false,
       "filterable": true,
       "allowed_values": [
         {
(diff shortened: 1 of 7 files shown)


finders/schemas/asylum-support-decisions.json
```diff
@@ -266,18 +286,25 @@
...
       "display_as_result_metadata": true,
       "filterable": false
     },
     {
       "key": "tribunal_decision_decision_date",
       "name": "Decision date",
+      "short_name": "Decided",
       "type": "date",
-      "preposition": null,
+      "preposition": "was decided",
```
tommyp: Should this be `was decided on`?

robmckinnon: Good spot. Changing it to `decided`. Then frontend search results appear like this:

1 decision **decided** between 27 April 2006 or 2 April 2006
6 decisions **decided** after 27 April 2006

Aside: noticed finder-frontend shows "27 April 2006 *or* 28 April 2006" when it should be "27 April 2006 *and* 28 April 2006".

robmckinnon:
Addressed comments on pull request with requested changes - should be good to merge hopefully.

PR #556, opened 2015-09-09
Add email signup configuration to asylum support decisions

robmckinnon:
Allow user to filter email subscription by tribunal decision category.


robmckinnon:
Subscription page looks like this:

asylum-support-email-signup-screenshot

PR #557, opened 2015-09-09
Make both body and hidden_indexable_content indexable

robmckinnon:
As recommended by search team members here: https://github.com/ministryofjustice/specialist-publisher/commit/512a73301be1f9909573b847e23222a4bcd126cd#commitcomment-12928833

"There's little harm if there's duplication between `body` and `indexable_content_indexable`, and there's
definite benefit to making sure any text we have available is searchable."

### DIFF
--- a/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
+++ b/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb
@@ -18,7 +18,7 @@ private
       tribunal_decision_landmark_name: expand_value(:tribunal_decision_landmark),
       tribunal_decision_reference_number: entity.tribunal_decision_reference_number,
       tribunal_decision_decision_date: entity.tribunal_decision_decision_date,
-      indexable_content: entity.hidden_indexable_content || entity.body
+      indexable_content: "#{entity.hidden_indexable_content}\n#{entity.body}".strip
     }
   end
 
--- a/spec/exporters/formatters/asylum_support_decision_indexable_formatter_spec.rb
+++ b/spec/exporters/formatters/asylum_support_decision_indexable_formatter_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe AsylumSupportDecisionIndexableFormatter do
   let(:document) {
     double(
       :asylum_support_decision,
-      body: double("body"),
+      body: double,
       slug: "/slug",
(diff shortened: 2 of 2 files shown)


tijmenb:
Great! Would it be good to also index `tribunal_decision_reference_number` here? Do you think people will search for that?

robmckinnon:
@tijmenb Are you suggesting we should remove `tribunal_decision_reference_number` from the `extra_attributes` at [this line](https://github.com/ministryofjustice/specialist-publisher/blob/asylum-support-index-body/app/exporters/formatters/asylum_support_decision_indexable_formatter.rb#L19) in the indexable formatter?

tijmenb:
No, suggesting that we do `indexable_content: "#{entity.hidden_indexable_content}\n#{entity.tribunal_decision_reference_number}\n#{entity.body}".strip` so that if people search for the reference number in the normal GOV.UK search they'll find the document. But this might not be necessary - no idea if people would actually search for those.

robmckinnon:
All tribunal decision examples I've seen so far have the reference number in the decision text. So don't think we need to explicitly include it in the `indexable_content`. I'll confirm this with @tomwm.

tomwm:
not sure i understand. Yes, reference numbers are 'usually' in the content of the decision doc (caveat: i cant guarantee this for all tribunals because we've not been in contact with them all and currently they use different decision formats within the tribunals themselves)

robmckinnon:
@tijmenb Suggest we merge this as is for now. Ok with you?

We could always change in rummager the `tribunal_decision_reference_number` field definition from `identifier` to `searchable_identifier` later.

tijmenb:
Yes! :+1:

rboulton:
Merged.

On point to note - I imagine lots of these finders will have a "decision reference number" - if at all possible, it would be good to make sure these go in the same field, so that it would be possible to allow searching across all tribunals given just a reference number, without having to know which tribunal it's for.  Similarly if there are any other "shared" reference numbers.  ie, don't make new fields for every finder, if there's some shared fields you can use instead.

robmckinnon:
@rboulton Where possible I'll reuse fields that are semantically the same across tribunal finders. As you say `tribunal_decision_reference_number` is a good example of a field where this reuse makes sense.

PR #559, opened 2015-09-10
Truncate metadata values to 140 chars in document summary

robmckinnon:
Long metadata values take up too much space on the summary page if left untruncated. e.g. Tribunal decision's `hidden_indexable_content` field, which is the length of the entire document.

### DIFF
--- a/app/views/specialist_documents/show.html.erb
+++ b/app/views/specialist_documents/show.html.erb
@@ -18,7 +18,7 @@
       <% document.humanized_attributes.each_pair do |label, values| %>
         <dt><%= label.to_s.humanize %></dt>
         <% Array(values).each do |value| %>
-          <dd><%= value %></dd>
+          <dd><%= truncate(value, length: 140) %></dd>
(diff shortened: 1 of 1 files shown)


robmckinnon:
Truncated value examples below, number of lines vary depending on screen width:

screen shot 2015-09-10 at 14 00 32

screen shot 2015-09-10 at 14 01 19

tommyp:
Nice work with this, especially testing the special casing :+1: 

Maybe we could make this change even simpler? Truncating at 140 will be fine for values of all the metadata across all the Specialist Document Formats. Plus, the less special cases we have in Specialist Publisher, the better. I think it's fine to reduce this PR down to `<dd><%= truncate(value, length: 140)` %></dd>`. Does that sound reasonable?

robmckinnon:
Removing the special case makes sense. I'll make that change.

robmckinnon:
Changed pull request to truncate all metadata values over 140 chars, as suggested.

PR #564, opened 2015-09-14
Stop stripping leading slashes from manual sections

issyl0:
- For consistency with how the HMRC Manuals API deals with manual `id`s,
  `link`s and (soon[1]) `manual` attributes in sections, don't strip leading
  slashes.
- This inconsistency was breaking search within a manual
  (https://trello.com/c/KAR7YtA7/119-searching-within-a-manual-uses-the-wrong-layout).

_NOTE_: As part of deploying this, the `republish_manuals` bin script needs to be run.

[1] - PR to be opened shortly.

(Paired with @mikejustdoit on this.)

### DIFF
--- a/app/exporters/formatters/manual_section_indexable_formatter.rb
+++ b/app/exporters/formatters/manual_section_indexable_formatter.rb
@@ -15,8 +15,12 @@ private
 
   def extra_attributes
     {
-      manual: manual.slug,
+      manual: manual_slug,
     }
+  end
+
+  def manual_slug
+    with_leading_slash(manual.slug)
   end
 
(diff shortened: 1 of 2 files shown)


tommyp:
@issyl0 Do you want to test running the script on Preview and if it's all good we can pair on doing it on Staging and Production?

tommyp:
Or rather, book a deploy slot and I'll be happy to pair with you on doing it.

issyl0:
@tommyp That sounds good. I'll do it on Preview a bit later and then if it works I'll book a deploy slot. Thanks!

PR #566, opened 2015-09-16
Validate tribunal decision sub_category matches parent category

robmckinnon:
This is a solution to a concern raised from the Finder team that it was possible for bad category/sub-category data to be entered by admin users creating tribunal decision documents.

A provided tribunal decision sub-category must be related to the supplied parent category. Some parent categories will not have any sub-categories related to them. A tribunal decision with a blank sub-category will be valid if the parent category has no sub-categories. Otherwise presence of a related sub-category is required.

Screenshots of the validation errors:
screen shot 2015-09-24 at 11 43 17
screen shot 2015-09-24 at 11 43 35
screen shot 2015-09-24 at 11 43 54
screen shot 2015-09-24 at 11 42 49

Changes:
* Add a `TribunalDecisionSubCategoryValidator` for use by tribunal decision finders.
* Validate UTAAC `tribunal_decision_sub_category` facet value is prefixed by the parent category facet value.
* Validate asylum support `tribunal_decision_sub_category` facet value is prefixed by the alternate parent category facet value prefix.
* Error message explains how to select valid sub-category.
* Allow sub-category to be optional.
* Use type-ahead select UI for tribunal decision sub-categories to allow blank sub-category to be provided.
* Validate that only a single sub-category is entered.
* Validate that a sub-category is provided, if there are sub-categories present for given parent category.

### DIFF
--- a/app/models/validators/asylum_support_decision_validator.rb
+++ b/app/models/validators/asylum_support_decision_validator.rb
@@ -1,6 +1,7 @@
 require "delegate"
 require "validators/date_validator"
 require "validators/safe_html_validator"
+require "validators/tribunal_decision_sub_category_relates_to_parent_validator"
 
 class AsylumSupportDecisionValidator < SimpleDelegator
   include ActiveModel::Validations
@@ -14,6 +15,16 @@ class AsylumSupportDecisionValidator < SimpleDelegator
   validates :tribunal_decision_judges, presence: true
   validates :tribunal_decision_landmark, presence: true
   validates :tribunal_decision_reference_number, presence: true
-  validates :tribunal_decision_sub_category, presence: true
+  validates :tribunal_decision_sub_category, presence: true, tribunal_decision_sub_category_relates_to_parent: true
 
+  def category_prefix_for(category)
+    case category
+    when "section-95-asylum-seekers"
+      "section-95"
+    when "section-4-2-failed-asylum-seekers"
+      "section-4-2"
+    when "section-4-1-neither-an-asylum-seeker-nor-a-failed-asylum-seeker"
+      "section-4-1"
+    end
+  end
 end
--- a/app/models/validators/asylum_support_decision_validator.rb
+++ b/app/models/validators/asylum_support_decision_validator.rb
@@ -15,7 +15,7 @@ class AsylumSupportDecisionValidator < SimpleDelegator
   validates :tribunal_decision_judges, presence: true
   validates :tribunal_decision_landmark, presence: true
   validates :tribunal_decision_reference_number, presence: true
-  validates :tribunal_decision_sub_category, presence: true, tribunal_decision_sub_category_relates_to_parent: true
+  validates :tribunal_decision_sub_category, tribunal_decision_sub_category_relates_to_parent: true
 
   def category_prefix_for(category)
     case category
--- a/app/models/validators/utaac_decision_validator.rb
+++ b/app/models/validators/utaac_decision_validator.rb
@@ -1,6 +1,7 @@
 require "delegate"
 require "validators/date_validator"
 require "validators/safe_html_validator"
+require "validators/tribunal_decision_sub_category_relates_to_parent_validator"
 
 class UtaacDecisionValidator < SimpleDelegator
   include ActiveModel::Validations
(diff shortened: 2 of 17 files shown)


app/models/validators/tribunal_decision_sub_category_validator.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    unless prefixed_by_parent_category? sub_category, validator
+      message = "change to be a sub-category of '#{category_label(validator)}' OR change Category"
+      validator.errors.add attribute, message
+    end
+  end
+
+  private
+
+  def prefixed_by_parent_category? sub_category, validator
```
tommyp: Method args should be wrapped in parenthesis as per [the styleguide](https://github.com/alphagov/styleguides/blob/master/ruby.md#syntax).

app/models/validators/utaac_decision_validator.rb
```diff
@@ -12,6 +13,6 @@ class UtaacDecisionValidator < SimpleDelegator
   validates :tribunal_decision_category, presence: true
   validates :tribunal_decision_decision_date, presence: true, date: true
   validates :tribunal_decision_judges, presence: true
-  validates :tribunal_decision_sub_category, presence: true
+  validates :tribunal_decision_sub_category, presence: true, tribunal_decision_sub_category: true
```
tommyp: This line looks a bit weird to me. Could the `tribunal_decision_sub_category: true` bit maybe explain _what_ is being validated here? Presumably that the subcategory matches the parent category?

robmckinnon: Rails uses this symbol to determine the name of the validator class. e.g. Rails uses the `DateValidator` class when `date: true`, the `TribunalDecisionSubCategoryValidator` class when `tribunal_decision_sub_category: true`, etc.

Based on our meeting this morning I'm working on extending the validator to perform additional checks beyond the parent category check. Think `TribunalDecisionSubCategoryValidator` / `tribunal_decision_sub_category: true` still works, although awkward. What do you think?

app/models/validators/asylum_support_decision_validator.rb
```diff
@@ -14,6 +15,16 @@ class AsylumSupportDecisionValidator < SimpleDelegator
   validates :tribunal_decision_judges, presence: true
   validates :tribunal_decision_landmark, presence: true
   validates :tribunal_decision_reference_number, presence: true
-  validates :tribunal_decision_sub_category, presence: true
+  validates :tribunal_decision_sub_category, tribunal_decision_sub_category: true
```
tommyp: Further to the recent conversation about this line, my point was that `validates :tribunal_decision_sub_category, tribunal_decision_sub_category: true` does not say how the `tribunal_decision_sub_category` validated. Might a better name for this validator be `tribunal_decision_sub_category_relates_to_parent`?

robmckinnon: I can rename the validator class to `TribunalDecisionSubCategoryRelatesToParentValidator`. Then `tribunal_decision_sub_category_relates_to_parent: true` can be used in the call to validates().

As noted in the previous line comment discussion, sub-category validation is being extended based on a meeting with HMCTS. Some parent categories will not have any sub-categories related to them. A tribunal decision with a blank sub-category will be valid if the parent category has no sub-categories. Otherwise presence of the sub-category will be required. The commits for these changes will be added to the pull request soon.

tommyp: `tribunal_decision_sub_category_relates_to_parent` was only a suggestion. The class should be named for how it's being validated. `validates :thing, date: true` is pretty self explanatory. If it's validating that it's related to a parent then `tribunal_decision_sub_category_relates_to_parent` is good. If it's doing more than that, maybe split the logic out into separate validators?

robmckinnon: I can make them separate validators if you think that works. Thought it might be better to leave it as a single validator so all sub-category validation logic is in one file. It gets tricky as the sub-category presence validation is contingent on the value of the parent category, so in a way is part of the same broader validation. Would you still like me to split this into separate validator classes?

robmckinnon: Agreed offline with @tommyp to keep logic in a single validator class for now.

PR #568, opened 2015-09-18
WIP: Scenario outlines for creating and editing specialist documents

robmckinnon:
Let's discuss this on Wed next week. Intention is to make it straight-forward to add new specialist documents, e.g. tribunal decisions, to the features in a consistent manner. Agreed previously we want a single approach to integration testing for all specialist document finders. This seems like a good consolidation.

Changes are:

* Use cucumber scenario outlines to define common steps for [creating and editing specialist documents](https://github.com/alphagov/specialist-publisher/pull/568/files/#diff-6) and [viewing specialist document lists](https://github.com/alphagov/specialist-publisher/pull/568/files/#diff-27).
* Rename some step definitions to match pattern in scenario outline steps.
* Delete the scenarios that are duplicates of the scenario outlines.
* Add scenario examples for viewing specialist document lists for [UTAAC tribunal decisions](https://github.com/ministryofjustice/specialist-publisher/commit/debe5a83fd35846b3cc3d8d89d53c7fd5e20affe) and [Asylum Support tribunal decisions](https://github.com/ministryofjustice/specialist-publisher/commit/fe4a082a4037f8effa85e3a21c404d5f66283a1b).

### DIFF
--- a/features/creating-and-editing-a-cma-case.feature
+++ b/features/creating-and-editing-a-cma-case.feature
@@ -11,10 +11,6 @@ Feature: Creating and editing a CMA case
     When I create another case with the same slug
     Then I see a warning about slug clash at publication
 
-  Scenario: Can view a list of all cases in the publisher
-    Given two CMA cases exist
-    Then the CMA cases should be in the publisher case index in the correct order
-
   Scenario: Change the title of a previously published document
     Given a published CMA case exists
     When I change the CMA case title and re-publish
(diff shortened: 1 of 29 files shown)


robmckinnon:
Quote from an email from @marksheldonGDS regarding this pull request:

> We understand that using outlines is probably a better approach, but would prefer to review this at a later date. We simply don't have the time to do that right now.  We also accept that this means duplicating code which doesn't improve things. However, we have many areas of our stack that need improvement and we've had to prioritise.

PR #571, opened 2015-09-21
Don't index preview_only Finders on Production

tommyp:
In order to prevent Finders going into Elastic Search on Production we use
the `preview_only` flag to decide if it should be published or not. This
commit follows the same logic in the `PublishingApiFinderPublisher` but
keeps them separate. There's nothing to stop these being merged in
future but as this is preventing deploys I thought I'd go down the
simplest route now.

This PR also removes the funtionality which stops Finders being deployed
without a content ID. As we can now restrict them to Preview, there's no
need to keep this feature around.

[Ticket](https://trello.com/c/Nn0DoEN2/120-preview-only-finders-should-not-be-published-to-rummager).

### DIFF
--- a/lib/publishing_api_finder_publisher.rb
+++ b/lib/publishing_api_finder_publisher.rb
@@ -10,7 +10,7 @@ class PublishingApiFinderPublisher
 
   def call
     finders.map do |finder|
-      if finder[:metadata].has_key?("content_id") && !preview_only?(finder)
+      if !preview_only?(finder)
         publish(finder)
       elsif preview_only?(finder)
         if preview_domain_or_not_production?
@@ -18,8 +18,6 @@ class PublishingApiFinderPublisher
         else
           logger.info("didn't publish #{finder[:metadata]["name"]} because it is preview_only")
         end
-      else
-        logger.info("didn't publish #{finder[:metadata]["name"]} because it doesn't have a content_id")
       end
     end
   end
--- a/lib/publishing_api_finder_publisher.rb
+++ b/lib/publishing_api_finder_publisher.rb
@@ -10,14 +10,10 @@ class PublishingApiFinderPublisher
 
   def call
     finders.map do |finder|
-      if !preview_only?(finder)
+      if should_publish_in_this_environment?(finder)
         publish(finder)
-      elsif preview_only?(finder)
-        if preview_domain_or_not_production?
-          publish(finder)
-        else
-          logger.info("didn't publish #{finder[:metadata]["name"]} because it is preview_only")
-        end
+      else
+        logger.info("Not publishing #{finder[:metadata]["name"]} because it is preview_only")
       end
     end
   end
@@ -28,6 +24,10 @@ private
   def publish(finder)
     export_finder(finder)
     export_signup(finder) if finder[:metadata].has_key?("signup_content_id")
+  end
+
+  def should_publish_in_this_environment?(finder)
+    !preview_only?(finder) || preview_domain_or_not_production?
   end
 
   def preview_only?(finder)
--- a/lib/rummager_finder_publisher.rb
+++ b/lib/rummager_finder_publisher.rb
@@ -9,8 +9,14 @@ class RummagerFinderPublisher
 
   def call
     metadatas.each do |metadata|
-      if metadata[:file].has_key?("content_id")
+      if metadata[:file].has_key?("content_id") && !preview_only?(metadata)
         export_finder(metadata)
+      elsif preview_only?(metadata)
+        if preview_domain_or_not_production?
+          export_finder(metadata)
+        else
+          logger.info("didn't publish #{metadata[:file]["name"]} because it is preview_only")
+        end
       else
         # Even though rummager doesn't use the content_id we only want to push
         # to rummager if this is live in content-store, so this needs to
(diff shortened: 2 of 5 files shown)


lib/rummager_finder_publisher.rb
```diff
@@ -9,8 +9,14 @@ def initialize(metadatas, logger: Logger.new(STDOUT))
 
   def call
     metadatas.each do |metadata|
-      if metadata[:file].has_key?("content_id")
+      if metadata[:file].has_key?("content_id") && !preview_only?(metadata)
```
rboulton: Should the `has_key?("content_id")` test also apply to `preview_only?` finders?

If so, I think this would be better done by nesting the preview_only test within the `has_key?("content_id")` test:

```
    if metadata[:file].has_key?("content_id")
      if preview_only?(metadata)
...
      else
....
      end
```

rboulton:
I think the title of this PR may be reversed: it looks like the intention is to index preview_only? finders on preview.  Should the title be "Don't index preview_only Finders on Production"?

tommyp:
@rboulton Yes. You're right. Changed the title.

rboulton:
Nitpick (but important) - the commit message also has the same problem as the title had.

Publishing these finders only in preview seems like a reasonable approach to the problem. I'm not sure if there's a better way to check whether the code is running in a preview environment.  Perhaps having our deployment process set up a configuration value saying whether "preview_only?" finders should be published would be a better approach?

lib/rummager_finder_publisher.rb
jennyd: We should remove the `content_id` check entirely and only use the feature flag - if a finder is merged then it should be fine for it to be published on Preview, and having two mechanisms for preventing publishing is confusing to everyone and complicates the logic here. Since `PublishingApiFinderPublisher` also checks the `content_id` I'd be happy for that to be a separate PR removing that check from both classes, rather than tying that change in behaviour to this fix.

jennyd:
We've got a [separate story](https://trello.com/c/R7iKwQzW) to switch to using a new specific environment variable for this feature flag, rather than relying on `GOVUK_APP_DOMAIN`.

tommyp:
@rboulton @jennyd I've changed the commit message as well and updated the commit to remove the ability to restrict Finders from deploy which don't have a `content_id`.

jennyd:
Could you separate this into 2 commits, one for removing the logic around `content_id` and one for the `RummagerFinderPublisher` changes?

tommyp:
@jennyd Done 😀

lib/rummager_finder_publisher.rb
```diff
@@ -9,20 +9,30 @@ def initialize(metadatas, logger: Logger.new(STDOUT))
...
-      else
-        # Even though rummager doesn't use the content_id we only want to push
-        # to rummager if this is live in content-store, so this needs to
-        # replicate the logic in PublishingApiFinderPublisher.
-        logger.info("didn't publish #{metadata[:file]["name"]} because it doesn't have a content_id")
+      elsif preview_only?(metadata)
+        if preview_domain_or_not_production?
+          export_finder(metadata)
+        else
+          puts "didn't publish #{metadata[:file]["name"]} because it is preview_only" if @log
```
jennyd: We should use `logger` instead of `puts` consistently in this class now (it was only added here recently).

tommyp: And that was me not redoing far enough...

lib/publishing_api_finder_publisher.rb
```diff
@@ -10,7 +10,7 @@ def initialize(finders, logger: Logger.new(STDOUT))
 
   def call
     finders.map do |finder|
-      if finder[:metadata].has_key?("content_id") && !preview_only?(finder)
+      if !preview_only?(finder)
```
jennyd: Could we reverse this so that the positive case comes first? (and the equivalent in the other class...) `if !` is equivalent to `unless` and `unless else` contravenes our [style guide](https://github.com/alphagov/styleguides/blob/master/ruby.md#syntax).

If we'd like to keep the default (not preview-only) case first, we could add another method called something like `publish_everywhere?(finder)` which returns `!preview_only?(finder)` and test that here instead.

mikejustdoit:
Now that we're not using empty `content_id`s, do we need to validate that a Finder's `content_id` isn't empty elsewhere? Like, in the schemas and/or validators.

And does this have any impact on the other apps that interact with finders?

rboulton:
It can't quite be made mandatory in the schemas yet, because some apps are still sending things without content ids.  My team is currently working on fixing a lot of those apps, but I'm not sure we'll get them all done.  Once we're in a position to satisfy it though, we do want to make content_ids mandatory in the schemas, at which point hopefully schema tests will help ensure that they're not empty.

tommyp:
@mikejustdoit The `content_id` is [called](https://github.com/alphagov/specialist-publisher/blob/master/app/presenters/finder_content_item_presenter.rb#L33) using `fetch`, so it will raise an exception if it isn't present. This is the same behaviour that we have for the other fields that the Publishing API uses at the root level of the hash.

mikejustdoit:
All right, I'm happy with this, then :+1:

PR #572, opened 2015-09-22
Bugfix: Duplicate document creation on validation errors

boffbowsh:
`updated_at` doesn't seem to be set correctly on the document
instance. The document is really just a wrapper around a
collection of editions, so to check if a document is persisted,
we need to check if any of the editions are persisted.

This fixes a bug where `persisted?` was returning a false negative
when a document was invalid, causing the form to post to the
`#create` method. This caused duplicate documents to be created,
which would then be invalid due to the duplicate slug.

/cc @jennyd @tommyp @mikejustdoit

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -54,7 +54,7 @@ GEM
     byebug (2.7.0)
       columnize (~> 0.3)
       debugger-linecache (~> 1.2)
-    capybara (2.2.1)
+    capybara (2.5.0)
       mime-types (>= 1.16)
       nokogiri (>= 1.3.3)
       rack (>= 1.0.0)
--- a/app/models/specialist_document.rb
+++ b/app/models/specialist_document.rb
@@ -145,6 +145,10 @@ class SpecialistDocument
     edition.save
   end
 
+  def persisted?
+    editions.any?(&:persisted?)
+  end
+
 protected
 
   attr_reader :slug_generator, :edition_factory
--- a/app/view_adapters/document_view_adapter.rb
+++ b/app/view_adapters/document_view_adapter.rb
@@ -15,7 +15,7 @@ class DocumentViewAdapter < SimpleDelegator
   end
 
   def persisted?
-    document && document.updated_at
+    document && document.persisted?
   end
 
   def to_param
--- a/features/support/aaib_report_helpers.rb
+++ b/features/support/aaib_report_helpers.rb
@@ -1,6 +1,6 @@
 module AaibReportHelpers
-  def create_aaib_report(*args)
-    create_document(:aaib_report, *args)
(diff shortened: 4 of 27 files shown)


mikejustdoit:
Unless there're any objections, I wouldn't mind reviewing this.

I'm wondering whether I should try to quickly investigate whether there's a root cause of this that the PR doesn't address? If you don't think that's within the scope of the review, just say.

boffbowsh:
This is the root cause. I'm not sure what you mean, @mikejustdoit?

mikejustdoit:
Sorry, yeah, I didn't word that as clearly as I could...

I was just considering whether there's a bug closer to the model layer than the place you've introduced the fix.

I guess I don't understand the difference between a valid and non-valid document. Is it obvious why one is "persisted" and the other isn't?

boffbowsh:
Persisted in this instance means "does the document (not the edition) exist in the database?" As there isn't an actual document, this really means "Are any editions with this document_id persisted in the database?"

The reason `persisted?` is the cause of the problem is that form_for uses it to work out whether it's updating an existing object or creating a new one. If `persisted?` is false, the object in question is a new object and should be created in the database using a POST to the collection path. If `persisted?` is true, the object already exists and should be updated with a PUT to the object path.

mikejustdoit:
I guess my point is that the fix to `persisted?` here is in a view adapter, which serves only a small part of the application. If there are other places where this logic is depended on, they might also be broken or at least in a precarious state.

To my knowledge, this bug has only been reported within the past week, so if you hadn't looked already, I was considering the possibility it was introduced recently.

I'm fairly familiar with the Document-Edition relationship and `form_for`, but I don't yet understand how that relates to validations.

boffbowsh:
The bug has been in existence for quite a while, at least a month. @tommyp has more context. This is a view adapter, correct, but it's being used as a model as far as `form_for` is concerned.

The validations part of this is more of a symptom than a cause. I don't know what `SpecialistDocument#updated_at` is supposed to return, but it's clear it's being used incorrectly here. It definitely shouldn't be used to provide the persistence status.

I've updated and rebased to define `persisted?` on the document, is that clearer?

mikejustdoit:
The context I heard from Tommy is what gave me the impression it's only been recently reported :grinning: 

Thanks, moving the references to `editions` was a good move. I think as much as possible, we try not to handle editions outside of the repository/models.

If you've got any more information about the relationship between `updated_at` and validations, I'd be interested to hear. It seems like there are some unanswered questions there.

tommyp:
To give some context around Mike's comments, I was thinking that [this](https://github.com/alphagov/specialist-publisher/blob/master/app/models/specialist_document.rb#L215-L222) may have had something to do with it.

PR #576, opened 2015-09-28
Feature flag for pre-production finders

mikejustdoit:
Finders that shouldn't be published in production environments are known as "pre-production finders". (The corresponding metadata attribute was recently renamed from `preview_only` to `pre_production` https://github.com/alphagov/specialist-publisher/pull/575.)

We're currently checking the environment variable `GOVUK_APP_DOMAIN` _and_ `Rails.env`, but a dedicated environment (`PUBLISH_PRE_PRODUCTION_FINDERS`) variable has just been introduced to the Preview and Development environments via Puppet (https://github.gds/gds/puppet/pull/3446) and here we've made the move to make use of that.

In this change, for both uses of the _feature flag_ we're accessing `ENV["PUBLISH_PRE_PRODUCTION_FINDERS"]` directly, but it might be cleaner to move it into application config (an initialiser). Opinions welcome.

### DIFF
--- a/config/initializers/publish_pre_production_finders_feature_flag.rb
+++ b/config/initializers/publish_pre_production_finders_feature_flag.rb
@@ -1,0 +1,1 @@
+SpecialistPublisher::Application.config.publish_pre_production_finders = ENV["PUBLISH_PRE_PRODUCTION_FINDERS"].present?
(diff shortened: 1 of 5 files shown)


config/initializers/publish_pre_production_finders_feature_flag.rb
```diff
@@ -0,0 +1 @@
+SpecialistPublisher::Application.config.publish_pre_production_finders = ENV["PUBLISH_PRE_PRODUCTION_FINDERS"].present?
```
tommyp: I like this. I'm going to steal this idea of moving all `ENV['VAR']` usage to an initializer :+1:

PR #579, opened 2015-09-30
Use environment in rake publish tasks

tommyp:
Without ` => :environment` after the name of each rake task, it fails to
load the environment. This commit adds it, allowing for the rake task
to load the Rails environment.


benlovell:
This isn't for environment variables, rather it loads the rails environment.

tommyp:
@benlovell True. I've amended the commit and the PR.

mikejustdoit:
Good find :+1:

PR #580, opened 2015-10-07
Add tax tribunal decisions finder

robmckinnon:
Add a finder for Upper Tribunal (Tax and Chancery Chamber) decisions.

Upper Tribunal (Tax and Chancery Chamber) publishes decisions that describe the outcome of a tribunal case. HMCTS may decide to also publish First-tier Tribunal (Tax) decisions using this same finder. The base_path slug has been named to be flexible enough to allow for this.

Tribunal decision finders are used by public, professionals and the tribunals themselves to search for past decisions.

Published decisions form a part of case law and are used for research purposes and to prepare for current or future cases.

**Requires this rummager pull request be merged**: https://github.com/alphagov/rummager/pull/526

### DIFF
--- a/app/models/tax_tribunal_decision.rb
+++ b/app/models/tax_tribunal_decision.rb
@@ -1,0 +1,9 @@
+require "document_metadata_decorator"
+
+class TaxTribunalDecision < DocumentMetadataDecorator
+  set_extra_field_names [
+    :hidden_indexable_content,
+    :tribunal_decision_category,
+    :tribunal_decision_decision_date,
+  ]
+end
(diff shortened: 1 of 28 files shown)


app/models/tax_tribunal_decision.rb
```diff
@@ -0,0 +1,9 @@
+require "document_metadata_decorator"
+
+class TaxTribunalDecision < DocumentMetadataDecorator
+  set_extra_field_names [
+    :hidden_indexable_content,
+    :tribunal_decision_category,
+    :tribunal_decision_decision_date
```
tommyp: Can you add a trailing comma here in accordance with the [styleguide](https://github.com/alphagov/styleguides/blob/master/ruby.md#collections)?

robmckinnon: will do

PR #581, opened 2015-10-14
Add new MHRA alert types

boffbowsh:
https://trello.com/c/tiCatwpE/117-add-two-new-alert-types-to-mhra-drug-device-alerts-finder-small

MHRA are removing these remaining alerts away from publications to a finder

Related:
https://github.com/alphagov/govuk-content-schemas/pull/133
https://github.com/alphagov/rummager/pull/530

### DIFF
--- a/finders/metadata/medical-safety-alerts.json
+++ b/finders/metadata/medical-safety-alerts.json
@@ -33,6 +33,20 @@
       "topic_name": "drug alerts",
       "body": "information published by the MHRA about defective medicines.",
       "prechecked": true
+    },
+    {
+      "key": "field-safety",
+      "radio_button_name": "Field safety notices",
+      "topic_name": "field safety notices",
+      "body": "information from medical device manufacturers.",
+      "prechecked": true
+    },
+    {
+      "key": "company-led-drug",
+      "radio_button_name": "Drug alerts: company-led",
+      "topic_name": "drug alerts: company-led",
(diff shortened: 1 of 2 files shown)


boffbowsh:
@elliotcm can you clarify what the email flow is here please? If I remember correctly, the topics will be created on-demand as people subscribe to them, and filtering will just be done by the `alert_type`?

elliotcm:
@boffbowsh Specialist publisher is a :snowflake: and I wasn't involved in writing the email stuff it does.  It talks directly to the email alert API and doesn't go via the email alert service.  @tommyp might know.

tommyp:
Hello! The email filter [are defined in the metadata](https://github.com/alphagov/specialist-publisher/blob/master/finders/metadata/medical-safety-alerts.json#L22). Specialist Publisher doesn't use Email Alert Frontend. We should fix this soon™.

boffbowsh:
Cool, thanks. So what I've done is sufficient?

tommyp:
User won't be able to sign up to alerts marked with the new alert types. Is that what we want?

boffbowsh:
In my testing, the email signup form for the finder https://www.gov.uk/drug-device-alerts/email-signup will get the two new types as options.

tommyp:
Sorry. Yes you're right. Ignore my previous comment.

PR #590, opened 2015-10-26
Add script to import MHRA alerts from whitehall

boffbowsh:
This is based on a resurrection of https://github.com/alphagov/specialist-publisher/pull/300.

This is used for importing Field Safety Notices and Company-led Drug
Alerts from their respective Whitehall publications into the
`/drug-device-alerts` finder (which is confusingly the
`medical_safety_alert` document type). A related Whitehall PR
(https://github.com/alphagov/whitehall/pull/2364) is used to export the
HTML attachments as a JSON file, each of which should be supplied to
the importer. The filename is used to determine the alert type to
import the document as.

The `issued_date` is parsed explicitly from the document title. This is
not ideal but works well with the current set of documents we have.
From this, the `public_updated_at` date is set before publish, in order
to ensure the alerts appear in the correct order in the finder. This is
done by prepending the `public_updated_at_setter` observer in a similar
pattern to the `publish_cma_cases` script.

Once this import has been completed in production, it’s expected that
we’ll revert this commit in much the same way as the initial imports
occurred, so please forgive the inclusion of `2015` in the date parsing
regex.

/cc @jamiecobbett @benilovj

### DIFF
--- a/app/importers/document_import.rb
+++ b/app/importers/document_import.rb
@@ -24,7 +24,8 @@ module DocumentImport
     end
 
     def success(document, data)
-      write("SUCCESS", document.import_notes.join("\t"), data)
+      import_notes = document.respond_to?(:import_notes) ? document.import_notes : []
+      write("SUCCESS", import_notes.join("\t"), data)
     end
 
     def failure(document, data)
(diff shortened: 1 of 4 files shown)


benlovell:
It's like Java, but Ruby! :speak_no_evil:

boffbowsh:
Also I guess @tommyp is well placed to review this. Sorry buddy.

boffbowsh:
Adding something here to output the before and after slugs so we can add redirects later.

tommyp:
LGTM 👍

PR #591, opened 2015-10-27
Output old and new slugs on stdout for import

boffbowsh:
When the import process is run, the old documents need withdrawing
in Whitehall. In order to do this, we need to know both the old and
new URLs.

This doesn't output them in a very nice way, but it's only a small
number of documents affected, so a manual process is fine.

/cc @tommyp

### DIFF
--- a/app/importers/medical_safety_alert_import/mapper.rb
+++ b/app/importers/medical_safety_alert_import/mapper.rb
@@ -9,11 +9,8 @@ module MedicalSafetyAlertImport
     def call(raw_data)
       document = document_creator.call(desired_attributes(raw_data))
       document = document_publisher.call(document.id)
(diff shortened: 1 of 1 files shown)


app/importers/medical_safety_alert_import/mapper.rb
```diff
@@ -9,13 +9,10 @@ def initialize(document_creator, document_publisher, repo)
     def call(raw_data)
       document = document_creator.call(desired_attributes(raw_data))
       document = document_publisher.call(document.id)
+      puts "#{raw_data["slug"]} => #{document.slug}"
       document
     end
 
-    def import_notes
```
tommyp: Did you mean to remove this method too?

boffbowsh: Yep, turns out it's not needed.

PR #599, opened 2015-11-11
Add GDS SSO & Mongo

tommyp:
This PR adds GDS SSO to Specialist Publisher V2. As we need a data store in which to hold the User model, this PR also adds Mongo (through Mongoid) as a data store. We've used Mongo with Master so it seemed easiest to go with it again.


benilovj:
It'd be good to use the [`gds-sso` lint specs](https://github.com/alphagov/gds-sso/blob/master/lib/gds-sso/lint/user_spec.rb) to make sure that the `User` model does what `gds-sso` expects.

PR #603, opened 2015-11-16
New documents

tommyp:
This PR contains the work to allow Editors to create new Documents. So far we've only added `CMA Case` as a format, but it's set up in such a way that adding new formats should be pretty painless. The interesting stuff is going on in `Document` and `CmaCase` models. We do some metaprogramming to set up the `attr_accessor`s for the format specific stuff. 

We've added an Integration test to make sure new Documents can be created. The models are light on logic it doesn't make much sense to have a unit test.

I've also added the views for the form, the frontend toolkit for colours and `select2` for the multi-selects.

This PR requires https://github.com/alphagov/gds-api-adapters/pull/388 to be merged and the version bumped here in order to pass. [Ticket](https://trello.com/c/64GwuCnc/2-create-new-cma-case).


app/controllers/documents_controller.rb
```diff
@@ -1,7 +1,58 @@
...
+  def filtered_params(params_of_document)
+    # TODO: Make this work like the ManualsController parameter filtering
+    # We shouldn't make our hashes indifferent. Let's make the keys consistently symbols
+    filter_blank_multi_selects(params_of_document).with_indifferent_access
+  end
+
+  # See http://stackoverflow.com/questions/8929230/why-is-the-first-element-always-blank-in-my-rails-multi-select
+  def filter_blank_multi_selects(values)
+    values.reduce({}) { |filtered_params, (key, value)|
+      filtered_value = value.is_a?(Array) ? value.reject(&:blank?) : value
```
benlovell: You could golf this down a little by just normalising `value` like `Array(value).reject(&:blank?)`

tommyp: We don't want to turn all values into Arrays here though, we only want to filter out the blank options that always come through from the multi-selects.

app/controllers/application_controller.rb
```diff
@@ -7,21 +7,29 @@ class ApplicationController < ActionController::Base
...
-        title: "CMA Cases",
-      ),
+    data = {
+      "CMA Case" => CmaCase
     }
-  end
 
-  def current_finder
-    document_types.fetch(params.fetch(:document_type, nil), nil)
+    data.map do |k, v|
```
danielroseman: This is sufficiently opaque that we probably need a comment showing the intended output.

app/controllers/documents_controller.rb
```diff
@@ -1,7 +1,56 @@
...
     redirect_to "/#{document_types.keys.first}" unless params[:document_type]
   end
 
+  def new
+    render :new, locals: { document: current_format.klass.new }
+  end
+
+  def create
+    document = current_format.klass.new(
+      filtered_params(params[:"#{current_format.format_name}"])
```
danielroseman: Since filtered_params returns a HashWithIndifferentAccess, you could just use a string key which would be a bit clearer.

app/controllers/documents_controller.rb
```diff
@@ -1,7 +1,56 @@
...
   def index
     redirect_to "/#{document_types.keys.first}" unless params[:document_type]
   end
 
+  def new
+    render :new, locals: { document: current_format.klass.new }
+  end
+
+  def create
+    document = current_format.klass.new(
```
danielroseman: Seems like `current_format.klass.new` could be extracted into a method (`document` perhaps?) since you use it here and above.

app/controllers/documents_controller.rb
```diff
@@ -1,7 +1,56 @@
...
+    request = publishing_api.put_content(document.content_id, presented_document.to_json)
+
+    if request.code == 200
+      flash[:success] = "Created #{document.title}"
+      redirect_to documents_path(document_type: current_format.document_type)
+    else
+      errors = request.body.error.fields.map { |k, v|
+        "#{k}: #{v}"
+      }
+      flash[:error] = errors.join(', ')
```
danielroseman: Presumably these are going to be put into the model errors hash at some point.

app/controllers/documents_controller.rb
tommyp: I think it's necessary to be clear when we're talking about an instance of it and the Class. What about `document_klass`?

app/models/cma_case.rb
```diff
@@ -0,0 +1,29 @@
...
+    "cma_case"
+  end
+
+  def public_path
+    "/cma-cases"
+  end
+
+  def organisations
+    # Org name: Competition And Markets Authority
+    ["957eb4ec-089b-4f71-ba2a-dc69ac8919ea"]
```
danielroseman: We shouldn't be hard-coding UUIDs anywhere. Previously the organisations were queried from the organisations API; we need to either continue doing this or find some other way to get the data dynamically.

app/models/document.rb
```diff
@@ -0,0 +1,55 @@
+class Document
+  include ActiveModel::Model
+
+  attr_accessor :content_id, :title, :summary, :body, :format_specific_fields, :state, :bulk_published, :minor_update, :change_note
+
+  def initialize(params = {}, format_specific_fields = [])
```
danielroseman: I'm not sure why format_specific_fields needs to be a parameter to the initialize method. They are available on the object already, you can just refer to them directly. There probably needs to be a default declaration of `FORMAT_SPECIFIC_FIELDS = []` on the base Document class which is overridden as you have done in CmaCase.

app/models/cma_case.rb
tommyp: Suggestions? Should I use the content register for this?

app/controllers/application_controller.rb
tommyp: How's this? https://github.com/alphagov/specialist-publisher/commit/84134b4489946636d8127b96758597b907856a8e

app/presenters/document_presenter.rb
```diff
@@ -0,0 +1,56 @@
...
+      },
+      routes: [
+        {
+          path: document.base_path,
+          type: "exact",
+        }
+      ],
+      redirects: [],
+      update_type: "major",
+    }.merge(links_attributes)
```
danielroseman: Links are not part of the content item and should be sent separately to the links endpoint.

app/presenters/document_presenter.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+private
+
+  attr_reader :document
+
+  def metadata
+    document.format_specific_fields.map do |f|
+      {
+        :"#{f}" => document.send(f)
```
danielroseman: format_specific_fields are already symbols, so no need to do the string substitution

app/views/metadata_fields/_cma_cases.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= f.label :opened_date %>
+<%= f.text_field :opened_date, placeholder: '2012-04-23', class: 'form-control' %>
+<%= f.label :closed_date %>
+<%= f.text_field :closed_date, placeholder: '2013-10-19', class: 'form-control' %>
+<%= f.label :case_type %>
+<%= f.select :case_type, f.object.facet_options(:case_type), {}, { class: 'form-control' } %>
```
danielroseman: Might be nice to write a helper method to get the individual facet options.

spec/features/creating_a_cma_case_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    fill_in "Opened date", with: "2014-01-01"
+    select "Energy", from: "Market sector"
+
+    click_button "Save as draft"
+
+    assert_publishing_api_put_item("4a656f42-35ad-4034-8c7a-08870db7fffe", request_json_including(cma_case_content_item))
+
+    expect(page.status_code).to eq(200)
+  end
+end
```
danielroseman: Need a spec for displaying errors on invalid document.

app/controllers/application_controller.rb
```diff
@@ -7,21 +7,48 @@ class ApplicationController < ActionController::Base
...
+      {
+        k.downcase.parameterize.pluralize => OpenStruct.new(
```
h-lame: It'd be good to harden this down to an actual object if we know that `klass`, `document_type`, `format_name`, and `title` are the full interface.  [OpenStruct is problematic](http://blog.honeybadger.io/how-openstruct-and-hashes-can-kill-performance/) from a performance POV.  Of course if we're still sketching out the interface then OpenStruct is fine, but it would be good to remember to replace when we're ready.

tommyp: What about something like `FormatStruct = Struct.new(:klass, :document_type, :format_name, :title)`?

h-lame: :+1: great success.

fofr:
Are there some screenshots showing how the UI looks?

app/models/cma_case.rb
danielroseman: Content register doesn't have a get endpoint for a specific entry, so it's not suitable here. Why can't you continue to use the existing organisations API?

app/controllers/documents_controller.rb
danielroseman: Sounds good.

app/models/cma_case.rb
tommyp: So we're not referring to pages with `base_path`s anymore, but we're still referring to Orgs with slugs?

danielroseman: Good point, we would want the content-id. But looking at the current code and what's in content store, it doesn't seem that specialist docs send any links at all. Does this actually need to happen?

tommyp: Ah yes. You are correct. We probably should send Orgs though.

tommyp: So it gets the Organisations from the Finder content item, but we probably want to send them with the individual Document because we're looking to phase out this behaviour.

tommyp:
@fofr It's pretty much the exact same as the current one.

app/models/cma_case.rb
rboulton: It would be good to talk to finding-things about this - we're planning on making organisation tagging possible centrally using the new content-tagger app (turning it on format-by-format as we make the publisher apps compatible), but setting a sensible default is important.  It seems most likely to me that copying the org links from the associated finder content item is the best approach for new specialist documents.

tommyp:
## 🎉 ALRIGHT PULL REQUEST PEOPLE 🎉

I've either amended commits or added new ones pertaining to everyone's comments. Have another look please and someone hit merge so I can get stuck into rebase hell this afternoon.

PR #604, opened 2015-11-23
Document index and show

tommyp:
This PR allows users to see a list of created Documents in the Publishing API V2 for a format (currently only CMA Cases) and to view the show page. The interesting commits are probably [this](https://github.com/alphagov/specialist-publisher/commit/3afd35ba6dc8683207af74a59e62a952b6ae078d), [this](https://github.com/alphagov/specialist-publisher/commit/3cd07eb0c365097c2d72a49817ca2baf18272cde) and [this](https://github.com/alphagov/specialist-publisher/commit/6761608a46ce62178210e7753fbd0034753014ea).

Requires https://github.com/alphagov/specialist-publisher/pull/603 to be merged first.


app/models/document.rb
```diff
@@ -46,6 +54,51 @@ def facet_options(facet)
     finder_schema.options_for(facet)
   end
 
+  def format_specific_metadata
```
elliotcm: This method is right confusing.

tommyp: I could put it in a `Wiring` file? :trollface: 

For real it is. I'll add some explanation as to what it does.

app/models/document.rb
```diff
@@ -46,6 +54,51 @@ def facet_options(facet)
...
+
+    document
+  end
+
+  def public_updated_at
+    @public_updated_at ||= Time.zone.now
+  end
+
+  def public_updated_at=(timestamp)
+    public_updated_at = Time.parse(timestamp)
```
elliotcm: This means that the getter outputs a different format than the setter accepts.

app/views/documents/show.html.erb
```diff
@@ -0,0 +1,42 @@
...
+          <dd><time><%= values.to_s(:govuk_date) %></time></dd>
+        <% else %>
+          <% Array(values).each do |value| %>
+            <dd><%= truncate(value.to_s, length: 140) %></dd>
+          <% end %>
+        <% end %>
+      <% end %>
+
+      <dt>Publication state</dt>
+      <dd><%# state(document) %></dd>
```
elliotcm: Should this be real code?

app/views/documents/index.html.erb
```diff
@@ -1,4 +1,28 @@
...
+          <ul class="metadata">
+            <li class="text-muted">Updated <%= time_ago_in_words(document.public_updated_at) %> ago</li>
+            <li>
+              <%= document.publication_state %>
+            </li>
+          </ul>
+        </li>
+      <% end %>
+    </ul>
+    <%# paginate documents %>
```
elliotcm: Let's shed this if we're not implementing it yet.

lib/finder_schema.rb
```diff
@@ -23,6 +23,11 @@ def humanized_facet_value(facet_key, value, &block)
     end
   end
 
+  def humanized_facet_name(key, &block)
+    facet_data_for(key).fetch("name", &block)
+  end
+
+
```
elliotcm: Whitespace.

spec/features/viewing_a_specific_cma_case_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'spec_helper'
+
+RSpec.feature "Creating a CMA case", type: :feature, js: true do
```
elliotcm: This feature name isn't right.

app/models/document.rb
elliotcm: If you can think of a way to break it up so it explains itself that would be best, otherwise yeah a comment would be good stuff.

h-lame: This is nitpicky, and I don't know if it solves the problem totally, but at the very least, for consistency with `humanized_attributes` it'd be good to use `inject`*.  I reckon the `.map { ... }.reduce({}, :merge)` pattern is fairly confusing.

*`.each.with_object({})` in both might be even clearer though.

tommyp:
Closing in favour of https://github.com/alphagov/specialist-publisher/pull/606.

tommyp:
I fixed all the commits in #606 but can't get CI to run the suite due to Capybara not starting. Feel free to comment on that PR with more comments.

PR #605, opened 2015-11-27
Add script and class to handle manual section reslugging

davidbasalla:
We add a script that calls a class for reslugging manual sections. We first withdraw the old section which places a `gone_item` into the content-store. Then we change the slug of the manual section in the local database. Finally we republish the manual which the section belongs to. This will republish all its sections, but will check to only publish the updated section.

The workflow is inspired by https://github.gds/pages/gds/opsmanual/2nd-line/changing-specialist-publisher-document-slug.html

After running this script in production, a change to router-data will also need to be made (see docs above). UPDATE: Instead of separate PR in router-data, we send a redirect-item to the content-store within this PR

### DIFF
--- a/bin/reslug_manual_section
+++ b/bin/reslug_manual_section
@@ -1,0 +1,44 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require "manual_section_reslugger"
+require "logger"
+
+def usage
+  $stderr.puts %Q{
+USAGE: #{File.basename(__FILE__)} (manual_slug, old_manual_section_slug, new_manual_section_slug)
+
+manual_slug:
+  slug of manual (eg 'guidance/countryside-stewardship-manual')
+
+old_manual_section_slug:
+  current slug of section to be renamed (eg '8-terms-and-conditions')
+
+new_manual_section_slug:
+  new slug for section (eg '8-scheme-requirements-and-procedures')
+
+}
+  exit(1)
+end
+
+def logger
+  @logger ||= begin
+    logger = Logger.new(STDOUT)
+    logger.formatter = Logger::Formatter.new
+    logger
+  end
+end
+
+usage unless ARGV.count >= 3
+manual_slug = ARGV[0]
+current_slug = ARGV[1]
+new_slug = ARGV[2]
+
+logger.info "Renaming manual document slug"
+
+begin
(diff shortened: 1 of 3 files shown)


h-lame:
This looks good. I wonder though, rather than requiring a change to router-data, which we'd need to do every time we ran this script, would it be possible to publish a redirect item in the content-store instead of a gone item for the old slug?

bin/reslug_manual_section
```diff
@@ -0,0 +1,44 @@
...
+USAGE: #{File.basename(__FILE__)} (manual_slug, old_manual_section_slug, new_manual_section_slug)
+
+manual_slug:
+  slug of manual (eg 'guidance/countryside-stewardship-manual')
+
+old_manual_section_slug:
+  current slug to be renamed (eg '8-terms-and-conditions')
+
+new_manual_section_slug:
+  current slug to be renamed (eg '8-scheme-requirements-and-procedures')
```
jennyd: This is the same description as `old_manual_section_slug` - maybe something like "new slug for the section" would be clearer?

davidbasalla:
thanks @jennyd, have fixed and rebased

jennyd:
This mostly looks good - I've tried it out on Preview, and the slug change and redirect works.

The only thing which isn't ideal is that the manual republishing is done as a major update in the case of the section we're doing this for (because it's modifying the existing published edition), which creates an empty change note on the updates page. If the manual is tagged to a topic then anyone subscribed to it would get an email, which could be confusing with the empty change note.

We've discussed this and @davidbasalla is looking into creating a new edition of the section to use for the slug change which can then be marked as a minor update, rather than using the update type of the existing last edition.

davidbasalla:
@jennyd Have just pushed a commit to address your last comment. What do you think about leaving the new edition as a major update and adding a change note to explain the slug update? Might be clearer for anyone looking at the history of the section in question?

h-lame:
Not super familiar with the internals of specialist publisher, but this seems to do the required edition wrangling as best I can tell.  Good stuff!

davidbasalla:
I have just updated this. We now post the redirect items after republishing the manual. We also update the way we republish the manual (by using the `publish` method rather than `queue_publish`)

jennyd:
:green_apple: :space_invader: :sparkles:

PR #607, opened 2015-12-02
Manuals: populate links.organisations in Publishing API

benilovj:
Before this change, the Publishing API organisations information
doesn't contain the content ID and is in the details hash.

Every format now supports links.organisations which is the more
generic solution for specifying organisations.


benilovj:
I've checked that this works on preview.

jamiecobbett:
:+1:

benilovj:
Unfortunately I forgot to do the same for manual sections. A follow-up PR is coming shortly.

PR #613, opened 2015-12-09
Send topic tags when republishing manuals

benilovj:
Before this change, when manuals were being republished, they were
no longer tagged to the topics that they were tagged to previously.

This change mirrors what happens in `PublishManualService`.

Before merging:

- [x] QA this change (on preview)

/cc @tommyp @jennyd

### DIFF
--- a/spec/services/republish_manual_service_spec.rb
+++ b/spec/services/republish_manual_service_spec.rb
@@ -1,0 +1,72 @@
+require "fast_spec_helper"
+require "republish_manual_service"
+
+RSpec.describe RepublishManualService do
+  let(:manual_id) { double(:manual_id) }
+  let(:manual_repository) { double(:manual_repository) }
+  let(:listener) { double(:listener) }
+  let(:listeners) { [listener] }
+  let(:manual) { double(:manual, id: manual_id, slug: "abc") }
+  let(:tag_fetcher) { double(:tag_fetcher) }
+
+  subject {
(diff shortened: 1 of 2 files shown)


spec/services/republish_manual_service_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+            type: "specialist_sector",
+          ),
+          slug: "government-digital-guidance/content-publishing",
+        )
+      ]
+    )
+    allow(TagFetcher).to receive(:new).and_return(tag_fetcher)
+  end
+
+  it "publishes the manual" do
```
tommyp: Wouldn't it be `republishes` the manual? It might be an idea for this test to also check that a draft Manual doesn't get published.

benilovj: I've fixed both issues and force-pushed.

tommyp:
:+1:

benilovj:
I've pushed this branch to preview and verified that republishing still works.

However, I wasn't able to reproduce the original error of tags going missing after republishing – do we definitely know that they get lost?

benilovj:
Ok, I wasn't reproducing the original bug correctly. If the topic tags are already in the Specialist Publisher DB, they are sent correctly, but if the DB didn't know about them (eg if the topics were added in Panopticon), they'd be lost. This fix is still valid and is now ready to be reviewed/merged.

PR #615, opened 2015-12-11
Publish a document

tommyp:
This PR allows editors to publish documents to the Publishing API and Rummager. The most interesting commits are:
- https://github.com/alphagov/specialist-publisher/commit/921378426bff8f9ad9446cf03fc2800cebb09a2b
- https://github.com/alphagov/specialist-publisher/commit/5ad6c0a11f10d41d90e172398779501224047a46
- https://github.com/alphagov/specialist-publisher/commit/977702f4a13f4191a170c318d14ecec2d7a07e23
- https://github.com/alphagov/specialist-publisher/commit/c8f192dbddb1327140a637427a722a097b24840d

There's some hairy metaprogramming going on but I need to (re)construct the Objects in 2 different ways. One from the params and one from the payload. I think it's a pretty clean solution and it keeps the weirdness to deserialisation method.

Following on from this work I want to extract all these methods to provide a more Active Record like interface to the Publishing API interaction that the Document Object has to do. [Ticket](https://trello.com/c/JSFw4pRD/429-publish-a-document).

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -82,12 +82,16 @@ class DocumentsController <  ApplicationController
   def publish
     indexable_document = SearchPresenter.new(@document)
 
-    publish_request = publishing_api.publish(params[:content_id], "major")
-    rummager_request = rummager.add_document(
-      @document.format,
-      @document.base_path,
-      indexable_document.to_json,
-    )
+    begin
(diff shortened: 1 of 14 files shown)


tommyp:
@tadast Handled in https://github.com/alphagov/specialist-publisher/commit/91b91862d058565769e920e0f589e92080585f10

PR #616, opened 2015-12-15
Always republish manual sections to Publishing API

benilovj:
Currently, republishing a manual won't republish the manual sections to Publishing API unless they're marked as `needs_exporting?`. This means that it's not possible to update the Publishing API payloads for manual sections without DB changes, which feels wrong.

Questions about this:

- is this the right way to implement this feature?
- how should this be done within the Specialist Publisher rewrite?

/cc @tommyp @jennyd @jamiecobbett

### DIFF
--- a/app/observers/manual_observers_registry.rb
+++ b/app/observers/manual_observers_registry.rb
@@ -44,7 +44,7 @@ class ManualObserversRegistry
 
 private
   def publication_logger
-    ->(manual) {
+    ->(manual, _: nil) {
       manual.documents.each do |doc|
         next unless doc.needs_exporting?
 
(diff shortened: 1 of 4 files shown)


app/services/republish_document_service.rb
```diff
@@ -17,7 +17,7 @@ def call
   attr_reader :document_repository, :listeners, :document_id
 
   def notify_listeners
-    listeners.each { |l| l.call(document) }
+    listeners.each { |l| l.call(document, "republish") }
```
jamiecobbett: I'd use a symbol, rather than a string - it shouldn't be mutated.

jamiecobbett:
This looks totally sensible to me, what do you think @tommyp?

tommyp:
Other than @jamiecobbett's comment it looks fine. As for how this will work in the new, new world we can contain this to the `publish` method.

benilovj:
I've written an integration (feature) test, which exposed a bug :flushed: which is now fixed. This PR is ready for another review /cc @jamiecobbett @tommyp

PR #619, opened 2015-12-18
Activerecord like interface

tommyp:
One of the main issues when working with the existing Specialist Publisher is that when trying to do anything via the console the App doesn't work like a standard Rails (or Ruby) App.

This PR gives document formats in Specialist Publisher an ActiveRecord like interface to the Publishing API. More methods may come later but these are the ones I needed to add to get to where I am now.

- `CmaCase#all`: Fetches all payloads and returns all instances of that format.
- `CmaCase#find(content_id)`: Fetches the payload for the content item with a matching `content_id` and returns an instance of the matching format.
- `CmaCase.save!`: Saves the Document to the Publishing API with the current properties.
- `CmaCase.publish!`: Sends the publish command to the Publishing API publishing the current payload and the current version to Rummager.

[Ticket](https://trello.com/c/7cia9ASC/467-active-record-style-interface).

This PR also adds Timecop as a more reliable way of predicting `public_updated_at` in tests.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -30,7 +30,11 @@ class Document
   end
 
   def format
-    "document"
+    raise NotImplementedError
+  end
+
+  def self.format
+    raise NotImplementedError
   end
 
   def phase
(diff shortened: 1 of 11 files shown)


app/models/document.rb
```diff
@@ -103,13 +107,81 @@ def public_updated_at
   end
 
   def public_updated_at=(timestamp)
-    @public_updated_at = Time.parse(timestamp) unless timestamp.nil?
+    @public_updated_at = Time.parse(timestamp.to_s) unless timestamp.nil?
   end
 
-private
+  # ACTIVE RECORD LIKE INTERFACE
```
jamiecobbett: I would either delete this comment or if it does feel like a separate concern, think about pulling it into a mixin.

elliotcm: Agreed.

tommyp: Removed the comment.

PR #622, opened 2016-01-07
Publish finders to V2

tommyp:
This PR adds the rake task and related Classes and Presenters that allow us to publish Finders to the Publishing API V2. Some of this code has changed from `master` to reflect the changes in the Publishing API. I also took this opportunity to refactor the 2 separate files per Finder into 1 file, which should make it easier to add new Finders. [Ticket](https://trello.com/c/jMruKoC3/448-publish-finders).


tuzz:
Personally, I'm a little weary of using `Struct` as it can sometimes lead to bugs at runtime, which could have been caught earlier, e.g.

```ruby
class Foo < Struct.new(:bar)
end

Foo.new
```

```ruby
class Foo
  attr_accessor :bar
  def initialize(bar)
    self.bar = bar
  end
end

Foo.new
# ArgumentError: wrong number of arguments (0 for 1)
```

That being said, the latter if a bit more clunky and it's up to you.

This may be relevant: http://thepugautomatic.com/2013/08/struct-inheritance-is-overused/

lib/publishing_api_finder_publisher.rb
```diff
@@ -0,0 +1,80 @@
+require "gds_api/publishing_api"
+require_relative "../app/presenters/finders/finder_content_item_presenter"
+require_relative "../app/presenters/finders/finder_links_presenter"
+require_relative "../app/presenters/finders/finder_signup_content_item_presenter"
+require_relative "../app/presenters/finders/finder_signup_links_presenter"
```
tuzz: These could probably just be `require`s if added to the Rails autoload path, e.g.

https://github.com/alphagov/publishing-api/blob/master/config/application.rb#L29

lib/tasks/publishing_api.rake
```diff
@@ -0,0 +1,11 @@
+namespace :publishing_api do
+  desc "Publish all Finders to the Publishing API"
+  task :publish_finders => :environment do
+    require "publishing_api_finder_publisher"
+    require "publishing_api_finder_loader"
```
tuzz: :+1:

spec/lib/publishing_api_finder_loader_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        timestamp: "yesterday"
+      },
+      {
+
+        file: { "name" => "format-2" },
+        timestamp: "today"
+      }
+    ])
+  end
+end
```
tuzz: Nice – it's good to see file access tested – something that's usually quite tricky to do.

spec/lib/publishing_api_finder_publisher_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+        ]
+      }
+
+      it "uses GdsApi::PublishingApiV2" do
+        stub_publishing_api_publish(finders[0][:file]["content_id"], {})
+        stub_publishing_api_publish(finders[0][:file]["signup_content_id"], {})
+        stub_publishing_api_publish(finders[1][:file]["content_id"], {})
+
+        expect(publishing_api).to receive(:put_content)
+          .with(finders[0][:file]["content_id"], be_valid_against_schema("finder"))
```
tuzz: :tada: schema tests :tada:

tuzz:
LGTM.

As mentioned above, I'm not terribly keen on `Struct`, but that seems to be the pattern so no strong objects here.

PR #624, opened 2016-01-07
Add new Safety study report type

boffbowsh:
Not all safety studies will have a date of occurrence, so we also need
to allow it to be blank when validating.

https://trello.com/c/4euyRTaI/229-add-additional-report-type-to-aaib-reports-finder-large

Requires https://github.com/alphagov/govuk-content-schemas/pull/189 to be merged.

### DIFF
--- a/app/models/validators/aaib_report_validator.rb
+++ b/app/models/validators/aaib_report_validator.rb
@@ -9,5 +9,7 @@ class AaibReportValidator < SimpleDelegator
   validates :summary, presence: true
   validates :body, presence: true, safe_html: true
 
-  validates :date_of_occurrence, presence: true, date: true
+  validates :date_of_occurrence, presence: true, date: true, unless: ->(report) {
+    report.report_type == "safety-study" && report.date_of_occurrence.blank?
+  }
 end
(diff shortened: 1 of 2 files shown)


app/models/validators/aaib_report_validator.rb
```diff
@@ -9,5 +9,5 @@ class AaibReportValidator < SimpleDelegator
   validates :summary, presence: true
   validates :body, presence: true, safe_html: true
 
-  validates :date_of_occurrence, presence: true, date: true
+  validates :date_of_occurrence, date: true, allow_blank: true
```
tommyp: Going by the A/C: `AAIB can leave the 'date of occurrence' blank for this new report type only`. Won't this change allow blank dates across the board/

fofr: I agree, this could lead to a user error when adding a standard report which must have a date of occurrence.

boffbowsh: Apols. Yep. Will fix.

jamiecobbett: When I sketched this out, and played with it in dev, I wondered if the validation message should be tweaked as well so that it says "date of occurrence is required unless it is a Safety study".

boffbowsh: Fixed.

boffbowsh: @jamiecobbett possibly, will discuss

boffbowsh: Current solution approved by BA/PM

PR #626, opened 2016-01-11
Support manual ID in CliManualDeleter

alexmuller:
It's possible (due to a bug) to create multiple manuals with the same slug. The CliManualDeleter correctly gives an ambiguous slug error and refuses to delete anything.

This commit allows `bin/delete_draft_manual` to be called with an optional identifier in addition to the slug. If the identifier is present it will be used instead of the slug.

---

This is my first time looking at specialist-publisher and I'm not set up in any way to test this so be gentle.


elliotcm:
Might be a bit too cautious with all the nils but looks :+1: 

Needs a rebase.

alexmuller:
Thanks! Rebased.

PR #629, opened 2016-01-18
Major and minor updates

tommyp:
This PR allows editors to specify major/minor updates do documents and to add a change note to the change history if the change is major. The workflow UI is slightly different to what exists in Specialist Publisher at the moment but it is more similar to the UI within Whitehall.  [Ticket](https://trello.com/c/guqmRnt9/459-major-and-minor-updates).

Screenshot:
screen shot 2016-01-18 at 14 53 25

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -185,7 +185,8 @@ class Document
     indexable_document = SearchPresenter.new(self)
 
     begin
-      publish_request = publishing_api.publish(content_id, "major")
+      update_type = self.update_type || 'major'
+      publish_request = publishing_api.publish(content_id, update_type)
       rummager_request = rummager.add_document(
         format,
         base_path,
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -45,12 +45,16 @@ class Document
     raise NoMethodError
   end
 
-  def live?
-    publication_state == "live"
+  %w{draft live redrafted}.each do |state|
+    define_method("#{state}?") do
+      publication_state == state
+    end
   end
 
-  def draft?
-    publication_state == "draft"
+  def published?
+    live? || redrafted?
+  end
+
   end
 
   def users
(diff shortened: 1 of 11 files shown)


app/models/document.rb
```diff
@@ -2,11 +2,14 @@ class Document
...
 
-  attr_accessor :content_id, :base_path, :title, :summary, :body, :format_specific_fields, :public_updated_at, :state, :bulk_published, :publication_state, :minor_update, :change_note
+  attr_accessor :content_id, :base_path, :title, :summary, :body, :format_specific_fields, :public_updated_at, :state, :bulk_published, :publication_state, :change_note
+  attr_writer :change_history, :update_type
 
   validates :title, presence: true
   validates :summary, presence: true
   validates :body, presence: true, safe_html: true
+  validates :update_type, presence: true, if: :live?
+  validates :change_note, presence: true, if: Proc.new { |d| d.update_type == 'major' && d.published? }
```
danielroseman: Would be better to have this check as an explicit method, eg `requires_change_note`

danielroseman:
One comment, otherwise :+1:

app/models/document.rb
tommyp: Moved this logic to another method and fixedup the commit.

tommyp:
@h-lame I agree with your comment, but we have payload mutating [happening in other Apps](https://github.com/alphagov/finder-frontend/blob/master/app/lib/finder_api.rb#L45-L61). If it's a blocker I'll change it to what you've suggested, but the logic is pretty opaque either way and IMO my code comments explain what's going on. No bother if you think otherwise and have a better way to do the same thing without becoming more confusing.

h-lame:
@tommyp If there's precedent that probably takes priority - and your comments are clear.  Leave it as-is.

PR #630, opened 2016-01-21
Schema Contract testing

tommyp:
This PR adds contract testing for the only format that Phase 2 of Specialist Publisher deals with, CMA Cases. In adding the schema contract testing I discovered that there were a few issues with the data I was sending:

- the format of the payload should be `specialist_document`
- blank metadata should be omitted from the Publishing API
- `public_updated_at` should be sent as RFC3339

This PR corrects those, updates the tests to reflect this and then tests CMA Cases against the content schema. [Ticket](https://trello.com/c/16tEBnby/444-schema-contract-testing).

### DIFF
--- a/app/presenters/document_presenter.rb
+++ b/app/presenters/document_presenter.rb
@@ -43,7 +43,7 @@ private
       {
         f => document.send(f)
       }
-    end.reduce({}, :merge).merge(document_type: document.format)
+    end.reduce({}, :merge).merge(document_type: document.format).reject { |k, v| v.blank? }
   end
 
   def public_updated_at
--- a/app/presenters/document_presenter.rb
+++ b/app/presenters/document_presenter.rb
@@ -12,7 +12,7 @@ class DocumentPresenter
       base_path: document.base_path,
       title: document.title,
       description: document.summary,
-      format: document.format,
+      format: "specialist_document",
       publishing_app: "specialist-publisher",
       rendering_app: "specialist-frontend",
       locale: "en",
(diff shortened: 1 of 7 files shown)


app/presenters/document_presenter.rb
```diff
@@ -43,11 +43,11 @@ def metadata
       {
         f => document.send(f)
       }
-    end.reduce({}, :merge).merge(document_type: document.format)
+    end.reduce({}, :merge).merge(document_type: document.format).reject { |k, v| v.blank? }
```
jamiecobbett: Too late now, but our styleguide says we shouldn't chain onto `end` - [`Never chain do...end`](https://github.com/alphagov/styleguides/blob/master/ruby.md#syntax)

app/presenters/document_presenter.rb
```diff
@@ -12,7 +12,7 @@ def to_json
       base_path: document.base_path,
       title: document.title,
       description: document.summary,
-      format: document.format,
```
jamiecobbett: If the format is "specialist_document", how will you fetch just a list of CMA Cases to show in Specialist Publisher?

tommyp: I've had a chat with some of the Publishing Platform team about this. I have 2 options:

- split the Specialist Document schema into numerable schemas
- filer on the `document_type` attribute that is sent as part of the metadata

As a result of the chat, I'm going to go with the latter, but won't be able to do it on the Publishing API side fully until we implement Elastic Search. Until that happens, I'm going to go with a temp fix and throw away all the Specialist Document formats I don't need.

PR #633, opened 2016-01-25
AAIB reports phase 2

tommyp:
This PR adds a model, schema file, form partial and test for AAIB Reports along with adding it to the hash of formats in the `ApplicationController`. [Ticket](https://trello.com/c/lZ4uMAlM/452-aaib-reports).

### DIFF
--- a/app/models/aaib_report.rb
+++ b/app/models/aaib_report.rb
@@ -19,12 +19,8 @@ class AaibReport < Document
     super(params, FORMAT_SPECIFIC_FIELDS)
   end
 
-  def format
-    "aaib_report"
-  end
-
   def self.format
-    new.format
+    "aaib_report"
   end
 
   def public_path
--- a/app/models/aaib_report.rb
+++ b/app/models/aaib_report.rb
@@ -1,0 +1,33 @@
+class AaibReport < Document
+
+  validates :date_of_occurrence, presence: true, date: true, unless: ->(report) {
+    report.report_type == "safety-study" && report.date_of_occurrence.blank?
+  }
+
+  FORMAT_SPECIFIC_FIELDS = [
+    :date_of_occurrence,
+    :aircraft_category,
+    :report_type,
+    :location,
+    :aircraft_type,
+    :registration,
+  ]
+
+  attr_accessor *FORMAT_SPECIFIC_FIELDS
+
+  def initialize(params = {})
+    super(params, FORMAT_SPECIFIC_FIELDS)
+  end
+
+  def format
+    "aaib_report"
+  end
+
+  def self.format
+    new.format
+  end
+
+  def public_path
+    "/aaib-reports"
+  end
+end
--- a/spec/models/aaib_report_spec.rb
+++ b/spec/models/aaib_report_spec.rb
@@ -1,0 +1,182 @@
+require 'spec_helper'
+
+describe AaibReport do
+
+  def aaib_report_content_item(n)
+    {
+      "content_id" => SecureRandom.uuid,
+      "base_path" => "/aaib-reports/example-aaib-report-#{n}",
+      "title" => "Example AAIB Report #{n}",
+      "description" => "This is the summary of example AAIB Report #{n}",
+      "format" => "specialist_document",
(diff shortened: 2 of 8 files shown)


jamiecobbett:
Nice to see only 126 lines of implementation :+1:

spec/models/aaib_report_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+      @aaib_reports << aaib_report_content_item(n)
+    end
+
+    publishing_api_has_fields_for_format('specialist_document', @aaib_reports, @fields)
+
+    @aaib_reports.each do |aaib_report|
+      publishing_api_has_item(aaib_report)
+    end
+
+    Timecop.freeze(Time.parse("2015-12-18 10:12:26 UTC"))
```
h-lame: You should have a corresponding `Timecop.return` in an `after` block in here (or in a global after block for every spec).

app/models/aaib_report.rb
```diff
@@ -0,0 +1,33 @@
...
+    super(params, FORMAT_SPECIFIC_FIELDS)
+  end
+
+  def format
+    "aaib_report"
+  end
+
+  def self.format
+    new.format
+  end
```
h-lame: Could this method implementation be pushed down to the `Document` class?  Are there expected to be any `Document` subclasses where the `<subclass>.format` is different to the `<subclass_instance>.format` ? If not I'd suggest flipping the implementation while you're at it (for boring micro performance optimization reasons - you'll have one copy of the format string per subclass, not one copy per instance):

```ruby
class Document
  def format
    self.class.format
  end

  def self.format
    raise NotImplementedError
  end
end

class FooBar < Document
  def self.format
    'foo_bar'
  end
end
```

app/models/aaib_report.rb
```diff
@@ -0,0 +1,33 @@
...
+    :report_type,
+    :location,
+    :aircraft_type,
+    :registration,
+  ]
+
+  attr_accessor *FORMAT_SPECIFIC_FIELDS
+
+  def initialize(params = {})
+    super(params, FORMAT_SPECIFIC_FIELDS)
```
h-lame: I wonder if there's a way to avoid having to send `FORMAT_SPECIFIC_FIELDS` to the superclass initializer?  Perhaps if `FORMAT_SPECIFIC_FIELDS` was a class method?  

```ruby
class Document
  def initialize(params = {})
    self.class.all_fields).each do |field|
      public_send(:"#{field.to_s}=", params.fetch(field, nil))
    end
  end

  def self.all_fields
    [:id, :title, ... ] # e.g. what COMMON_FIELDS is right now
  end
end

class FooBar < Document
  def self.all_fields
    super + [:my, :extra, :fields]
  end
end
```

I might be missing something about how the current COMMON_FIELDS and FORMAT_SPECIFIC_FIELDS are used, but I reckon there'd be a way of reducing the burden of boilerplate on the subclasses.

tommyp: I still need to set them as `attr_accessor`s. Can I do that with your solution or is that something you missed?

h-lame: I reckon that `attr_accessor *self.all_fields` would work (assuming it comes after you define `all_fields` in the subclass.  Maybe you'll want:

```ruby
class Document
  def self.common_fields
    [:id, :name, :etc...]
  end
  attr_accessor *self.common_fields

  def self.custom_fields
    []
  end

  def self.all_fields
    common_fields + custom_fields
  end
end

class FooBar < Document
  def self.custom_fields
    [:my, :extra, :fields]
  end
  attr_accessor *self.custom_fields
end
```

Of course, it's annoying to have to remember to override custom_fields *and* provide the attr_accessor so I'd maybe start thinking about some kind of `has_custom_fields` method that meta-programs it all for you.  Getting a bit magical though.  It's a tradeoff between the things you can forget when implementing a new subclass and how complicated the meta-programming is.

h-lame:
My suggested refactoring for the initializer was breaking a load of tests so we'll look at it later as it needs more thought before it can be implemented (or discarded) properly.

PR #635, opened 2016-01-27
Access limiting

tommyp:
This PR introduces Access limiting by Organisation to Specialist Publisher Phase 2. It does this by comparing the `organisation_content_id`s listed in the Finder schema to the `organisation_content_id` received from Signon as part of the User hash, using this to get a list of the formats that the User can access, and on every action checking that the format they're attempting to access is within that list. 

This PR also updates `database_cleaner` to a version which works with Mongoid and fixes a typo.

[Ticket](https://trello.com/c/ge13v8RF/474-org-permissions).

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -6,16 +6,16 @@ class DocumentsController <  ApplicationController
   include ActionView::Helpers::TextHelper
 
   before_action :fetch_document, only: [:edit, :show, :publish, :update]
+  before_action :permitted?, if: :document_type
 
   def index
-    unless params[:document_type]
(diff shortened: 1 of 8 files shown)


danielroseman:
:+1: seems highly sensible.

app/controllers/documents_controller.rb
```diff
@@ -124,4 +124,13 @@ def rummager
     @rummager ||= SpecialistPublisher.services(:rummager)
   end
 
+  def has_permission?
```
benlovell: I think `permitted?` may be more suitable given this is a predicate.

tommyp: Changed the name of the method to `permitted?`.

PR #636, opened 2016-01-28
Multiple finder nav

tommyp:
This commit pulls over the Multiple Finder from Specialist Publisher. The Frontend is very similar to Master. If you're a `gds_editor`, it will render the list as a dropdown. If not, it will render each format the User have access to as an element in the Nav. [Ticket](https://trello.com/c/oAZWkOFl/473-multiple-finder-nav).

## Screenshot:
screen shot 2016-01-28 at 15 00 42

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,4 +12,8 @@ class User
   field :organisation_content_id, type: String
   field :disabled, type: Boolean, default: false
 
+  def gds_editor?
+    permissions.include?('gds_editor')
+  end
+
 end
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -12,4 +12,8 @@ class User
   field :organisation_content_id, type: String
   field :disabled, type: Boolean, default: false
 
+  def is_gds_editor?
```
benlovell: Perhaps `gds_editor?` :blush: Since predicates answer true/false then the `is_` part is superfluous.

tommyp: Good call. Fixed.

PR #640, opened 2016-02-03
Migrate formats for RAIB reports

Rosa-Fox:


tommyp:
I have a look over this and I'm giving it a :+1: but I'd rather we spread knowledge about the new Specialist Publisher, so if someone can take a look over this and hit merge it would be much appreciated.

PR #641, opened 2016-02-05
Refactor Document.document_type

Rosa-Fox:
This PR separates the `Document.document_type` method into `publishing_api_document_type` and `search_document_type`. This is needed for the formats in which these 2 strings are different. The `search_document_type` is the same as the `filter["document_type"]` in the Schema, so this PR also pulls the string from there.

Some comments have been added to the FormStruct methods within the application_controller to describe to what they are and what they are used for.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,11 +40,11 @@ private
     #
     # {
     #   "gds-reports" => FormatStruct.new(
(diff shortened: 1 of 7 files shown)


tommyp:
I paired with @Rosa-Fox on this so if someone else could review and hit merge that would be great.

PR #643, opened 2016-02-10
Speed up sub-format check

issyl0:
- This avoids making lots of requests to get expanded `details`, and
  just checks `base_path` the finder's base path, as that's all we need.

(Paired completely with @tommyp on this.)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -157,21 +157,17 @@ class Document
       fields: [
         :base_path,
         :content_id,
-        :title,
-        :public_updated_at,
-        :details,
-        :description,
       ]
     ).to_ostruct
 
-    # Fetch individual payloads for each `specialist_document`
-    payloads = response.map { |payload| publishing_api.get_content(payload.content_id).to_ostruct }
-
     # Select the ones which match the current document type
-    payloads_of_format = payloads.select { |payload| payload.details.metadata.document_type == self.publishing_api_document_type }
+    payloads_of_format = response.select { |payload| payload.base_path.starts_with?(self.finder_schema.base_path) }
+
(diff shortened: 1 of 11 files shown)


app/models/document.rb
```diff
@@ -157,21 +157,17 @@ def self.all
...
     # Select the ones which match the current document type
-    payloads_of_format = payloads.select { |payload| payload.details.metadata.document_type == self.publishing_api_document_type }
+    payloads_of_format = response.select { |payload| payload.base_path.starts_with?(self.finder_schema.base_path) }
+
+    # Fetch individual payloads for each `specialist_document`
+    payloads = payloads_of_format.map { |payload| publishing_api.get_content(payload.content_id).to_ostruct }
 
     # Deserialize the payloads into real Objects and return them
-    payloads_of_format.map { |payload| self.from_publishing_api(payload) }
+    payloads.map { |payload| self.from_publishing_api(payload) }
```
benlovell: You could golf this down to `payloads.map(&method(:from_publishing_api))`.

Using `&method` is different to your usual `Symbol#to_proc` in that the receiver is `self` and the yielded argument is passed along, rather than the receiver being the argument yielded to the block.

In explaining this I feel I've argued why you shouldn't do this, perhaps :laughing:

tommyp: We could've, but it's been discussed before that people are largely unfamiliar with `&method`. I'm very wary of introducing concepts within this App that people aren't familiar with before they've had a chance to get to grips with it.

benlovell: K.I.S.S :+1:

benlovell:
LGTM. :ship:

PR #644, opened 2016-02-11
Medical Safety Alerts require the `bulk_published` flag, so persist it

issyl0:


jamiecobbett:
I was wondering what this was actually used for, and then I looked in [specialist-frontend](https://github.com/alphagov/specialist-frontend/blob/master/app/presenters/document_presenter.rb#L74) :+1:

PR #645, opened 2016-02-11
Don't send blank values from metadata to Rummager

issyl0:
- We don't send blank (nil) values to the Publishing API so we
  shouldn't send them to Rummager either, to be consistent.


tommyp:
I paired with @issyl0 on this so if someone else could hit merge that would be great. This change may break some tests in `phase-two` but it should only require removing `key: nil` from expected Rummager responses.

PR #646, opened 2016-02-11
Add the Medical Safety Alert specialist document format

issyl0:
- Add a schema for it, metadata form fields, a model and its
  validations, and model specs.

https://trello.com/c/UFpwZWOS/537-msa

### DIFF
--- a/app/views/metadata_fields/_medical_safety_alerts.html.erb
+++ b/app/views/metadata_fields/_medical_safety_alerts.html.erb
@@ -1,0 +1,6 @@
+<%= f.label :alert_type %>
+<%= f.select :alert_type, f.object.facet_options(:alert_type), { include_blank: true }, { class: 'form-control' } %>
+<%= f.label :medical_specialism %>
+<%= f.select :medical_specialism, f.object.facet_options(:medical_specialism), {}, { class: 'select2', multiple: true, data: { placeholder: 'Select a Medical specialism' } } %>
+<%= f.label :issued_date %>
+<%= f.text_field :issued_date, placeholder: '2012-04-23', class: 'form-control' %>
--- a/spec/models/medical_safety_alert_spec.rb
+++ b/spec/models/medical_safety_alert_spec.rb
@@ -1,0 +1,175 @@
+require 'spec_helper'
+
+describe MedicalSafetyAlert do
+
+  def medical_safety_alert_content_item(n)
+    {
+      "content_id" => SecureRandom.uuid,
+      "base_path" => "/drug-device-alerts/example-medical-safety-alert-#{n}",
+      "title" => "Example Medical Safety Alert #{n}",
+      "description" => "This is the summary of example Medical Safety Alert #{n}",
+      "format" => "specialist_document",
+      "publishing_app" => "specialist-publisher",
+      "rendering_app" => "specialist-frontend",
+      "locale" => "en",
+      "phase" => "live",
+      "public_updated_at" => "2015-11-16T11:53:30",
+      "publication_state" => "draft",
+      "details" => {
+        "body" => "## Header" + ("\r\n\r\nThis is the long body of an example Medical Safety Alert" * 10),
+        "metadata" => {
+          "alert_type" => "company-led-drugs",
+          "issued_date" => "2016-02-01",
+          "document_type" => "medical_safety_alert"
+        },
+        "change_history" => [],
+      },
+      "routes" => [
+        {
+          "path" => "/drug-device-alerts/example-medical-safety-alert-#{n}",
+          "type" => "exact",
+        }
+      ],
+      "redirects" => [],
+      "update_type" => "major",
+    }
+  end
+
+  let(:non_medical_safety_alert_content_item) {
+    {
+      "content_id" => SecureRandom.uuid,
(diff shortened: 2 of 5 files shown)


spec/models/medical_safety_alert_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+      "description" => "This is the summary of example Medical Safety Alert #{n}",
+      "format" => "specialist_document",
+      "publishing_app" => "specialist-publisher",
+      "rendering_app" => "specialist-frontend",
+      "locale" => "en",
+      "phase" => "live",
+      "public_updated_at" => "2015-11-16T11:53:30",
+      "publication_state" => "draft",
+      "details" => {
+        "body" => "## Header" + ("\r\n\r\nThis is the long body of an example Medical Safety Alert" * 10),
```
jamiecobbett: Should this really be Govspeak?

spec/models/medical_safety_alert_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+      "base_path" => "/government/organisations/medicines-and-healthcare-products-regulatory-agency",
+      "content_id" => "240f72bd-9a4d-4f39-94d9-77235cadde8e",
+      "title" => "Medicines and Healthcare products Regulatory Agency",
+      "format" => "placeholder_organisation",
+      "need_ids" => [],
+      "locale" => "en",
+      "updated_at" => "2015-12-01T15:51:43.521Z",
+      "public_updated_at" => "2015-06-29T14:33:29.000+00:00",
+      "phase" => "live",
+      "analytics_identifier" => "EA63",
```
jamiecobbett: Is this field needed? The original purpose was to use it for organisations and world locations.

app/views/metadata_fields/_medical_safety_alerts.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= f.label :alert_type %>
+<%= f.select :alert_type, f.object.facet_options(:alert_type), {include_blank: true}, { class: 'form-control' } %>
```
jamiecobbett: Our [styleguide says](https://github.com/alphagov/styleguides/blob/master/ruby.md#general) we should put spaces inside `{` and `}`.

spec/models/medical_safety_alert_spec.rb
tommyp: This is just copying what is sent from master. As part of migration work, master will need to send Govspeak and rendered Govspeak, in which case these tests will need to be updated.

spec/models/medical_safety_alert_spec.rb
tommyp: This was copied from the existing response. Only the `base_path`, `content_id` and `title` are required.

jamiecobbett: Is it realistic then? Do Specialist Documents ever use that field?

tommyp: It's realistic in that it is the response, but it's not used.

spec/models/medical_safety_alert_spec.rb
tommyp: Also of note, this does *need* to be Govspeak right now, as this is what the User is editing every time they edit the document.

app/views/metadata_fields/_medical_safety_alerts.html.erb
issyl0: Fixed.

boffbowsh:
See the changes in https://github.com/alphagov/specialist-publisher/pull/655 for the style of these specs

issyl0:
Thanks @tommyp, @boffbowsh. I'll get round to updating the specs tomorrow, promise. :sunny:

issyl0:
@tommyp @boffbowsh I've changed the style of the specs as per #656.

spec/models/medical_safety_alert_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+    publishing_api_has_fields_for_format('specialist_document', medical_safety_alerts, fields)
+
+    medical_safety_alerts.each do |medical_safety_alert|
+      publishing_api_has_item(medical_safety_alert)
+    end
+
+    Timecop.freeze(Time.parse("2015-12-18 10:12:26 UTC"))
+  end
+
+  describe "#all" do
```
boffbowsh: Nitpicky, but these should be `describe ".all"` and `describe ".find"` as they're class methods.

issyl0: Ugh. I'll make a note to update where I copied and pasted them from then as well as here and presumably https://github.com/alphagov/specialist-publisher/pull/656/files#diff-87a2b2747a4f709e7ef875915e7fcee5R106 too.

issyl0: Thanks. Done.

Rosa-Fox:
@issyl0 Teeny tiny thing which is that `".save!"` on 135 of `medial_safety_alert_spec.rb` should be `#save!`

issyl0:
@Rosa-Fox Oh! Thanks. Fixed.

PR #653, opened 2016-02-15
Render manuals

tommyp:
This PR adds Manuals as a format. It includes barebones `Manual` and `Section` models, changes to navigation and routes, access control and a spec for ensuring that a `Manual` can be rendered. Viewing of sections will be done in a later PR. As `links` are important to how Manuals are constructed, this work involves merging the `/content` and `/links` responses into one payload and using that for deserialization. Respecting the order of Sections will involve some work within master which is still to come. [Ticket](https://trello.com/c/5RJAIg00/532-render-manuals).


app/models/manual.rb
```diff
@@ -0,0 +1,163 @@
...
+private
+
+  def self.content_ids
+    response = self.publishing_api.get_content_items(
+      content_format: "manual",
+      fields: [
+        :content_id,
+      ]
+    ).to_ostruct
+  end
```
h-lame: The coercion to open struct here seems redundant.  What if this method actually returned an array of content ids, instead of an array of `OpenStruct` instances whose only purpose was to have `.content_id` called on them? e.g. something like `response = [get them].map { |x| x[:content_id] }`.

Also, the `private` keyword doesn't actually affect methods on `self` so this method is still public.  Use `private_class_method :content_ids` if you really want this to be private on the class.

app/models/manual.rb
```diff
@@ -0,0 +1,163 @@
...
+        content = publishing_api.get_links(content_id).to_hash
+        payload = content.deep_merge(links.to_hash)
+      else
+        return nil
+      end
+    else
+      content = publishing_api.get_content(content_id).to_hash
+      links = publishing_api.get_links(content_id).to_hash
+      payload = content.deep_merge(links)
+    end
```
h-lame: Could we push this pattern (get content, get links, merge responses) down into the api adaptor (or even into the api itself)?

app/models/manual.rb
```diff
@@ -0,0 +1,163 @@
...
+      content.deep_merge!(payload.links)
+    }
+
+    # Deserialize the payloads into real Objects and return them
+    payloads.map { |payload| self.from_publishing_api(payload) }
+  end
+
+  def self.find(content_id:, organisation_content_id: nil)
+    if organisation_content_id
+      links = publishing_api.get_links(content_id).o_struct
```
h-lame: Is this a typo?  Should it be `.to_ostruct` at the end?

app/models/manual.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+
+  OrganisationStruct = Struct.new(:content_id, :base_path, :title)
+
+  def organisations
+    @organisations ||= @organisation_content_ids.map { |content_id|
+      payload = publishing_api.get_content(content_id).to_hash
+      OrganisationStruct.new(payload["content_id"], payload["base_path"], payload["title"])
+    }
+  end
```
h-lame: Why not just call these `Organisation` and give them the same treatment as `Section` above by making them first class models?  (Even if they are tiny).

app/models/manual.rb
```diff
@@ -0,0 +1,163 @@
...
+      }
+    )
+
+    manual.base_path = payload["base_path"]
+    manual.update_type = payload["update_type"]
+
+    if payload["links"]
+      manual.organisations = payload["links"].fetch("organisations", [])
+      manual.sections = payload["links"].fetch("sections", [])
+    end
```
h-lame: It feels a little odd to me that we don't / can't supply `base_path`, `update_type`, `organisations`, or `sections` as part of the initializer.  I don't think it would make the method significantly different or shorter, but it feels a little strange.

app/models/manual.rb
tommyp: Yeah. The `to_ostruct` isn't required. I forgot to delete it when I did some refactoring. I'll remove it. 

And TIL that `private` doesn't work like that. It doesn't need to be accessed anywhere else, so I can use `private_class_method`.

app/models/manual.rb
tommyp: It should. Fixed.

app/models/manual.rb
tommyp: We can do, but I'm not sure if it's going to be that common. The reason I'm having to do this here is in order to filter by `organisation`. When we have a proper search interface, I won't need to do this in within the App.

app/models/manual.rb
tommyp: I can move it to a full model, but this is the extent of the logic surrounding Organisations in `master`. If you'd prefer this approach I can do it. I just went with the simplest option.

app/models/manual.rb
tommyp: I wanted to be able to construct a `Manual` from params the same way we construct a `Document`. I haven't started on persisting a `Manual` yet, but I imagine a lot of the methods will be similar such as:

```ruby
def base_path
    @base_path ||= "guidance/#{title.parameterize}"
  end
```

There are probably other ways to add `base_path` to the initializer and still follow this behaviour, but if I'm able to follow similar approaches for both and refactor or DRY them up later I thought that would be better. `update_type` won't be supplied from the `params` with Manuals as it'll have to have logic deciding this based on the `update_type` of the sections.

tommyp:
I've just realised I haven't added a test for viewing Manuals when you aren't a GDS editor. Feel free to continue to comment whilst I do that though.

app/models/manual.rb
h-lame: Fair enough - I thought it might be more common.  Leave as-is then, we can always revisit later.

app/models/manual.rb
h-lame: That's useful context.  Maybe you could work this into the relevant commit message(s)?

app/models/manual.rb
```diff
@@ -0,0 +1,164 @@
...
+  validates :summary, presence: true
+  validates :body, safe_html: true
+
+  def initialize(params)
+    @content_id = params.fetch(:content_id, SecureRandom.uuid)
+    @title = params.fetch(:title, nil)
+    @summary = params.fetch(:summary, nil)
+    @body = params.fetch(:body, nil)
+    @publication_state = params.fetch(:publication_state, nil)
+    @public_updated_at = params.fetch(:public_updated_at, nil)
```
h-lame: Should this go through the setter defined below?  Wondering if `:public_updated_at` is already turned into a `Time` object for us by the API or not?  Maybe all of these should go through the setters so that we get the benefit of any data-massaging we define in the future?

app/models/manual.rb
tommyp: `update_type` is set here as it needs to be persisted if the User makes no changes to a section and saves it again. If any sections are changed, the changed `Manual` will be saved again with a recalculated `update_type`. What should I be noting in the commit message as all of this behaviour is being carried over from master?

app/models/section.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def initialize(params)
+    @content_id = params.fetch(:content_id, SecureRandom.uuid)
+    @title = params.fetch(:title, nil)
+    @summary = params.fetch(:summary, nil)
+    @body = params.fetch(:body, nil)
+    @publication_state = params.fetch(:publication_state, nil)
+    @updated_at = params.fetch(:updated_at, nil)
+    @public_updated_at = params.fetch(:public_updated_at, nil)
+    @manual = params.fetch(:manual_content_id)
```
h-lame: Same question I asked about the `Manual` initializer.  Seems more pressing here as this method sets `@manual` with the value of :manual_content_id in the payload, but it probably means to set `@manual_content_id`.

app/models/section.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def public_updated_at
+    @public_updated_at ||= Time.zone.now
+  end
+
+  def public_updated_at=(timestamp)
+    @public_updated_at = Time.parse(timestamp.to_s) unless timestamp.nil?
+  end
+
+  def manual=(manual_content_id)
```
h-lame: Seems a little bit dangerous that `Section#manual` returns a `Manual` instance, but `Section#manual=` expects a content id.  Perhaps breaking this into `Section#manual`/`Section#manual=` and `Section#manual_content_id`/`Section#manual_content_id=` pairs that deal with `Manual` instances and content ids would be clearer (and for better or worse, more like ActiveRecord).

app/models/manual.rb
tommyp: It's not turned into it by the API, but it should be sent as one. I think in this instance using the setter or the instance var will both work.

app/models/section.rb
tommyp: Yes. Good catch. I'll change it

app/models/manual.rb
h-lame: In that case we should set `public_updated_at` in `from_publishing_api` outside the call to `new` as we're passing in the untouched value from the API response `public_updated_at` which is likely a `String`, not a `Time`.

app/models/section.rb
tommyp: You're right that does make more sense. What do the setter methods for the model become then?

app/models/manual.rb
h-lame: I'd say all of it.  Even though it's the same repo there's no git history to trace back from this new version to the old version for anyone from the future to trace back and unpick even though it's the intended behaviour.

tommyp:
I've updated the `viewing_a_manual` spec to use both a `gds_editor` and an Organisation editor factory.

app/models/manual.rb
```diff
@@ -0,0 +1,173 @@
+class Manual
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  attr_accessor :content_id, :base_path, :title, :summary, :body, :updated_at, :public_updated_at, :publication_state, :update_type
```
boffbowsh: As you have explicit getters/setters for `updated_at` and `public_updated_at` below, they should probably be removed from here to avoid confusion

app/models/section.rb
```diff
@@ -0,0 +1,116 @@
+class Section
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  attr_accessor :content_id, :base_path, :title, :summary, :body, :updated_at, :public_updated_at, :organisations, :update_type
```
boffbowsh: Same comment here about setters/getters

spec/fixtures/factories.rb
```diff
@@ -13,6 +13,11 @@
     permissions %w(signin editor)
   end
 
+  factory :gds_editor, parent: :user do
```
boffbowsh: This only creates a user affiliated with GDS doesn't it? Not an editor?

tommyp: There's functionality in master for users to be able to log in, but not edit or publish documents. Since I was going MVP as fast as possible, I didn't pull across this functionality but it's in the factories. I've noted it on the board.

h-lame:
I think I'm happy, passing the baton to @boffbowsh.

app/models/manual.rb
tommyp: Removed these here as well as in `Section`.

PR #654, opened 2016-02-15
Migrate formats for Employment Tribunal Decisions

Rosa-Fox:
https://trello.com/c/PT4A0cdJ/540-et-decisions

### DIFF
--- a/spec/models/employment_tribunal_decision_spec.rb
+++ b/spec/models/employment_tribunal_decision_spec.rb
@@ -1,0 +1,178 @@
+require 'spec_helper'
+
+describe EmploymentTribunalDecision do
+
+  def employment_tribunal_decision_content_item(n)
+    {
+      "content_id" => SecureRandom.uuid,
+      "base_path" => "/employment-tribunal-decisions/example-employment-tribunal-decision-#{n}",
+      "title" => "Example Employment Tribunal Decision #{n}",
+      "description" => "This is the summary of example Employment Tribunal Decision #{n}",
+      "format" => "specialist_document",
+      "publishing_app" => "specialist-publisher",
+      "rendering_app" => "specialist-frontend",
+      "locale" => "en",
+      "phase" => "live",
+      "public_updated_at" => "2015-11-16T11:53:30",
+      "publication_state" => "draft",
+      "details" => {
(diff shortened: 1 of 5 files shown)


spec/models/employment_tribunal_decision_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      "public_timestamp" => "2015-11-16T11:53:30+00:00",
+      "tribunal_decision_categories" => ["age-discrimination"],
+      "tribunal_decision_country" => "england-and-wales",
+      "tribunal_decision_decision_date" => "2015-07-30",
+      "organisations" => ["employment-tribunal"],
+    }
+  }
+
+  before do
+    @fields = [
```
boffbowsh: This and `@employment_tribunal_decisions` could be `let`s too I think.

spec/models/employment_tribunal_decision_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+    publishing_api_has_fields_for_format('specialist_document', @employment_tribunal_decisions, @fields)
+
+    @employment_tribunal_decisions.each do |decision|
+      publishing_api_has_item(decision)
+    end
+
+    Timecop.freeze(Time.parse("2015-12-18 10:12:26 UTC"))
+  end
+
+  context "#all" do
```
boffbowsh: It's better to use `describe` for blocks describing the behaviour of a method, then `context` to define the context for a particular set of behaviours for that method/object.

http://betterspecs.org/#describe

spec/models/employment_tribunal_decision_spec.rb
tommyp: What's the correct syntax for turning an iteration into a `let` block?

boffbowsh:
```ruby
let(:fields) { %i[base_path content_id] }
let(:employment_tribunal_decisions) { 10.times.map { |n| employment_tribunal_decision_content_item(n) } }
```

tommyp: Ah yeah, I had a bit of a brain fart there.

spec/models/employment_tribunal_decision_spec.rb
Rosa-Fox: Thanks for the feedback, I have altered and pushed.

PR #655, opened 2016-02-15
Update model specs

tommyp:
As per the comments [here](https://github.com/alphagov/specialist-publisher/pull/654#discussion_r52914670) this commit updates the model spec files to use `let`s and `describe`s.


tommyp:
Forgot AAIB.

PR #656, opened 2016-02-15
Update model specs

tommyp:
As per the comments [here](https://github.com/alphagov/specialist-publisher/pull/654#discussion_r52914670) this commit updates the model spec files to use `let`s and `describe`s.

### DIFF
--- a/spec/models/aaib_report_spec.rb
+++ b/spec/models/aaib_report_spec.rb
@@ -89,21 +89,14 @@ describe AaibReport do
     }
   }
 
+  let(:fields) { %i[base_path content_id] }
+
+  let(:aaib_reports) { 10.times.map { |n| aaib_report_content_item(n) } }
+
   before do
-    @fields = [
-      :base_path,
-      :content_id,
-    ]
+    publishing_api_has_fields_for_format('specialist_document', aaib_reports, fields)
 
-    @aaib_reports = []
-
-    10.times do |n|
-      @aaib_reports << aaib_report_content_item(n)
-    end
-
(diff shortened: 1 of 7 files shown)


spec/models/aaib_report_spec.rb
```diff
@@ -89,60 +89,53 @@ def aaib_report_content_item(n)
...
-      :content_id,
-    ]
-
-    @aaib_reports = []
+  let(:fields) { %i[base_path content_id] }
 
-    10.times do |n|
-      @aaib_reports << aaib_report_content_item(n)
-    end
+  let(:aaib_reports) { 10.times.map { |n| aaib_report_content_item(n) } }
```
boffbowsh: I serendipitously discovered last night that Rubocop would prefer this to be written as:
```ruby
Array.new(10) { |n| aaib_report_content_item(n) }
```

I have no strong feelings either way.

boffbowsh:
:boom: up to you if you want to address that comment. :+1: otherwise.

spec/models/aaib_report_spec.rb
tommyp: I'm inclined to leave this as is. I don't think the more Rubocop way is as clear to people new to Ruby and/or this Application.

boffbowsh: Agreed. This is more explicit.

issyl0:
This needs a rebase.

Rosa-Fox:
Hi, I have rebased and cherry picked from a test I fixed on a separate PR so this should be ok to merge now.

PR #659, opened 2016-02-16
Add rendering of sections to manuals

brenetic:
- This follows the same approach as manuals and also adds a new test.

[Ticket](https://trello.com/c/mdlivZJi/605-render-sections)

### DIFF
--- a/app/controllers/manual_sections_controller.rb
+++ b/app/controllers/manual_sections_controller.rb
@@ -1,3 +1,9 @@
 class ManualSectionsController < ApplicationController
+  def show
+    @section = Section.find(content_id: params[:content_id])
+  end
 
+  def edit
+
+  end
 end
--- a/app/controllers/manual_sections_controller.rb
+++ b/app/controllers/manual_sections_controller.rb
@@ -1,6 +1,15 @@
 class ManualSectionsController < ApplicationController
   def show
-    @section = Section.find(content_id: params[:content_id])
+    @section = Section.find(
+      content_id: params[:content_id],
+      manual_content_id: params[:manual_content_id]
+    )
+  rescue Section::RecordNotFound => e
+    flash[:danger] = e.message
+
+    redirect_to manuals_path
+
+    Airbrake.notify(e)
   end
 
   def edit
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -70,12 +70,22 @@ class Section
     }
   end
 
-  def self.find(content_id:)
-    self.from_publishing_api(content_id: content_id)
+  def self.find(content_id:, manual_content_id: nil)
+    section = self.from_publishing_api(content_id: content_id)
+
+    if manual_content_id && section.manual_content_id != manual_content_id
+      raise RecordNotFound.new("Section does exist, but not within the supplied manual")
+    end
+
+    section
   end
 
   def self.from_publishing_api(content_id:)
-    content = self.publishing_api.get_content(content_id).to_hash
+    content_item_response = self.publishing_api.get_content(content_id)
+
+    raise RecordNotFound.new("Section not found") unless content_item_response
+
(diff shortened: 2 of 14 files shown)


app/controllers/manual_sections_controller.rb
```diff
@@ -1,3 +1,9 @@
 class ManualSectionsController < ApplicationController
+  def show
+    @section = Section.find(content_id: params[:content_id])
```
h-lame: As this action is routed via a nested url (`/manuals/:manual_content_id/sections/:content_id`) do we want to make sure that the supplied `content_id` is really a child of the manual with the supplied `manual_content_id`?  Probably something like `raise <whatever-our-404-exception-is> if @section.manual_content_id != params[:manual_content_id]`

h-lame: Maybe making that functionality part of `Section.find` would be interesting?

tommyp: I've followed a similar approach for `Manual.where` so it would make sense to do this in `Section.find`. I haven't got around to adding 404s for all these AR like things, but it's on a post it note on the board.

tommyp: @h-lame I'm going add some 404 errors to the App this morning. Master redirects to `manuals_path` and displays `flash[:danger]`. Is that enough for this or should I be doing `status: 404`, raising `ActiveRecord::RecordNotFound` or raising `ActionController::RoutingError.new('Not Found')`?

h-lame: I reckon a 404 would be better than a redirect with flash.  I don't think using `ActiveRecord::RecordNotFound` is the right thing to do though as the app isn't using ActiveRecord, so while we might get the right result (a 404) it'd be an unusual way to do it.  We could raise our own `ManualNotFound` error and add a `rescue_from` to the `ApplicationController` to make it more explicit.

steventux:
@h-lame I've attempted to finish up this PR by implementing the section/manual validation in `Section.find` as discussed, it's not an entirely comfortable fit as it kind of pollutes the AR-iness of the `Section.find` method (hence the optional named param). That said, I think this is a worthwhile trade-off for the validation behaviour you suggested.
Also added some authentication support for features, this could be picked into a later PR if need be.

h-lame:
Looks reasonable, however it seems the pattern in this app is to redirect when we get a document not found style exception (see https://github.com/alphagov/specialist-publisher/pull/665 and https://github.com/alphagov/specialist-publisher/pull/662).  Might be worth following the pattern established in those PRs (raise a ManualSection::RecordNotFound and redirect - in both "this section doesn't exist" and "this section does exist, but not within the supplied manual" cases).

steventux:
Fair enough, I didn't notice this, will amend.

steventux:
@h-lame Thanks for reviewing, updated as discussed.

app/models/section.rb
```diff
@@ -102,6 +109,10 @@ def self.from_publishing_api(content_id:)
     section
   end
 
+  def validate_manual(manual_content_id)
+    raise ManualSection::RecordNotFound if self.manual_content_id != manual_content_id
```
h-lame: I don't think there's a need to raise a different type of error here.  I consider this as being like what ActiveRecord models does when doing a find on a scoped relation; it raises `ActiveRecord::RecordNotFound` regardless of the specific reason it can't be found (e.g. `Post.find(12)` vs. `Post.where(author_id: Author.first).find(12)` - both would raise `ActiveRecord::RecordNotFound`).  

ps. I think I said `ManualSection::RecordNotFound` in my previous review, but that's just because I forgot these were just called `Section`, sorry!

steventux:
@h-lame that makes sense and simplifies, thanks, updated.

h-lame:
:+1:

PR #660, opened 2016-02-16
Employment appeal tribunal decisions phase two

Rosa-Fox:
https://trello.com/c/Vs9h7o4g/541-eat-decisions


spec/models/employment_appeal_tribunal_decision_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    it "rejects any non Employment Appeal Tribunal Decisions" do
+      all_specialist_documents = [non_employment_appeal_tribunal_decision_content_item] + employment_appeal_tribunal_decisions
+      publishing_api_has_fields_for_format('specialist_document', all_specialist_documents , fields)
+      publishing_api_has_item(non_employment_appeal_tribunal_decision_content_item)
+
+      expect(described_class.all.length).to be(employment_appeal_tribunal_decisions.length)
+    end
+  end
+
+  describe "#find" do
```
boffbowsh: Just realised this is referring to a class method. The standard Ruby notation is for class methods to use a dot, and instance methods to use a hash. So this should be `describe ".find"` and below should stay `describe "#save!"`. Sorry for not spotting this before.

Rosa-Fox: @boffbowsh thanks! I have pushed these changes.

PR #661, opened 2016-02-17
Add `TaxTribunalDecision` Specialist Publisher format

gpeng:
Part of migration to 'Phase Two'

### DIFF
--- a/spec/models/tax_tribunal_decision_spec.rb
+++ b/spec/models/tax_tribunal_decision_spec.rb
@@ -1,0 +1,177 @@
+require "rails_helper"
+
+describe TaxTribunalDecision do
+
+  def tax_tribunal_decision_content_item(n)
+    {
+      "content_id" => SecureRandom.uuid,
+      "base_path" => "/tax-and-chancery-tribunal-decisions/example-tax-tribunal-decision-#{n}",
+      "title" => "Example Tax Tribunal Decision #{n}",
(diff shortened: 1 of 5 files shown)


spec/models/tax_tribunal_decision_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+    publishing_api_has_fields_for_format('specialist_document', tax_tribunal_decisions, fields)
+
+    tax_tribunal_decisions.each do |decision|
+      publishing_api_has_item(decision)
+    end
+
+    Timecop.freeze(Time.parse("2015-12-18 10:12:26 UTC"))
+  end
+
+  describe "#all" do
```
boffbowsh: These should be `describe ".all"` and `describe ".find"`.

boffbowsh:
Nitpicky comments, otherwise :+1:

spec/models/tax_tribunal_decision_spec.rb
gpeng: I usually do `.` for class methods and `#` for instance methods. Happy to change if it's inconsistent...

gpeng: Or plainly wrong! My mistake... These are class methods. 🙊

gpeng: Fixed in 07e8880

PR #662, opened 2016-02-19
Redirect to index if document not found

tommyp:
If the User tries to visit a page for a Document which doesn't exist then this change will mean they are redirected back to the index for that format. We do this by raising a `Document::RecordNotFound` in the model and rescuing from it in the `DocumentsController`, which redirects the user to the index for that document format. This PR also bumps `gds-api-adapters` to get new `publishing_api_does_not_have_item` test helper.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -101,7 +101,14 @@ private
   end
 
   def fetch_document
-    @document = document_klass.find(params[:content_id])
+    begin
+      @document = document_klass.find(params[:content_id])
+    rescue Document::RecordNotFound => e
+      flash[:danger] = "Document not found"
+      redirect_to documents_path(document_type: document_type)
+
+      Airbrake.notify(e)
+    end
   end
 
   def filtered_params(params_of_document)
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ end
 if ENV["API_DEV"]
   gem "gds-api-adapters", :path => "../gds-api-adapters"
 else
-  gem "gds-api-adapters", "28.0.2"
+  gem "gds-api-adapters", "28.1.1"
 end
 
 group :development, :test do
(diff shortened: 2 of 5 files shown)


app/controllers/documents_controller.rb
```diff
@@ -101,7 +101,16 @@ def document_error_messages
...
-    @document = document_klass.find(params[:content_id])
+    begin
+      @document = document_klass.find(params[:content_id])
+    rescue Document::RecordNotFound => e
+      flash[:danger] = "Document not found"
+      redirect_to documents_path(document_type: document_type)
+
+      Airbrake.notify(e)
+
+      false
```
h-lame: You don't need to return false in filters any more to halt the chain; a call to `render`, `redirect_to` or `head` will do that for you.

app/controllers/documents_controller.rb
```diff
@@ -101,7 +101,16 @@ def document_error_messages
   end
 
   def fetch_document
-    @document = document_klass.find(params[:content_id])
+    begin
+      @document = document_klass.find(params[:content_id])
+    rescue Document::RecordNotFound => e
+      flash[:danger] = "Document not found"
+      redirect_to documents_path(document_type: document_type)
```
h-lame: Seems a bit strange to do a flash/redirect dance, any reason not to just issue a 404?  Or is it because this is an admin app, not a frontend one?

tommyp: I think doing a 404 here will render the standard Rails one. We want to redirect the User to a sensible place and give them some indication why. I'm not sure if this is standard in other Publishing Apps but this is the same behaviour as master, other than redirecting to the format instead of `manuals_path`.

app/controllers/documents_controller.rb
tommyp: Fixed.

app/controllers/documents_controller.rb
h-lame: Fair enough - replicating existing behaviour seems legit.  FWIW I checked contacts-admin and imminence and they seem happy to 404 in admin pages, so maybe something to add to a board or whatever to check for a wider pattern and conform later.

PR #665, opened 2016-02-19
Redirect if manual not found

tommyp:
Similar to #662 this PR will redirect the User to `manuals_path` if the Manual isn't found. We do this by raising a `Manual::RecordNotFound` in the model and rescuing from it in the `ManualsController`, which redirects the user to the `manuals_path`.


h-lame:
Looks the same as #662.  Merging.

PR #668, opened 2016-03-02
Fix broken links to finders in README

rgarner:
* And some titles, too


issyl0:
Oh, hello! :star:

rgarner:
Hai :small_orange_diamond:

jamiecobbett:
:wave:

elliotcm:
@rgarner were those URLs never live?  There's no redirects in place from the old ones.

elliotcm:
Github won't even let me do a manual merge while that status check is pending :/

rgarner:
@elliotcm don't think so - same sort of typos in them, looks like they were transcribed by hand and never clicked*

*speculation

PR #669, opened 2016-03-02
Fix broken links to finders in README

elliotcm:
Pulled from https://github.com/alphagov/specialist-publisher/pull/668


issyl0:
Thanks Elliot, and Russell!

PR #670, opened 2016-03-03
Change employment appeal tribunal decision sub-categories validation

robmckinnon:
Domain experts want employment appeal tribunal decision sub-categories not to be mandatory as some parent categories do not have any sub-categories.

Remove the validation requiring `tribunal_decision_sub_categories` be present.


h-lame:
This looks straight forward enough.  Merging.

PR #672, opened 2016-03-03
Update administrative appeals tribunal decisions sub-categories values

robmckinnon:
Update administrative appeals tribunal decisions sub-categories values.

Fix numeric sorting errors.

Add a few new sub-categories.


h-lame:
@robmckinnon - this seems clear enough - but it needs a rebase before we can merge.

PR #676, opened 2016-03-11
Redirect a manual and sections

brenetic:
- Adds a bin script wrapped around a utility class
- Passing a base_path and destination will find the manual with the slug matching the base_path and redirect it and the associated sections by publishing redirects to the publishing api.

The task for Specialist Publisher Rebuild [is here](https://github.com/alphagov/specialist-publisher-rebuild/pull/702).

eg. `bin/redirect_manual_and_sections /manual-url /redirected-url`

with @steventux

### DIFF
--- a/bin/redirect_manual_and_sections
+++ b/bin/redirect_manual_and_sections
@@ -1,0 +1,26 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require "specialist_publisher"
+require "manual_and_sections_redirecter"
+
+def usage
+  STDERR.puts %(
+USAGE:
+
+  ./#{File.basename(__FILE__)} base_path destination
+  )
+end
+
+unless ARGV.size == 2
+  usage
+  exit 1
+end
+
+base_path = ARGV.fetch(0)
+destination = ARGV.fetch(1)
+
+ManualAndSectionsRedirecter.new(
+  base_path: base_path,
+  destination: destination,
+).redirect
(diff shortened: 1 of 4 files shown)


h-lame:
Does this deal with the `<manual-slug>/updates` route ?  I don't think it does, but TBH I'm not sure, maybe the updates route is in the `:links` in the content store?  It's fairly confusing that this works over many different levels of the data - uses the content store to get links, directly uses `ManualRecord` and `SpecialistDocumentEdition` rather than the repository objects in the wiring.  But!  It's arguably simpler for doing it this way.  As we're re-writing SP anyway, maybe it's not a problem to just Get It Working™ for now and make sure the re-write includes similar rake tasks (or UI) for this functionality.

benilovj:
Most command-line utilities in this app seem to be triggered from `bin`, rather than `rake`, and having these two places has caused confusion this week. I'm planning to create a PR which removes `lib/tasks/publishing_api.rake` to simplify things(the other tasks in this file are obsolete, I believe). Is it possible to turn this into a `bin` script?

benilovj:
I've updated this code to turn the rake task into a bin script. This ought to be ready to merge – could somebody cast an eye on this? /cc @tuzz

benilovj:
Actually, just saw @h-lame's comment about the `<manual-slug>/updates` route... Closing until that's added.

chrisroos:
This hasn't been touched since 10 May 2016 so I'm deleting the branch.

PR #677, opened 2016-03-14
Use document type

Rosa-Fox:


app/presenters/document_presenter.rb
```diff
@@ -12,6 +12,8 @@ def to_json
       base_path: document.base_path,
       title: document.title,
       description: document.summary,
+      document_type: document.document_type,
+      schema_name: document.schema_name,
       format: "specialist_document",
```
tijmenb: I think you can't specify the `format` if you send `document_type` & `schema_name`.

Rosa-Fox: Hi @tijmenb,

I closed the PR for a bit as I realised I am not sure where the `schema_name` needs to come from... I assumed it would be the name of the schema relevant to the format in the `schemas` directory but still trying to figure out how to get it there and what it is used for. If you are able to run through it that would be great, otherwise I will keep looking at it.

There seems to be a `format`, `document_type` and `schema_name` in [this spec](https://github.com/alphagov/publishing-api/blob/3c84604be392c13d2f7307dfbda93175a6f31264/spec/support/request_helpers/mocks.rb#L21) but not too sure.

PR #678, opened 2016-03-15
Use document type

Rosa-Fox:
publishing-api now supports a top-level document_type and schema_name field. This commit modifies specialist-publisher so that these fields are sent to publishing-api in the document payload.
specialist-publisher index pages can now use document_type and schema_name to query documents.


danielroseman:
This is good so far, but it doesn't do the second part, which is querying by that type on the index page. This will probably involve changing the `all` classmethod in Document to use the document_type and remove the manual filtering there.

christophwong:
the last two commits are quite huge due to a Gem upgrade. The main change in logic is in app/models/document.rb in "Document#all"

PR #679, opened 2016-03-15
Redirect a manual and sections Publishing API V2

brenetic:
This performs the same function as the task in [this PR](https://github.com/alphagov/specialist-publisher/pull/676) but is using the version 2 of the publishing API.

- Adds a rake task wrapped around a utility class
- Passing a base_path and destination will find the manual with the slug matching the base_path and redirect it and the associated sections by publishing redirects to the publishing api.

eg. bundle exec rake publishing_api:redirect_manual_and_sections[/manual-url,/redirected-url]

with @steventux

### DIFF
--- a/lib/manual_and_sections_redirecter.rb
+++ b/lib/manual_and_sections_redirecter.rb
@@ -1,0 +1,71 @@
+class ManualAndSectionsRedirecter
+  def initialize(args)
+    @publishing_api = SpecialistPublisher.services(:publishing_api)
+    @logger = args[:logger] || STDOUT
+    @base_path = args[:base_path]
+    @destination = args[:destination]
+  end
+
+  def redirect
+    run(redirect: true)
+  end
+
+  def report
+    run(redirect: false)
+  end
+
+private
+
+  attr_reader :base_path, :destination, :logger, :publishing_api
+
+  def run(redirect:)
+    manuals_response = publishing_api.get_content_items(publishing_app: "specialist-publisher", document_type: "manual", fields: %w(base_path content_id))
+
+    raise "Could not retrieve specialist-publisher manuals from publishing api" unless manuals_response.code == 200
+
+    manual = manuals_response.to_hash["results"].find { |item| item["base_path"] == base_path }
+
+    raise "Could not find manual with base_path: #{base_path}" unless manual
+
+    manual_content_id = manual["content_id"]
+
+    redirect_item(manual_content_id, redirect_payload(base_path))
+    redirect_item(manual_content_id, redirect_payload("#{base_path}/updates"))
+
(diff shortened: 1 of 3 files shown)


danielroseman:
I don't understand why a separate v2 task is needed. The v2 api replaces the original API; is the original PR going to be closed?

On a more practical level, we should not be using the content-store at all here; publishing-api supports has an endpoint for getting content items and links by content_id. Publishing apps should not query the content store.

steventux:
2nd line and the core team have had a number of requests to [redirect manuals](https://govuk.zendesk.com/agent/tickets/1268328), as an interim measure https://github.com/alphagov/specialist-publisher/pull/676 was made to allow us to easily publish redirects for manuals and the associated sections. There's currently no task to do this which covers both manuals and sections, often the redirects are simply placed in router-data and this means the content-store is no longer truthful about the format of the content for the manual and sections at those base paths.
This PR echoes the same functionality of publishing redirects but tailored to the phase-two branch and publishing api v2 endpoints.

steventux:
As discussed with @danielroseman the content store call to get content should be replaced with 2 calls to the publishing api - one to get the content item from GET `/v2/content/<content_id>` and one from GET `/v2/linked/<content_id>` this should provide the same manual and section information.

lib/manual_and_sections_redirecter.rb
```diff
@@ -0,0 +1,74 @@
+require "gds_api/content_store"
+
+class ManualAndSectionsRedirecter
+  def initialize(args)
+    @publishing_api ||= SpecialistPublisher.services(:publishing_api)
+    @content_store ||= GdsApi::ContentStore.new(Plek.current.find("content-store"))
```
h-lame: Totally minor: we don't need optional assignment (`||=`) on these two lines because you can't set `@publishing_api` or `@content_store` before we call `initialize` so a simple assignment (`=`) will do.  The optional assignment will give someone looking at it pause wondering "is something special happening here" when the answer is "no" so while it's programatically harmless, it's still worth fixing.

h-lame: On closer inspection we no longer use the `@content_store` at all so we can remove it (and the `require` and the `attr_reader`) entirely - the comment still applies to `@publishing-api` though.

h-lame:
Might be worth checking with the specialist-publisher team if they have a story for this sort of functionality and folding this in with that.  We get these sorts of requests enough that it will be useful for the re-write to include it, but maybe we want it in the UI rather than a rake task.

brenetic:
@h-lame Rebased with changes.

benilovj:
This has been moved to https://github.com/alphagov/specialist-publisher-rebuild/pull/702; closing.

PR #680, opened 2016-03-18
Add attachment to specialist-document

mathildathompson:
- Attachment uploaded to asset-manager
- Attachment added to specialist-document content-schema (nested inside details hash)
- Attachment controller added to render new attachment form and handle attachment being saved inside specialist-document content-item
- Feature and unit test for attachment_presenter, attachment model and document model

@danielroseman I have currently only added: "#save! with attachments", "#upload" to cma_case_spec, I want to check this is correct before I duplicate it inside all the other tests;

- Do I need to add controller test, or are the feature test sufficient to cover this?
- I still need to complete attachment edit functionality, but there is a lot in here so though it would be a good idea to get it reviewed;


app/controllers/attachments_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  def edit
+
+  end
+
+  private
+  def fetch_document
+    begin
```
danielroseman: There's no need to use explicit `begin`/`end` when it encompasses the whole method.

app/models/document.rb
```diff
@@ -222,6 +224,19 @@ def publish!
     end
   end
 
+  def upload(attachment)
```
danielroseman: This doesn't feel like it belongs on the model; it's quite tightly linked to the request, since it relies on the params, so might be better in the controller.

app/models/document.rb
```diff
@@ -125,6 +126,7 @@ def self.from_publishing_api(payload)
     # and set it as the document's change note
     document.change_note = payload.details.change_history.pop["note"] if document.redrafted? && payload.details.change_history.length > 1
 
+    document.attachments = payload.details.attachments.map{|attachment|Attachment.new(attachment)} if payload.details.attachments
```
danielroseman: This is a bit hard to read, maybe extract the map into a separate method.

app/presenters/document_presenter.rb
```diff
@@ -23,6 +23,7 @@ def to_json
         body: document.body,
         metadata: metadata,
         change_history: change_history,
+        attachments: (document.attachments.map{|attachment| AttachmentPresenter.new(attachment).to_json} if document.attachments) }.delete_if{ |k,v| v.nil?
```
danielroseman: Again this should really be extracted into a method.

app/views/attachments/new.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= form_for(@attachment, :url => document_attachments_path(params[:document_type], params[:document_content_id])) do |f| %>
```
danielroseman: Styling on this page doesn't seem right - we are missing the surrounding page elements. In the original app the form was in _form.html.erb and there is extra stuff in the new/edit views.

app/views/attachments/_form.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= form_for(@attachment, :url => document_attachments_path(params[:document_type], params[:document_content_id])) do |f| %>
+    <%= f.label :title %>
+    <%= f.text_field :title, class: 'form-control input-md-8' %>
+
+    <%= f.file_field :file %>
```
danielroseman: Need a `label` for this field as well.

app/views/attachments/new.html.erb
```diff
@@ -0,0 +1,4 @@
+
```
danielroseman: Needs <h1> with New attachment.

spec/controllers/attachments_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  describe "GET new" do
+    it "renders the new attachment form" do
+      get :new, document_type: document_type, document_content_id: document_content_id
+      expect(response).to render_template :new
+    end
+  end
+
+  describe "POST create" do
+    it "renders the specialist document edit page" do
```
danielroseman: "redirects to"

spec/controllers/attachments_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      document = CmaCase.find(document_content_id)
+      allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
+      stub_any_publishing_api_put_content
+      stub_any_publishing_api_patch_links
+      request = stub_request(:post, "#{Plek.find('asset-manager')}/assets").
```
danielroseman: The gds-api-adapters test helpers provide `asset_manager_has_an_asset` which would be useful here.

spec/controllers/attachments_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    it "renders the new attachment form" do
+      get :new, document_type: document_type, document_content_id: document_content_id
+      expect(response).to render_template :new
+    end
+  end
+
+  describe "POST create" do
+    it "renders the specialist document edit page" do
+      document = CmaCase.find(document_content_id)
+      allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
```
danielroseman: I'd prefer to stub `CmaCase.find` here.

spec/controllers/attachments_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+  end
+
+  describe "GET edit" do
+    it "renders the edit attachment form" do
+      document = CmaCase.find(document_content_id)
+      allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
+      get :edit, document_type: document_type, document_content_id: document_content_id, attachment_content_id: attachment_content_id
+
+      expect(response).to render_template :edit
```
danielroseman: More useful to check the assigns for `@document`.

spec/controllers/attachments_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      post :create, document_type: document_type, document_content_id: document_content_id, attachment: {file: Rack::Test::UploadedFile.new("spec/support/images/cma_case_image.jpg", "mime/type"), title: 'test attachment upload'}
+
+      expect(response).to redirect_to(edit_document_path(document_type: document_type, content_id: document_content_id))
+    end
+  end
+
+  describe "GET edit" do
+    it "renders the edit attachment form" do
+      document = CmaCase.find(document_content_id)
+      allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
```
danielroseman: Again better to stub the model. And since these are repeated it probably makes sense to extract into a before block in a shared context.

spec/models/cma_case_spec.rb
```diff
@@ -162,4 +218,30 @@ def cma_case_content_item(n)
       assert_rummager_posted_item(indexable_attributes)
     end
   end
+
+  describe "#upload" do
+    it "uploads an attachment" do
+
+      request = stub_request(:post, "#{Plek.find('asset-manager')}/assets").
```
danielroseman: again `asset_manager_has_an_asset` would be good

app/models/document.rb
mathildathompson: I have moved this method inside the controller. After some research, many peoples opinion is that private methods inside controllers should not be tested. I feel a bit uncomfortable leaving this method untested?

spec/controllers/attachments_controller_spec.rb
mathildathompson: Is this not stubbed in the before do block?

mathildathompson: We discuss removing the fetch_document stub, but I think we still need this because I need access the the document variable that it returns?

app/models/attachment.rb
```diff
@@ -0,0 +1,23 @@
...
+    @content_type = params[:content_type]
+    @url = params[:url]
+    @content_id = params[:content_id] || SecureRandom.uuid
+    @created_at = params[:created_at]
+    @updated_at = params[:updated_at]
+    @has_changed = false
+  end
+
+  def update_attributes(new_params)
+    new_params.each do |k, v|
```
dougdroper: Should these params be restricted in some way? Would throw a NoMethodError (or worse) if something unexpected was passed in?

app/models/attachment.rb
```diff
@@ -0,0 +1,23 @@
...
+      self.public_send(:"#{k}=", v)
+    end
+    self.has_changed = true
+  end
+end
```
dougdroper: I think git has issues with there being no new line at the end of files, I believe you can automatically add them on in your editor.

app/models/document.rb
```diff
@@ -222,6 +224,28 @@ def publish!
     end
   end
 
+  def upload(attachment)
+    begin
+      response = SpecialistPublisher.services(:asset_api).create_asset(:file => attachment.file)
+      attachment.url = response.file_url
+      attachment.content_type = attachment.file.content_type
+
+      unless attachment.has_changed
```
dougdroper: I think I would have a changed? method on attachment. More idiomatic?

app/presenters/attachment_presenter.rb
```diff
@@ -0,0 +1,26 @@
...
+      url: @attachment.url,
+      title: @attachment.title,
+      content_type: @attachment.content_type,
+      updated_at: updated_at,
+      created_at: created_at,
+      content_id: @attachment.content_id
+    }
+  end
+
+  private
```
dougdroper: Just needs a line below

dougdroper: Oh and I think for consistency in this project I think private is indented 2 spaces behind everything else.

gpeng: The [style guide](https://github.com/alphagov/styleguides/blob/master/ruby.md#classes) reckons `private` should be on the same indent level as class. I think the linter usually catches this. Is it enabled on this repo?

PR #681, opened 2016-03-21
Adding new manuals

christophwong:
[Trello Card](https://trello.com/c/uF0IfPyH)

As of now, an editor is able to add a manual and sections for the manual.
The functionality of publishing and editing has not been implemented.

The code functions in the following way:
On save, a manual is PUT to publishing API.
When saving a section, 3 requests are made: 
1. to put content for the section
2. to patch links for section with the content id of the manual to which it belongs.
3. to patch links for the manual with said section's content-id.

I have yet to implement the links for "organisations" but hope to open this PR for comments.

Also, I think it will be good if we can have a better strategy for dealing with content-items in tests, there seems to be huge duplications of data because of the need to stub a lot of api-calls.


app/controllers/manual_sections_controller.rb
```diff
@@ -12,6 +12,30 @@ def show
     Airbrake.notify(e)
   end
 
+  def new
+    @section = Section.new({manual_content_id: params["manual_content_id"]})
```
danielroseman: style issue: don't use {} when the only parameter is a hash.

app/controllers/manual_sections_controller.rb
```diff
@@ -12,6 +12,30 @@ def show
...
+  end
+
+  def create
+    section_data = params["section"]
+    section_data["manual_content_id"] = params["manual_content_id"]
+
+    @section = Section.new(section_data)
+
+    if @section.valid?
+      if @section.save!
```
gpeng: `save!` will raise if it fails so the `if\else` won't happen. `save` returns a boolean. It also checks `valid?` so you could shorten this down to `if save` and remove the outer `valid?` branch. If an error is raised along the way then it will fall through to the default error page which I would say is as informative as the error message you're trying to return. It will also presumably get reported back to Errbit.

app/controllers/manuals_controller.rb
```diff
@@ -19,4 +19,24 @@ def show
...
 
+  def new
+    @manual = Manual.new
+  end
+
+  def create
+    @manual = Manual.new(params["manual"])
+
+    if @manual.valid?
+      if @manual.save!
```
gpeng: Same here...

app/controllers/manual_sections_controller.rb
gpeng: Actually... we're not in ActiveRecord here are we... Possibly disregard that :)

app/models/section.rb
```diff
@@ -112,7 +116,38 @@ def self.from_publishing_api(content_id:)
...
+      begin
+        item_request = publishing_api.put_content(self.content_id, presented_section)
+        section_link_request = publishing_api.patch_links(self.content_id, presented_section_links)
+        item_request.code == 200 && updated_manual_links? && section_link_request.code == 200
+      rescue GdsApi::HTTPErrorResponse => e
+        Airbrake.notify(e)
+        false
+      end
+    else
+      false
```
gpeng: This is inconsistent with `Manual#save!` which raises a `RecordNotSaved`

gpeng:
Discussed offline with @christophwong. For consistency with the `ActiveRecord` API patterns it would probably make sense to implement `save` rather than `save!` here. The current `save!` implementation is returning a boolean or raising which is inconsistent with the AR interface, confusing and causing extra complexity in the controllers. If `save` is implemented to return a boolean then `save!` could be implemented as in AR if required e.g.

```
def save!
  save || raise(RecordNotSaved)
end

```

gpeng:
Looks good. Much better.

app/models/section.rb
```diff
@@ -112,7 +116,38 @@ def self.from_publishing_api(content_id:)
     section
   end
 
-private
+  def updated_manual_links?
```
gpeng: Just looking at this again... I wouldn't expect this method to update the Publishing API based on its name.

christophwong: I added in the question mark because it returns a boolean, but looking at it again, the method #save is a similar pattern. I will change it to "update_manual_links"

gpeng:
This looks pretty much good to go. One final thing though... There is inconsistent indentation in a lot of the files, particularly in json and html but some of the ruby too. It should be two spaces throughout but there are a lot of instances where four is used.

christophwong:
@gpeng hi I've updated this PR with some indentation fixes, please let me know if I missed something.

PR #683, opened 2016-03-23
[Merge after add attachments PR] refactor content items into payload

mathildathompson:
- Test were becoming very long and unreadable, so extracted the cma_case_content_item into a Payload;
- This branch was created form the add-attachments branch because it uses some of the code in it;


danielroseman:
Needs a rebase.

PR #685, opened 2016-03-29
Email alerts

christophwong:
Need to configure new env var EMAIL_ALERT_API_BEARER_TOKEN
When a finders document is published with an update_type of "major", it will notify email-alert-api to send email to the respective subscriber_list.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -217,7 +217,15 @@ class Document
         base_path,
         indexable_document.to_json,
       )
-      email_alert_api.send_alert({})
+
+      if self.update_type == "major"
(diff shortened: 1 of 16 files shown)


app/models/document.rb
```diff
@@ -218,6 +218,15 @@ def publish!
         indexable_document.to_json,
       )
 
+      if self.update_type == "major"
+        email_alert_api.send_alert(
+          {
```
danielroseman: This doesn't seem complete: don't we need the subject body of the email alert to be sent here? The existing app sends a whole lot of formatted HTML to email-alert-api.

PR #687, opened 2016-03-30
Email alerts, first pass

christophwong:
This PR is the first pass of speciality publisher integrating with email-alert-api.
[Trello Card](https://trello.com/c/ploRXLJV)
This PR adds feature that "On publishing a document with a major update, email-alert-api is triggered to send out notification emails.

I couldn't find an updated documentation of what the email-alert-api consumes so there isn't a contract test for what the payload should look like.
It currently has the keys `subject`, `body`, `links`, and `document_type`.

Testing I’ve done apart from the automated test, I ran email-alert-api locally and tailed the sidekiq logs, and on "publish with a major update for a finders document” I can see that it triggers an async-job to talk to GovDelivery (that’s where I had to stop without staging credentials for govdelivery).

still need to implement:
  - documents of type "medical-safety-alerts" uses a different template -- [reference](https://github.com/alphagov/specialist-publisher/blob/4350dfb395b334a0e895db9e3b55b997eb03e745/app/views/email_alerts/medical_safety_alerts/publication.html.erb#L26-L40)
  - documents of type "drug-safety-update" should NOT send email-alerts on publish or major updates.


app/presenters/email_alert_presenter.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def body
+    view_renderer.render(
+      template: "email_alerts/publication",
+      formats: ["html"],
+      locals:   {
+        document_title: document.title,
+        document_summary: document.summary,
+        document_url: [Plek.current.website_root, document.base_path].join(""),
```
gpeng: You should probably use `File.join` here. There is also `URI.join` but that's a bit temperamental. `File.join` would handle any missing slashes.

gpeng:
Is there are reason why all the payload files, methods etc use `"key" => val` syntax rather than the >1.9 `key: val`?

christophwong:
@gpeng I think the payloads are in the old style to look more like JSONs

gpeng:
I thought it would be something like that. I still think they should probably be in the preferred syntax but not a problem if you disagree unless anybody else has an opinion. One other thing... Have the email templates you've used just been copied from the `master` or are they new? Have they been tested?

christophwong:
@gpeng All the email templates are copied over from `master`.
In terms of testing, the templates and full end-to-end with GovDelivery are not done, as I have no access to GovDelivery for this app.

gpeng:
Ok... Fair enough. If there are no changes from `master` then all should be well in the world. I'm happy to merge unless anybody else has an opinion on the `Hash` syntax.

jackscotti:
I think you'll need to remove the style from the HTML files and add a stylesheet where required.

christophwong:
@jackscotti regarding those html files, I believe the reason why the styles are inline is because they aren't rendered in our app, but will be sent as a string to GovDelivery to send out.

tuzz:
Has the HTML email been constructed as part of this work or is it copied over from somewhere else? If it's been constructed, has it been tested to work in the majority of email clients?

The integration with Email Alert API looks ok. The fields you're passing through look correct (https://github.com/alphagov/email-alert-api/blob/master/app/controllers/notifications_controller.rb#L12-L16).

If you'd like to do a full end-to-end test on staging, I can help with providing access to govdelivery (I'm on 2ndline this week)

christophwong:
@tuzz the email templates are copied from master branch.
Let's do an end-to-end-test, then we can merge it.
Thank you everyone

PR #689, opened 2016-03-30
Editing manual sections

mathildathompson:
- Users can edit manual_section content; 
- We have added body field to the section show page;
[Trello Card](https://trello.com/c/3F0sXPVn/44-add-ability-to-edit-manual-sections)

### DIFF
--- a/app/controllers/manual_sections_controller.rb
+++ b/app/controllers/manual_sections_controller.rb
@@ -17,10 +17,7 @@ class ManualSectionsController < ApplicationController
   end
 
   def create
-    section_data = params["section"]
-    section_data["manual_content_id"] = params["manual_content_id"]
-
-    @section = Section.new(section_data)
+    @section = Section.new(section_params.merge(manual_content_id: params[:manual_content_id]))
 
     if @section.save
       flash[:success] = "Created #{@section.title}"
@@ -37,16 +34,19 @@ class ManualSectionsController < ApplicationController
 
   def update
     @section = Section.find(content_id: params[:content_id], manual_content_id: params[:manual_content_id])
-    params[:section].each do |k, v|
-      @section.public_send(:"#{k}=", v)
-    end
 
-    if @section.save
+    if @section.update_attributes(section_params)
       flash[:success] = "#{@section.title} has been updated"
       redirect_to manual_section_path(@section.manual.content_id, @section.content_id)
     else
       flash.now[:danger] = "There was an error updating #{@section.title}. Please try again later."
       render :edit
     end
+  end
+
+private
+
+  def section_params
+    params.require(:section).permit(:title, :summary, :body)
   end
 end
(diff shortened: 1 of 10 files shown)


spec/features/editing_a_manual_section_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    scenario 'editing a section of a manual' do
+      visit manual_path(manual_content_id)
+      click_link 'First section'
+      click_link 'Edit section'
+
+      expect(page.status_code).to eq(200)
+
+      expect(page).to have_field('Title')
```
davidbasalla: You could probably remove these `expect(page).to have_field(x)` lines, as the `fill_in` on the next line will also run a `find` on the element and fail if it can't find it.

app/controllers/manual_sections_controller.rb
```diff
@@ -17,10 +17,7 @@ def new
   end
 
   def create
-    section_data = params["section"]
-    section_data["manual_content_id"] = params["manual_content_id"]
-
-    @section = Section.new(section_data)
+    @section = Section.new(section_params.merge("manual_content_id" => params["manual_content_id"]))
```
davidbasalla: Prefer to use Ruby 1.9 hash syntax -> `manual_content_id: params["manual_content_id"]`

boffbowsh: Worth noting that it's only relevant in this case because `permit` (the source of `section_params`) returns a `HashWithIndifferentAccess`. If it was a normal `Hash`, the suggested version wouldn't be equivalent.

mathildathompson:
@davidbasalla thanks for the comments, it should be ready to merge now.

spec/features/editing_a_manual_section_spec.rb
```diff
@@ -31,7 +31,6 @@
 
       expect(page.status_code).to eq(200)
 
-      expect(page).to have_field('Title')
       fill_in "Title", with: "My updated first section"
```
davidbasalla: It would probably be better if you could make this change in the [original commit](https://github.com/alphagov/specialist-publisher/pull/689/commits/1f0346f0c7f627321d4b6a6d8dee7e088d841b83) where this was introduced to avoid confusion.

You can also remove the similar lines that check for 'Summary' and 'Body' fields.

mathildathompson:
@davidbasalla hopefully ready to merge now.

app/views/manual_sections/show.html.erb
```diff
@@ -15,6 +15,15 @@
   </div>
 </div>
 
+<% if @section.body.present? %>
```
benilovj: I think it'd be worth adding an assertion to `viewing_a_manual_section_spec` to check for the body being rendered...

benilovj:
Minor comment about test coverage, apart from that, :+1:

app/views/manual_sections/show.html.erb
benilovj: I've taken the liberty to add this assertion to the test (see a42ca66c055ade6b2ce62dc040de04e35f3466d9) and force-pushed the branch.

PR #692, opened 2016-04-01
Add attachments to manual sections

mathildathompson:
- We discovered inconsistencies between the manuals, section and documents #find and #form_publishing_api methods during development. Redesigned to use hash everywhere as opposed to ostruct; 
- Refactor manuals controller update action to use #update_attributes method;
- Create a manual_sections_attachment_controller. We initially wanted to use the same attachment controller, but because params and redirect URLs are different, we decided it would become too unreadable;
- manual_sections_attachment_controller logic identical to attachments controller;
- When you are reviewing, please let us know if there is anything else we can share between attachments and manual_section_attachments views. We are currently only sharing _attachments_form.html.erb. 
- [Trello Card](https://trello.com/c/ZK2nTIo6/43-handle-attachments)


h-lame:
Note: we've based this on #689 so we'll have to do a thing once that's merged.

danielroseman:
This is failing on a lint error.

h-lame:
Rebased.  If you're happy to merge this we can just close #689 as this includes those commits.

benilovj:
@h-lame I'm about to look at #689, please leave it open and I'll rebase this PR once it's merged.

app/models/attachment.rb
```diff
@@ -25,4 +26,9 @@ def update_attributes(new_params)
   def changed?
     @changed
   end
+
+  def self.all_from_publishing_api(payload)
```
benilovj: I'm not 100% sure about the name for this method. Initially when I saw it, I thought it would be similar to `Manual.all`, not extracting the attachments from the Publishing API payload.

What do others think?

benilovj: Oh, I see, it's similar to the `from_publishing_api` methods...

h-lame: Yeah, we originally had it as `from_publishing_api` but thought it needed something to differentiate because it would extract all the attachments, rather than just one.

benilovj:
:+1:

PR #697, opened 2016-04-07
Render Govspeak on save

christophwong:
[Trello](https://trello.com/c/y9wLTaUD)
Changing presenters so we send both Govspeak and HTML versions for the body of a specialist-document. This will break until govuk-content-schema is updated.

There was quite a big refactoring so this PR is best checks per commit. I've isolated the changes most directly related to the feature change in separates commits.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -104,12 +104,26 @@ class Document
     end
   end
 
+  def self.extract_body_from_payload(payload)
+    body_attribute = payload['details']['body']
+
+    case
+    when body_attribute.class == Array
+      govspeak_body = body_attribute.detect do |body_hash|
+        body_hash['content_type'] == 'text/govspeak'
+      end
+      govspeak_body['content']
+    when body_attribute.class == String
+      body_attribute
+    end
+  end
+
   def self.from_publishing_api(payload)
     document = self.new(
       content_id: payload['content_id'],
       title: payload['title'],
       summary: payload['description'],
-      body: payload['details']['body'],
+      body: extract_body_from_payload(payload),
       publication_state: payload['publication_state'],
       public_updated_at: payload['public_updated_at']
     )
--- a/spec/models/cma_case_spec.rb
+++ b/spec/models/cma_case_spec.rb
@@ -2,39 +2,17 @@ require 'spec_helper'
 
 RSpec.describe CmaCase do
   def cma_case_content_item(n)
-    {
-      "content_id" => SecureRandom.uuid,
+    Payloads.cma_case_content_item(
       "base_path" => "/cma-cases/example-cma-case-#{n}",
       "title" => "Example CMA Case #{n}",
       "description" => "This is the summary of example CMA case #{n}",
-      "document_type" => "cma_case",
-      "schema_name" => "specialist_document",
-      "publishing_app" => "specialist-publisher",
-      "rendering_app" => "specialist-frontend",
-      "locale" => "en",
-      "phase" => "live",
-      "public_updated_at" => "2015-11-16T11:53:30",
-      "publication_state" => "draft",
-      "details" => {
-        "body" => "## Header" + ("\r\n\r\nThis is the long body of an example CMA case" * 10),
-        "metadata" => {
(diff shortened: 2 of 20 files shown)


app/models/document.rb
```diff
@@ -109,7 +109,7 @@ def self.from_publishing_api(payload)
       content_id: payload['content_id'],
       title: payload['title'],
       summary: payload['description'],
-      body: payload['details']['body'],
+      body: payload['details']['body'][0]['content'],
```
danielroseman: I think we need to deal with the case where the body has the previous format of just a string, otherwise this will present a blank text field and overwrite the previous content.

danielroseman:
One comment, otherwise great.

app/models/document.rb
mathildathompson: Can you be sure the first element is always going to be govspeak.I remember is share-sale-publisher, we did something like this:
```
{ |format| format[:content_type] == "text/govspeak" }[:content]
```

christophwong:
@danielroseman It's updated
[src](https://github.com/alphagov/specialist-publisher/pull/697/files#diff-da766db961b1ebe7e7e8de051dac4096R107)
[spec](https://github.com/alphagov/specialist-publisher/pull/697/files#diff-1e4809daaea2c7f56ec14a37face6073R132)

app/models/document.rb
```diff
@@ -104,12 +104,23 @@ def humanized_attributes
     end
   end
 
+  def self.extract_body_from_payload(payload)
+    body_attribute = payload['details']['body']
+
+    case
+    when body_attribute.class == Array
+      body_attribute[0]['content']
```
danielroseman: We also need to take into account Matilda's comment, ie we don't know for sure that the govspeak is always the first element.

spec/models/cma_case_spec.rb
```diff
@@ -142,14 +120,27 @@ def cma_case_content_item(n)
...
+      expect(cma_case.body).to          eq(cma_cases[0]["details"]["body"][0]["content"])
       expect(cma_case.opened_date).to   eq(cma_cases[0]["details"]["metadata"]["opened_date"])
       expect(cma_case.closed_date).to   eq(cma_cases[0]["details"]["metadata"]["closed_date"])
       expect(cma_case.case_type).to     eq(cma_cases[0]["details"]["metadata"]["case_type"])
       expect(cma_case.case_state).to    eq(cma_cases[0]["details"]["metadata"]["case_state"])
       expect(cma_case.market_sector).to eq(cma_cases[0]["details"]["metadata"]["market_sector"])
       expect(cma_case.outcome_type).to  eq(cma_cases[0]["details"]["metadata"]["outcome_type"])
     end
+
+    it "should be able backward compatible for a single string representation of body in payload" do
```
mathildathompson: it "should be backward compatible for a single string representation of body in payload" ?

spec/models/cma_case_spec.rb
```diff
@@ -142,14 +120,27 @@ def cma_case_content_item(n)
...
+      simple_cma_case_payload = Payloads.cma_case_content_item("details" => { "body" => "single string body" })
+      publishing_api_has_item(simple_cma_case_payload)
+
+      content_id = simple_cma_case_payload["content_id"]
+      cma_case = described_class.find(content_id)
+
+      expect(simple_cma_case_payload["details"]["body"].class).to eq(String)
```
mathildathompson: Is line 139 required? Does 142 not test the same thing?

spec/models/cma_case_spec.rb
christophwong: Can you please elaborate what you find problematic with this description?

spec/models/cma_case_spec.rb
christophwong: I am not sure whether it is required and they don't actually test the same things.

I was worried that there could be a risk on 142 that they are both arrays and are equal and the test gives a false positive.
What do you think?

christophwong:
This pull request has been migrated to a new repo https://github.com/alphagov/specialist-publisher-rebuild/pull/695

PR #703, opened 2016-04-15
Remove the Publishing API-related tasks

benilovj:
As far as I understand, these tasks publish content for the first time
to the Publishing API. They have already been run on production and
any republishing to Publishing API is supposed to be done with the
scripts in `bin` folder.

These tasks should be deleted to avoid confusion with what's in `bin`.
I've also deleted the classes they make direct calls to, because I couldn't
find any other references to them (apart from the specs).

/cc @jennyd (does the above sound right?)


h-lame:
The `publishing_api:publish_finders` doesn't have an equivalent in `./bin`.  It's used to publish the finders that live as static json in `./finders` rather than publish documents stored in the DB.  I think it's used as part of the deploy.

I don't think the two other tasks, `publishing_api:publish_documents_as_placeholders` and `publishing_api:draft:publish_manuals` have equivalents either, although I'm less sure why they exist.  

I do think it's sensible to rationalise from having things in `bin` and `lib/tasks` is a worthy goal but this isn't the right way unfortunately.  I'd pull everything out of `bin` and into `lib/tasks` because it's a more "standard" way of doing things, but this repo can't have much time left so it might not be worth the effort.

benilovj:
@h-lame that's fair feedback.

* I'll back out the `finders` deletion, and turn it into a script in `bin` instead, so there's at least a bit more consistency
* I am still keen to remove `publishing_api:publish_documents_as_placeholders` and `publishing_api:draft:publish_manuals` as I believe them to be one-off scripts that aren't documented in the Ops manual (and nobody I've spoken to thought otherwise)

h-lame:
That sounds fair.

benilovj:
@h-lame I've updated this PR as proposed this morning. I also migrated another rake task to a script for consistency (the push of finders to Rummager).

h-lame:
:+1: - if we suddenly remember that we need the others, they're in git.

PR #707, opened 2016-04-27
Duplicate deleter: clean up all unknown content IDs

benilovj:
This is a follow-up to #706.

Before this change, the duplicate deleter script didn't get rid of
all duplicate editions if there were multiple duplicates with the same
content ID (eg if a publisher archived a duplicate, there would be
one archived edition and one draft edition with the same content ID).

After this change, if a content ID is unknown to Publishing API,
the script will delete all editions with that content ID.

/cc @h-lame


h-lame:
Glad it was our fault and not another mongo weirdness that was the cause here. 👍

PR #717, opened 2016-05-25
Send organisation links to publishing api

MatMoore:
Each type of specialist document has a fixed list of organisations associated with
it. We pass these to rummager already, but not publishing api.

https://trello.com/c/oqjOp8jd/639-specialist-publisher-documents-don-t-have-organisations-in-the-publishing-api

### DIFF
--- a/app/lib/specialist_publisher.rb
+++ b/app/lib/specialist_publisher.rb
@@ -42,6 +42,54 @@ private
     "vehicle_recalls_and_faults_alert" => VehicleRecallsAndFaultsAlertObserversRegistry,
   }.freeze
 
+  ORGANISATIONS = {
+    "aaib_report" => %w(
+      38eb5d8f-2d89-480c-8655-e2e7ac23f8f4
+    ), # air-accidents-investigation-branch
+    "asylum_support_decision" => %w(
+      7141e343-e7bb-483b-920a-c6a5cf8f758c
+    ), # first-tier-tribunal-asylum-support
+    "cma_case" => %w(
+      957eb4ec-089b-4f71-ba2a-dc69ac8919ea
+    ), # competition-and-markets-authority
+    "countryside_stewardship_grant" => %w(
+      8bf5624b-dec2-44fa-9b6c-daed166333a5
+      de4e9dc6-cca4-43af-a594-682023b84d6c
+      d3ce4ba7-bc75-46b4-89d9-38cb3240376d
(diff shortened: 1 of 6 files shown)


app/lib/specialist_publisher.rb
```diff
@@ -51,6 +99,6 @@ def document_repositories
   end
 
   def observer_registry(document_type)
-    OBSERVER_MAP.fetch(document_type).new
+    OBSERVER_MAP.fetch(document_type).new(ORGANISATIONS.fetch(document_type, {}))
```
MatMoore: TODO: Default is wrong

tijmenb:
Would we be able to use the metadata? https://github.com/alphagov/specialist-publisher/blob/master/finders/metadata/aaib-reports.json#L12

benilovj:
Would you be able to add the `links` part the assertions in https://github.com/alphagov/specialist-publisher/blob/master/spec/features/republishing_documents_spec.rb, so those assertions are more exhaustive?

MatMoore:
Thanks for the comments @benilovj and @tijmenb. I'll do both those things.

MatMoore:
One thing I haven't addressed in this PR is the links will be sent for drafts as well as published content.

Normally we would only send the links at publish time because otherwise the user can make changes to a draft that have immediate effect. In this case though I think it's fine because the organisations aren't controlled by the user and won't change after the content is published for the first time.

I'm assuming that we will disable organisation tagging in content tagger, since the organisations are meant to be fixed.

MatMoore:
I've updated the republishing documents spec to check for links.

@tijmenb I started trying to use the metadata JSONs but I had trouble matching them to the specialist publisher models.

Mostly these appear to map 1-1, but If you go by the filenames, you need to account for plurals and underscores. Even then there are some exceptions:

- *asylum-support-tribunal-decisions* vs *asylum_support_decision*
- *utaac_decision* is missing a metadata file

MatMoore:
We discussed this offline and we're not going to try to reuse the metadata JSON for determining the organisations to tag to.

The metadata is used at the moment when publishing the finders itself, but when publishing the specialist documents themselves, the organisations sent to rummager come from hard coded slugs in the *indexer_formatter classes. The content ids I've hard coded for publishing-api are based on these slugs.

benilovj:
👍  from me (at least as sure as one can be on this code base...)

tijmenb:
👍 great.

PR #738, opened 2016-09-09
Remove taggy things from manuals

tijmenb:
This PR is basically a revert of https://github.com/alphagov/specialist-publisher/pull/532. That PR changed the publishing flow so that this app would fetch the topics/specialist sectors from the content-api before publishing to rummager.

This was needed at the time, because otherwise the payload sent to rummager wouldn't contain the tags set in panopticon, and the manual wouldn't be displayed on a /topic page.

Since then, we've started indexing the links from the publishing-api in rummager. As part of this we stopped sending specialist sectors to rummager (https://github.com/alphagov/specialist-publisher/pull/723).

At the time of that PR we didn't realise that the only reason for the fetching of tags from content-api was now gone. This PR removes the last (now basically unused) code, which means that we are no longer dependent on content-api.

✌️


mobaig:
👏  👍

PR #745, opened 2016-09-20
Bump Ruby version and various other gems

klssmith:
Bumped Ruby to 2.3.1, which resulted in a number of other gems also needing to be bumped. Some changes to the tests were also necessary to prevent warnings and failures which resulted from the updated gems.

There is still one warning remaining when running the tests - `The 'safe' write concern option has been deprecated in favor of 'w'.` This appears to relate to Mongoid, but upgrading this requires upgrading govuk_content_models, which then requires Rails to be upgraded to Rails 4. As a result, this warning has been left for now.

[Trello card](https://trello.com/c/RyruKiPM/494-upgrade-apps-to-supported-ruby-versions-update-at-stand-up)


klssmith:
I've kept each gem upgrade as a separate commit for now as there are a lot of them, but am happy to merge these all into one very large commit so that there are no failing commits in Jenkins if this seems better.

sihugh:
if you run `bundle exec rake` on master, it runs all the tests.  On this branch, it seems to have stopped.  It might be worth tweaking the [default rake task](https://github.com/alphagov/manuals-publisher/blob/98877cc4e07dc6e87d258993ccd1df2d77a70634/Rakefile#L15) or something?

Looks good otherwise!  (Also needs a rebase now)

klssmith:
Thanks @sihugh , I've changed the default rake task order and rebased. The tests seem to all be running on my machine now, but if you could test that it works for you too that would be great.

sihugh:
👍

PR #754, opened 2016-09-28
Update name for Democratic Republic of the Congo

Davidslv:
**trello**: https://trello.com/c/rxjklHjL/191-country-name-change-democratic-republic-of-the-congo
>The country formerly known as 'Democratic Republic of Congo' is now officially called 'Democratic Republic of **the** Congo'.

### DIFF
--- a/finders/schemas/international-development-funds.json
+++ b/finders/schemas/international-development-funds.json
@@ -34,7 +34,7 @@
           "value": "burma"
         },
         {
-          "label": "Democratic Republic of Congo",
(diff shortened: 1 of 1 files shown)


rubenarakelyan:
Will changing the "value" cause any issues with existing content?

Davidslv:
@rubenarakelyan you right it will, so I've reverted to be in line with what was agreed on the PR in Rummager: https://github.com/alphagov/rummager/pull/700

(we agreed to keep the value as is)

rubenarakelyan:
👍

PR #757, opened 2016-10-19
Manuals to use V2 of the Publishing API

kevindew:
This updates the API methods to use V2 of the Publishing API rather than the V1 endpoints. Thus the approach has changed somewhat from pushing full content items to instead push only a full representation to PUT content and just a small one to publish and unpublish. This has meant creating new exporter classes that represent this distinction.

As part of this work I've:
- Changed it so the publish button only shows when you have a draft (you receive an error on V2 of the Publishing API if you try to publish a published item)
- Updated published state to be draft whenever you update an item (I assume it's a bug that this doesn't happen already as the submit button says "Save as Draft")
- Discard drafts from Publishing API when the delete task is run


tijmenb:
You'll also need to change https://github.com/alphagov/manuals-publisher/blob/0654c51223819b2a88ab0f473245a6d4f6fa0748/spec/support/govuk_content_schema_helpers.rb#L5 to `publisher_v2` so the contract testing uses the new schemas.

kevindew:
@tijmenb Thanks 👍

spec/exporters/manual_section_publishing_api_links_exporter_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      minor_update?: true,
+      attributes: { body: "##Some heading\nmanual section body" },
+      attachments: []
+    )
+  }
+
+  it "exports links for the document" do
+    subject.call
+
+    expect(export_recipent).to have_received(:call).with(
```
tijmenb: If you want to be 100% correct, you can also test the payload against the schemas with the `be_valid_against_links_schema` helper.

spec/lib/publishing_api_finder_publisher_spec.rb
```diff
@@ -2,6 +2,12 @@
 require "publishing_api_finder_publisher"
 
 describe PublishingApiFinderPublisher do
+  before do
+    @schema_type = GovukContentSchemaTestHelpers.configuration.schema_type
+    GovukContentSchemaTestHelpers.configuration.schema_type = "publisher"
```
tijmenb: This means we won't be able to remove the old v1 schemas from content schemas. We could also remove this test or `skip: true` it?

features/support/publishing_api_helpers.rb
```diff
@@ -1,14 +1,20 @@
 require "gds_api/test_helpers/publishing_api"
+require "gds_api/test_helpers/publishing_api_v2"
 
 module PublishingAPIHelpers
   include GdsApi::TestHelpers::PublishingApi
```
tijmenb: I think we don't need `GdsApi::TestHelpers::PublishingApi` anymore. If we remove it we can get rid of the code in gds-api-adapters.

spec/lib/publishing_api_finder_publisher_spec.rb
kevindew: That file is on it's way out in [this PR](https://github.com/alphagov/manuals-publisher/pull/755) 🎉 - so it's just whatever keeps this green 'til then really.

spec/exporters/manual_section_publishing_api_links_exporter_spec.rb
kevindew: cool - I'll check that out.

features/step_definitions/deleting_manuals_steps.rb
```diff
@@ -23,6 +23,8 @@
 
 Then(/^the manual and its documents are deleted$/) do
   check_manual_does_not_exist_with(@manual_fields)
+  check_draft_has_been_discared_in_publishing_api(@manual.id)
```
danielroseman: typo quibble: "discarded" (and below)

features/support/publishing_api_helpers.rb
```diff
@@ -11,8 +11,10 @@ def stub_publishing_api
     stub_any_publishing_api_put_content
     stub_any_publishing_api_patch_links
     stub_any_publishing_api_publish
-    # This needs adding to gds-api-adapters
+    # FIXME - Update these methods once https://github.com/alphagov/gds-api-adapters/pull/611
+    # is merged
```
danielroseman: that's been merged, this can be updated.

features/step_definitions/deleting_manuals_steps.rb
kevindew: Good catch 💥

kevindew:
@danielroseman should be good to merge now - can either do it now (assuming still looks sane) and put another PR in once GDS API Adapters is building again or wait until that's sorted and include it in this.

spec/lib/publishing_api_finder_publisher_spec.rb
tijmenb: 🎉

features/support/publishing_api_helpers.rb
tijmenb: 👍

kevindew:
OK updated GDS API Adapters now, gave it another test on integration. Think it's all looking good.

PR #759, opened 2016-10-24
Fix content schema tests

kevindew:
These were broken by the change in path to two of the exporter files. 

I've also added a schema test for a manuals links, and included both links tests in the schema tests.

### DIFF
--- a/jenkins-schema.sh
+++ b/jenkins-schema.sh
@@ -39,13 +39,7 @@ git checkout $SCHEMA_GIT_COMMIT
 cd ../..
 
 time bundle install --path "${HOME}/bundles/${JOB_NAME}" --deployment
-# TODO as schemas are added for them, change this to include more formats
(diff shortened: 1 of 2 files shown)


jenkins-schema.sh
```diff
@@ -42,8 +42,10 @@ time bundle install --path "${HOME}/bundles/${JOB_NAME}" --deployment
 # TODO as schemas are added for them, change this to include more formats
 RAILS_ENV=test GOVUK_CONTENT_SCHEMAS_PATH=tmp/govuk-content-schemas time bundle exec rspec \
   spec/lib/publishing_api_finder_publisher_spec.rb \
-  spec/manual_publishing_api_exporter_spec.rb \
-  spec/manual_section_publishing_api_exporter_spec.rb \
```
tijmenb: I'd recommend to just run all the tests in this script, so this won't happen again.

PR #760, opened 2016-10-25
Prevent hash responses

steventux:
Reverts the change made to accommodate hash responses as this will not work
with the `Attachable` trait in `govuk_content_models`. The upgrade path here is
impractical despite the fix being relatively simple.
Adds `GdsApi.config.hash_response_for_requests = false` to workaround this problem.


tijmenb:
What's the problem with attachable? The plan is to remove the ostruct behaviour entirely so curious to see.

steventux:
https://github.com/alphagov/govuk_content_models/blob/master/app/traits/attachable.rb#L65 doesn't expect a hash response.
Simple enough to fix until an older codebase like this needs to be updated several major versions of content models, gds-sso dependencies, mongoid major version bumps etc etc, given there's a bright new future for manuals on the horizon this is the practical route to unblocking this https://github.com/alphagov/manuals-publisher/pull/755 which gets rid of a few hexagons.

PR #764, opened 2016-11-04
Publication logs for major manual updates only

NeilvB:
[Trello card](https://trello.com/c/x6vSw5LM/55-update-history)

Minor manual changes intermingled with major ones are confusing to users.

This PR prevents minor updates from creating publication logs for manuals, and also implements a script for rebuilding all historical logs to exclude minor updates.

### DIFF
--- a/app/observers/manual_observers_registry.rb
+++ b/app/observers/manual_observers_registry.rb
@@ -48,6 +48,7 @@ private
     ->(manual, _: nil) {
       manual.documents.each do |doc|
         next unless doc.needs_exporting?
+        next if doc.minor_update?
 
         PublicationLog.create!(
           title: doc.title,
--- a/bin/republish_manuals
+++ b/bin/republish_manuals
@@ -9,13 +9,22 @@ logger.formatter = Logger::Formatter.new
 
 repository = ManualsPublisherWiring.get(:repository_registry).manual_repository
 
-count = repository.all.count
+begin
+  count = repository.all.count
+rescue DocumentAssociationMarshaller::RemovedDocumentIdNotFoundError
+  count = "?"
+end
 
 logger.info "Republishing #{count} manuals..."
 
 repository.all.each.with_index do |manual, i|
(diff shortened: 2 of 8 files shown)


app/observers/manual_observers_registry.rb
```diff
@@ -48,6 +48,7 @@ def publication_logger
     ->(manual, _: nil) {
       manual.documents.each do |doc|
         next unless doc.needs_exporting?
+        next if doc.minor_update?
```
h-lame: I don't know if there's an existing test for this functionality, but I think it'd be good if there was one to make it clear this is expected behaviour.

NeilvB: Added publishing test https://github.com/alphagov/manuals-publisher/blob/a265bd135c07e53862c8cee8af88f08ffce33afe/spec/features/publishing_manuals_spec.rb#L1

bin/republish_manuals
```diff
@@ -9,13 +9,22 @@ logger.formatter = Logger::Formatter.new
...
 logger.info "Republishing #{count} manuals..."
 
 repository.all.each.with_index do |manual, i|
-  logger.info("[% 3d/% 3d] id=%s slug=%s" % [i + 1, count, manual.id, manual.slug])
-  ManualServiceRegistry.new.republish(manual.id).call
+  begin
+    logger.info("[ #{i} / #{count} ] id=#{manual.id} slug=#{manual.slug}]")
+    ManualServiceRegistry.new.republish(manual.id).call
+  rescue DocumentAssociationMarshaller::RemovedDocumentIdNotFoundError => e
+    logger.warn("Manual with id=#{manual.id} slug=#{manual.slug} has removed document which was not found: #{e.message}")
```
NeilvB: Thanks. Updated message and used 'error' log level in https://github.com/alphagov/manuals-publisher/blob/44d961a9ec26f00d137ce86e5f14d1c4ae7601fd/bin/republish_manuals#L25

NeilvB:
Thanks for reviewing this @h-lame. I've force-pushed changes as you suggested. But I renamed the branch as I wasn't able to re-open the same PR following a force-push.

PR #765, opened 2016-11-04
Migrate guidance/capital-funding-guide-hca

steventux:
https://trello.com/c/EQScfnzQ/712-delete-rename-hca-manual

to `guidance/capital-funding-guide`
This moving the existing manual out of the way then changing the slug of the manual we wish to move.

### DIFF
--- a/db/migrate/20161031154415_remove_redirect_capital_funding_guide.rb
+++ b/db/migrate/20161031154415_remove_redirect_capital_funding_guide.rb
@@ -1,9 +1,14 @@
 require "manual_relocator"
+require "cli_manual_deleter"
+require "stringio"
 
 class RemoveRedirectCapitalFundingGuide < Mongoid::Migration
   def self.up
+    bad_manual_id = "7228d773-a416-4ec1-9549-5fb91b1caaf5"
     manual_slug = "guidance/capital-funding-guide"
     new_manual_slug = "#{manual_slug}-hca"
+
+    CliManualDeleter.new(manual_slug, manual_id: bad_manual_id, stdin: StringIO.new("y")).call
 
     ManualRelocator.move(new_manual_slug, manual_slug)
   end
--- a/db/migrate/20161031154415_remove_redirect_capital_funding_guide.rb
+++ b/db/migrate/20161031154415_remove_redirect_capital_funding_guide.rb
@@ -1,3 +1,20 @@
+# The history
+# ===========
+#
+# The Homes and Communities Agency have a manual "The Capital Funding Guide"
+# that they wanted to make extensive revisions to.  Removing sections
+# wasn't possible when they made their changes so they created a new version of
+# the manual at a new slug and worked on that.  They asked if we could remove
+# the existing one and move the new one onto that old slug.  That wasn't easy
+# so we issued some redirects to send visits to the old slug over to the
+# new one.  Now that we're ready to tidy this up, we are planning to remove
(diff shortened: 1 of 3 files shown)


h-lame:
I've tested this on dev and it seems to do the right thing.  Perhaps @steventux or @kevindew might take a look to see if there's something missing?

lib/manual_relocator.rb
```diff
@@ -0,0 +1,352 @@
...
+
+    puts "Issuing gone for #{old_manual.manual_id}"
+    send_gone(old_manual.manual_id, old_manual.slug)
+  end
+
+  def old_sections_reused_in_new_manual
+    @old_sections_reused_in_new_manual ||= _calculate_old_sections_reused_in_new_manual
+  end
+
+  def _calculate_old_sections_reused_in_new_manual
```
steventux: What's the intent (presumably a convention) of method names beginning with an underscore? I'm asking out of plain old ignorance.

h-lame: I use them to indicate that even if it's a private method, it's not one that you should be calling yourself.  If calculating the old sections to be reused was simpler this method wouldn't exist and it'd all be inlined in the memoized method above.

It's a habit I picked up when I did loads of python (double underscore methods are private) but I probably shouldn't keep doing it in ruby as I have to keep explaining it.

steventux:
This looks good to me, I didn't realise how complicated this would be.
I will run it against fresh data and verify.

db/migrate/20161031154415_remove_redirect_capital_funding_guide.rb
```diff
@@ -1,3 +1,20 @@
+# The history
+# ===========
+#
+# The Homes and Communities Agency have a manual "The Capital Funding Guide"
+# that they wanted to make extensive revisions to.  Removing sections
+# wasn't possible when they made their changes to they created a new version of
```
kevindew: s/to/so/ ?

db/migrate/20161031154415_remove_redirect_capital_funding_guide.rb
```diff
@@ -1,10 +1,15 @@
...
+require "cli_manual_deleter"
+require "stringio"
 
 class RemoveRedirectCapitalFundingGuide < Mongoid::Migration
   def self.up
+    bad_manual_id = "7228d773-a416-4ec1-9549-5fb91b1caaf5"
     manual_slug = "guidance/capital-funding-guide"
     new_manual_slug = "#{manual_slug}-hca"
 
+    CliManualDeleter.new(manual_slug, manual_id: bad_manual_id, stdin: StringIO.new("y")).call
```
kevindew: sneaky 👍

lib/manual_relocator.rb
```diff
@@ -224,6 +220,30 @@ def send_draft(manual, manual_edition, document_editions)
     end
   end
 
+  def send_gone(document_id, slug)
+    # We should be able to use
+    #   publishing_api.unpublish(document_id, type: 'gone')
+    # here, but that doesn't leave the base_path in a state where
+    # publishing_api will let us re-use it.  Sending a draft gone object
+    # and then publishing it does though.  Might want to check if we can
+    # go back to the unpublish version at some point though.
```
kevindew: We've got a card for this on our planning board: https://trello.com/c/4RRvSMeQ/316-update-substitution-helper

PR #766, opened 2016-11-07
Remove removed_document_ids entries that point to deleted documents

h-lame:
For: https://trello.com/c/4aFn7Vji/54-delete-sections

At some point in the past, we fully deleted some sections from a manual
so that they no longer exist.  Unfortunately we left references to their
ids in the removed_document_ids of the manual they used to belong to. This
means that we can't republish that manual, or do other publishing tasks as
publishing does things with the removed_document_ids.

To solve this we run a migration that finishes the job by removing any ids
in removed_document_ids that no longer point to an instance in
SpecialistDocumentEdition.  We have done some manual work to check that the
ids we remove don't still exist in the publishing-api and they don't so our
cleanup is all we need to do.

### DIFF
--- a/db/migrate/20161107113454_remove_deleted_removed_document_ids_from_manuals.rb
+++ b/db/migrate/20161107113454_remove_deleted_removed_document_ids_from_manuals.rb
@@ -1,0 +1,43 @@
+# The history
+# ===========
+#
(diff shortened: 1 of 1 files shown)


NeilvB:
Looks good!

PR #768, opened 2016-11-09
Fixes for rebuild manual publication log script

NeilvB:


h-lame:
Don't really understand why the `to_a` is needed on `manuals_record`, but I suspect it's because `.each` on a mongo query doesn't return an enumerable like it does on an array, so we can't chain `with_index` onto it.  Fixes the problems though - :+1:

PR #769, opened 2016-11-09
Withdraw manuals sections via the ui

h-lame:
For: https://trello.com/c/4aFn7Vji/54-delete-sections

This extends the existing bin script functionality for withdrawing manual sections to the UI (for gds editors and editors only) and improves it to allow removal of published sections.  We also make sure that the publishing-api is kept up-to-date with these changes, something the bin script version didn't do (mostly because it only had to deal with drafts so the inconsistency wasn't a huge problem).

There are some changes that require a bit of thought, so best to review this commit by commit as there are useful notes.


h-lame:
Flagged as DNM for now, as I want to investigate existing removed documents and what the publishing-api thinks of them.  We may need a final commit with a script to run to tidy up existing removed_documents.

spec/controllers/manuals_controller_spec.rb
```diff
@@ -3,11 +3,12 @@
 describe ManualsController, type: :controller do
   describe "#publish" do
     let(:manual_id) { "manual-1" }
+    let(:services) { spy(AbstractManualDocumentServiceRegistry) }
```
NeilvB: Do you think it might be worth adding a `context` block around these (while we're here) to indicate that our set up involves a user who doesn't have permission to publish. I'm thinking of something like:

`context "when the user lacks the publish permission" do ... `

db/migrate/20161110113139_tidy_up_removed_documents.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  def self.publishing_api_v2
+    ManualsPublisherWiring.get(:publishing_api_v2)
+  end
+
+  def self.down
+    # Whilst it would be possible to reverse this, it would be a lot of work
+    # for something that is unlikely to ever get run.
+    # raise IrreversibleMigration
```
NeilvB: Comment typo?

db/migrate/20161110113139_tidy_up_removed_documents.rb
```diff
@@ -0,0 +1,99 @@
...
+
+class TidyUpRemovedDocuments < Mongoid::Migration
+  def self.up
+    ManualRecord.all.to_a.each do |manual_record|
+      puts %{Collecting removed_document_editons from "#{manual_record.slug}":#{manual_record.manual_id}}
+      removed_document_ids = manual_record.editions.map do |manual_edition|
+        manual_edition.removed_document_ids || []
+      end.flatten.uniq
+      puts %{Found #{removed_document_ids.count} removed documents}
+      removed_document_ids.map do |document_id|
```
NeilvB: I'm nitpicking but I don't think we need `map` here as we're not using the returned array. We could use `each`?

NeilvB:
This looks good overall - I added a few small comments. Well done!

db/migrate/20161110113139_tidy_up_removed_documents.rb
h-lame: Ah, yes, was leaving it commented out so I could re-run it when I made mistakes or realised I needed to add something.  Good spot!

db/migrate/20161110113139_tidy_up_removed_documents.rb
h-lame: Will do.

h-lame:
Added the final commit with a migration to tidy things up, so removing DNM.

h-lame:
Do want another set of eyes on it from people who have more experience with manuals-publisher though (Hi @steventux!).

app/models/specialist_document.rb
```diff
@@ -108,11 +108,17 @@ def publish!
   end
 
   def withdraw!
-    published_edition.archive if published_edition
+    latest_edition.archive unless withdrawn?
+  end
+
+  def withdraw_and_mark_as_exported_to_live_publishing_api!
```
h-lame: Yeah, maybe just `mark_as_exported` would be simpler.

app/models/specialist_document.rb
```diff
@@ -111,9 +111,9 @@ def withdraw!
     latest_edition.archive unless withdrawn?
   end
 
-  def withdraw_and_mark_as_exported_to_live_publishing_api!
+  def withdraw_and_mark_as_exported!(exported_at = Time.zone.now)
```
steventux: :+1: tidy!

PR #770, opened 2016-11-11
Fix republish script issues

NeilvB:
This PR includes a few fixes for the task of rebuilding publication logs and republishing, following  https://github.com/alphagov/manuals-publisher/pull/767

### DIFF
--- a/lib/manual_publication_log_filter.rb
+++ b/lib/manual_publication_log_filter.rb
@@ -2,7 +2,10 @@ class ManualPublicationLogFilter
   def delete_logs_and_rebuild_for_major_updates_only!(slug)
     PublicationLog.with_slug_prefix(slug).destroy_all
 
-    document_editions_for_rebuild(slug).each do |edition|
+    manual_record = ManualRecord.where(slug: slug).first
+    edition_ordering = EditionOrdering.new(document_editions_for_rebuild(slug), manual_record.latest_edition.document_ids)
+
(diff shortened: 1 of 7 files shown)


lib/manual_publication_log_filter.rb
```diff
@@ -14,6 +17,31 @@ def delete_logs_and_rebuild_for_major_updates_only!(slug)
...
+      editions_not_matching_supplied_documents = @editions.where(:document_id.nin => @document_ids)
+      editions_matching_supplied_documents = @editions.where(:document_id.in => @document_ids)
+
+      order_by_document_ids(editions_matching_supplied_documents).concat(editions_not_matching_supplied_documents.order_by(:created_at, :asc).to_a)
+    end
+
+    private
+
+    def order_by_document_ids(editions)
+      editions.to_a.sort do |a, b|
```
h-lame: Minor nitpick: we can use `.sort_by` here to get a tiny performance hit:

```ruby
editions.to_a.sort_by { |a| @document_ids.index(a.document_id) }
```

With `sort` the provided block is called for every comparison of two elements in the list.  With `sort_by` the provided block is called once for every element in the list, and then the values returned from that block are used to work out the sort order.  If the work you do in the block is expensive then it can be better to run that work once - particularly if the `sort` block has to do the work twice each time it is called, once for each element.

In this case it's probably not much better, as the work isn't expensive, and it's a script we plan to run once, but a useful idiom to know about.

I'll leave it up to you if you want to change it or not.

PR #771, opened 2016-11-15
Move manual editions to own collection

h-lame:
For: https://trello.com/c/gSCJJpNu/59-amend-text-in-manuals-not-working

We move ManualRecord::Editions from being embedded in the ManualRecord documents to living in their own collection.  This avoids manuals with many editions from hitting the 16Mb document limit in Mongo and not accepting any more changes.

There are some more notes in each commit.  The intention is to run this migration during a deploy; it doesn't take very long (5s for the migration, 10s for the whole `rake db:migrate` process on the dev VM), but the old code and new code/schema are incompatible so we need to turn off the publisher for the duration to avoid losing any edits that occur during the migration.

### DIFF
--- a/db/migrate/20161111135556_move_manual_record_editions_to_own_collection.rb
+++ b/db/migrate/20161111135556_move_manual_record_editions_to_own_collection.rb
@@ -1,0 +1,35 @@
+# The history
+# ===========
+#
+# ManualRecord embeds the editions, however a handful of the manuals have
+# so many editions that they are starting to butt up against the 16Mb document
+# size limit in our versio of mongo.  This means users can't make changes and
+# get strange behaviour from the system.
+#
+# To get around this and keep the API as simple as possible we turn the
+# `embeds_many` association on ManualRecord into a `has_many` association
+# and move all the existing editions out of ManualRecord's collection and
+# into their own collection.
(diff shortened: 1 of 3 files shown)


h-lame:
Flagged as DNM because it sits on top of #769 and we'll want to rebase this away once that's merged.  If you're reviewing, look at the last 2 commits - they're the ones for this PR.

h-lame:
Removed DNM because #769 is merged now and we've rebased.  Will probably hold off on merging until #769 is deployed to prod though.

db/migrate/20161111135556_move_manual_record_editions_to_own_collection.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+
+    manual_records_collection.update({}, {:'$unset' => {editions: 1}}, multi: true)
+  end
+
+  def self.down
+    # Whilst it would be possible to reverse this, it would be a lot of work
+    # for something that is unlikely to ever get run.
+    # raise IrreversibleMigration
```
steventux: Assume this should raise given the comment?

steventux:
:star: :star: :star:  _unlocks Manual of Patent Practice_ :star: :star: :star:

PR #773, opened 2016-11-24
Tighten up PublicationLog#change_notes_for

h-lame:
This method would return all publication logs that had the supplied slug
as a prefix to allow us to get the publication logs for a manual section
at the same time as the publication logs for the manual.  However this
didn't account for manuals whose slugs might be a prefix of another
manual.  For example the manual slug 'guidance/test' is a prefix of the
manual slug 'guidance/test-manual'.  This means when asking for the logs
for 'guidance/test' you would get the logs for both manuals (and all their
sections) which is almost certainly not what you wanted.

We tighten up the regexp used to extract the logs so that the slug is
followed by a / (so we get logs for child sections) or the end of the
string (so we get logs for the manual).

This has happened as there are a handful of manuals whose slugs are a
prefix of another manual, but in practice they're all either withdrawn
or unpublished, so there are no logs.  But it's likely to become a bigger
problem if we don't fix it.

### DIFF
--- a/app/models/publication_log.rb
+++ b/app/models/publication_log.rb
@@ -12,7 +12,7 @@ class PublicationLog
 
   alias_attribute :published_at, :created_at
 
-  scope :with_slug_prefix, ->(slug) { where(slug: /^#{slug}.*/) }
+  scope :with_slug_prefix, ->(slug) { where(slug: /^#{slug}(\/|$)/) }
 
   def self.change_notes_for(slug)
     with_slug_prefix(slug)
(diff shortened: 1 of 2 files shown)


app/models/publication_log.rb
```diff
@@ -12,7 +12,7 @@ class PublicationLog
 
   alias_attribute :published_at, :created_at
 
-  scope :with_slug_prefix, ->(slug) { where(slug: /^#{slug}.*/) }
+  scope :with_slug_prefix, ->(slug) { where(slug: /^#{slug}(\/|$)/) }
```
NeilvB: We don't seem to have change notes on manual editions at the moment so I'm not sure if we strictly need the alternative match on the end of the slug string (`$`). But it makes sense to have it as these could easily be added later.

NeilvB:
Looks good - well spotted 👍

app/models/publication_log.rb
h-lame: Yeah, we don't (but it seems odd that we don't).

PR #774, opened 2016-11-24
Fix url generation to remove extra query string params

h-lame:
Urls in the system were being generated with path params duplicated
as query params as follows:

    /manuals/<manual_id>?id=<manual_id>
    /manuals/<manual_id>/sections/<section_id>?id=<section_id>&manual_id=<manual_id>

After some digging we discovered that this was because the objects we were
using for url generation were not the ManualRecord or
SpecialistDocumentEdition mongoid documents, or even the Manual and
SpecialistDocument model objects.  Instead they were deeply wrapped up in
SimpleDelegator classes like NullValidator or ManualWithDocuments.

The rails routing algorithm uses the provided arguments to fill in the
path segments in the url, but does equality to decide if the argument
has been consumed and should be removed for consideration later when
building query string params.  The SimpleDelegator objects mean that
the manual and section instances were not considered equal and so were
not removed from the list of arguments used already in building the
url and so ended up in the query string as well as the path.  We can
see the SimpleDelegator and eql? behaviour below:

    class A; end
    a1 = A.new
    a2 = A.new
    a1.eql? a2 #=> false
    a1.eql? a1 #=> true

    class B < SimpleDelegator; end
    ba1 = B.new(a1)
    ba2 = B.new(a2)
    ba1.eql? ba2 #=> false
    ba1.eql? ba1 #=> false
    ba1.eql? a1 #=> true

The delegate thinks it is equal to the delegated object, but does not
think it is equal to another delegate - even when that other delegate is
itself.  This is because the delegate delegates eql? to the wrapped object
and this does object equality, so really `ba1.eql? ba1` is doing
`a1.eql? ba1`.  If instead we define our own eql? method on the wrapped
object we get the following behaviour:

    class C
      attr_reader :id
      def initialize(id)
        @id = id
      end
      def eql?(other)
        @id.eql? other.id
      end
    end

    c1 = C.new(1)
    c1_2 = C.new(1)
    c2 = C.new(2)
    c1.eql? c1 #=> true
    c1.eql? c1_2 #=> true
    c2.eql? c1 #=> false

    bc1 = B.new(c1)
    bc1_2 = B.new(c1_2)

    bc1.eql? bc1 #=> true
    bc1.eql? c1 #=> true
    bc1_2.eql? c1 #=> true
    bc1.eql? c1_2 #=> true
    bc1.eql? bc1_2 #=> true

So this is what we do, we add eql? implementations to Manual and
SpecialistDocumentEdition to make sure they are considered equal,
not matter how deeply wrapepd in simple delegators they are.

There's a closed (but not fixed) issue report on journey (the rails
routing algorithms) that explains this situation here:

https://github.com/rails/journey/issues/29

### DIFF
--- a/app/models/manual.rb
+++ b/app/models/manual.rb
@@ -23,6 +23,10 @@ class Manual
     id
   end
 
+  def eql?(other)
+    id.eql? other.id
+  end
+
   def attributes
     {
       id: id,
--- a/app/models/specialist_document.rb
+++ b/app/models/specialist_document.rb
@@ -155,6 +155,10 @@ class SpecialistDocument
     editions.any?(&:persisted?)
   end
 
+  def eql?(other)
+    id.eql?(other.id)
+  end
+
 protected
 
   attr_reader :slug_generator, :edition_factory
--- a/spec/features/manual_urls_spec.rb
+++ b/spec/features/manual_urls_spec.rb
@@ -1,0 +1,25 @@
+require "spec_helper"
+
+RSpec.describe "manual urls", type: :feature do
+  before do
+    login_as(:gds_editor)
+  end
+
+  let!(:manual) { create_manual_without_ui(title: "A manual", summary: "A manual summary", body: "A manual body") }
+  let!(:section) { create_manual_document_without_ui(manual, { title: "Section 1", summary: "A section summary", body: "A section body" }) }
+
+  it "should respond with 'OK'" do
+    visit "/manuals"
+
+    expect(page).to have_link("A manual", href: %r{/manuals/#{manual.id}$})
+
+    click_on "A manual"
+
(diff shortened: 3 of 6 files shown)


jennyd:
OMG this has been something we've wanted to fix for ages (and have tried and failed in the past) ✨ 🍰 💯 🇬🇶 💕

h-lame:
Decided that as the system doesn't make it easy to obtain a specific version of a manual or specialist document it doesn't make sense to declare the objects different if the version number is different.  Particularly as what has prompted us to care about for equality here is url generation, and version 1 of a manual and version 2 of a manual would currently have the same url, so the system should treat them as equal.  For now.

NeilvB:
Looks good!

PR #778, opened 2016-12-02
Publication log rebuild includes first manual edition timestamp

NeilvB:
Trello card: https://trello.com/c/x6vSw5LM/55-update-history

Best reviewed commit-by-commit.

We don't seem to be able to always rely on document editions to faithfully reproduce the timestamps of publication logs that are currently seen in production.

The main commit in this PR involves a special case for those document editions that are associated with a first manual edition. The publication logs for these cases have a timestamp that is set to the manual `updated_at` time rather than the document edition timestamp. This appears to improve the accuracy of the rebuilt publication log timestamps in relation to what's currently seen on production.

### DIFF
--- a/bin/rebuild_major_publication_logs_for_manuals
+++ b/bin/rebuild_major_publication_logs_for_manuals
@@ -7,16 +7,15 @@ require "logger"
 logger = Logger.new(STDOUT)
 logger.formatter = Logger::Formatter.new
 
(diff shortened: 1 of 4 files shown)


h-lame:
This makes sense from a code point of view.  We're yet to get confirmation from interested publishers that the approach covers all the edgecases but it's an improvement over what we already have so we're merging so that we don't have to keep rebasing.

PR #779, opened 2016-12-09
Make ManualRelocator be more robust with what it considers published

h-lame:
If the latest edition is published, that's acceptable not just when it's
the only edition.

We deployed #765 to integration and the migration failed because the HCA version of the manual was currently published, but had more than one edition and it failed.

### DIFF
--- a/lib/manual_relocator.rb
+++ b/lib/manual_relocator.rb
@@ -49,8 +49,8 @@ private
 
   def manual_is_currently_published?(manual)
     # to be currently published either...
-    # 1. it's got one edition that is published
-    (manual.editions.count == 1 && manual.latest_edition.state == "published") ||
+    # 1. the latest edition is published
+    (manual.latest_edition.state == "published") ||
     # or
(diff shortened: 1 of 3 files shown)


h-lame:
We shouldn't need this 2nd commit, but as we've run the migration partly and failed later from a different branch, the "bad" manual is already gone, and so the migration when run a second time will fail.

h-lame:
Deployed this to integration with success and checked that things have been redirected and reslugged as expected.

lib/manual_relocator.rb
```diff
@@ -49,8 +49,8 @@ def validate_manuals
 
   def manual_is_currently_published?(manual)
     # to be currently published either...
-    # 1. it's got one edition that is published
-    (manual.editions.count == 1 && manual.latest_edition.state == "published") ||
+    # 1. the latest edition is published
+    (manual.latest_edition.state == "published") ||
```
steventux: :+1: I saw this problem attempting to run this

PR #780, opened 2016-12-14
Let publishing api deal with timestamps

h-lame:
Publishing API can set timestamps correctly based on the api events we send it for the content and the update_type params used.  This PR attempts to fix: https://trello.com/c/OC3RskhE/63-timestamps-on-manuals by making sure our update_types are set correctly and not sending public_updated_at to the API anymore.  More notes in the commits.

### DIFF
--- a/app/exporters/manual_publishing_api_exporter.rb
+++ b/app/exporters/manual_publishing_api_exporter.rb
@@ -41,7 +41,6 @@ private
       document_type: "manual",
       title: rendered_manual_attributes.fetch(:title),
       description: rendered_manual_attributes.fetch(:summary),
-      public_updated_at: rendered_manual_attributes.fetch(:updated_at).iso8601,
       update_type: update_type,
       publishing_app: "manuals-publisher",
       rendering_app: "manuals-frontend",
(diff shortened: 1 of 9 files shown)


kevindew:
@h-lame Code looks sane 👍 though `public_updated_at` [is required](https://github.com/alphagov/publishing-api/blob/f4c0c3d9044033ad9a680e644b4cee389c827521/doc/model.md#public_updated_at) so I'd have thought you'd hit issues submitting this?

h-lame:
@kevindew - interesting.  I've ran this code locally and publishing api didn't seem to worry about it.  I'll check again though.  Given I want to let publishing-api deal with it entirely is there anything else I could do (for example would submitting blank values work?).

kevindew:
Hmm yeah looks like our documentation may not be accurate in that regard (and my searching skills ;) ) as I've now found [the exact thing you want/are using](https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/publish.rb#L115) - Sorry about that, will update our docs.

h-lame:
Oh, cool.  The docs aren't totally inaccurate, as it was reading the docs in the first place that led me to think I could do this in the first place.

h-lame:
OK, so this works, but it breaks previewing drafts because the content-store version no longer has a `public_updated_at` value.  Will have to work out if we should fix manuals-frontend, or try another approach.  A published document will have a public_updated_at, but it seems our drafts don't (which almost makes sense).

PR #781, opened 2016-12-14
Remove email-alert cruft

tijmenb:
Since the specialist - manuals publisher schism this app hasn't used email-alert-api directly. This removes some cruft around email alerts in tests, so that I don't get confused about this anymore.


h-lame:
Only deletions and no additions you say?  ➕ 💯

PR #782, opened 2016-12-20
Always send first editions to publishing api as major updates

h-lame:
When republishing manuals for #778 after deploying #780 some of the republishes failed with a 500 error from publishing-api (see: https://errbit.integration.publishing.service.gov.uk/apps/552637940da1157bfa000399/problems/585908a76578636a1ecc0000).  Turns out publishing-api doesn't like it if the first edition of a content item has update_type of "minor".  This is on the publishing-api backlog to turn into a proper error (see: https://trello.com/c/3S4m5wAJ/251-throw-an-error-if-the-first-publish-does-not-use-a-major-update-type) but we need to handle it in manuals publisher.  Turns out that's painful because manuals-publisher doesn't really keep things in lock-step with what is in publisher.

There are more notes in the commits if you're interested.

### DIFF
--- a/app/exporters/manual_publishing_api_exporter.rb
+++ b/app/exporters/manual_publishing_api_exporter.rb
@@ -60,7 +60,23 @@ private
   end
 
   def update_type
-    manual.documents.select(&:needs_exporting?).all?(&:minor_update?) ? "minor" : "major"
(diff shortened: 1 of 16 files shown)


h-lame:
Thanks @cbaines!

PR #783, opened 2016-12-21
Fix flappy test

h-lame:
To fix the flappy test detected when we merged #782 to master (see: https://ci.dev.publishing.service.gov.uk/job/govuk_manuals_publisher/722/)


h-lame:
This didn't quite get to the bottom of it as I got a local failure about dates in manual_publication_log_filter_spec.  I think we need to be even more explicit about the dates of the editions in this spec to make sure stuff will be in the right order.  /cc @NeilvB

PR #787, opened 2017-01-04
Publish change notes along with relevant manual edition

NeilvB:
[Trello card](https://trello.com/c/WpgrE9ge/71-change-notes-in-manuals-publisher-are-sent-to-the-publishing-api-on-the-wrong-publish-event)

Adds an additional manual draft export step at the point of
publishing a manual. This was done in order to address the problem of
a change note only being sent to the publishing API following
a second, subsequent creation of a manual draft and its publication. In
other words, the change notes always lagged "one step behind" the
publication of the content changes that they described.

By adding a draft export step to the publication of a manual we ensure
that the change notes that were added to the current draft are then sent
through to the publishing API.

Also:
  - Modified some publishing specs to take account of this additional
  draft export step.

### DIFF
--- a/app/observers/manual_observers_registry.rb
+++ b/app/observers/manual_observers_registry.rb
@@ -9,8 +9,13 @@ class ManualObserversRegistry
   def publication
     # The order here is important. For example content exporting
     # should happen before publishing to search.
+    #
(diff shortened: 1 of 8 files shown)


h-lame:
Looks good, seems like the simplest solution too.

PR #792, opened 2017-01-24
Rake task for republishing manuals

alecgibson:
This Rake task will be required to republish all manuals, so that they are re-indexed in Rummager with a populated `content_store_document_type` field.

Trello: https://trello.com/c/M1cXENeB/341-add-content-store-document-type-to-all-guidance-content-in-rummager?menu=filter&filter=@alec_gibson

### DIFF
--- a/lib/tasks/manuals.rake
+++ b/lib/tasks/manuals.rake
@@ -1,0 +1,11 @@
+namespace :manuals do
+  desc "Republishes all published manuals"
+  task republish: :environment do
+    registry = ManualServiceRegistry.new
+
+    registry.list({}).call.select { |manual| manual.published? }.each do |manual|
+      puts "Republishing #{manual.id} #{manual.title}"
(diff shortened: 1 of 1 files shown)


h-lame:
There's already a bin script for doing this: https://github.com/alphagov/manuals-publisher/blob/64b6619c9b40aa1cd0f9f37d3c9592b599df08dc/bin/republish_manuals so we should probably just use that to avoid confusion.

For some reason this app uses bin scripts not rake tasks and someone explicitly made that choice as they recently consolidated to make sure we only had bin scripts.

Do note that the republish action only republishes manuals without a new draft - if you need to republish things that are already published and have a new draft waiting, you might need to do something else.  See https://trello.com/c/w8XMP7h2/73-investigate-potential-bug-with-republishing

alecgibson:
@h-lame okay - I missed that. Will close this PR.

PR #793, opened 2017-01-25
Allow setting a previously published date

h-lame:
For: https://trello.com/c/hvHGxTuD/94-change-published-date-in-manuals

This PR adds functionality inspired by specialist-publisher and whitehall.

1. Editing change notes is even more available than previously (you can edit them even on the first edition, but you can't make it a minor change).
2. The change note UI is clearer, using radiobuttons instead of a checkbox to indicate minor change
3. There is a new UI element when creating a manual for setting an originally published date, like the similar UI in whitehall
4. On subsequent edits, the originally published date is editable via a new UI action (mostly because the results of changing this require us to do more work than a normal update and we can't easily detect this change otherwise).

### creating a manual UI with new date entry
manuals-publisher-create

### editing a manual's original publication date
manuals-publisher-edit-date

### adding a new section with change note UI
manuals-publisher-new-section-change-note

### editing a published section with change note UI
manuals-publisher-editing-section-change-note


h-lame:
This is a big one, probably best reviewed commit-by-commit as there are notes that try to explain things as we go.

app/models/specialist_document.rb
```diff
@@ -86,7 +86,11 @@ def update(params)
   def published?
     editions.any?(&:published?)
   end
-  alias_method :has_ever_been_published?, :published?
+
+  def has_ever_been_published?
+    return false if @editions.size == 1 && needs_exporting?
+    published?
```
cbaines: I'm definitely missing context, but looking at this, I'm unsure why its mixing in exporting and publishing? I'm guessing that publishing can be interpreted in the context of the Publishing API, but I'm less sure what exporting is?

h-lame: When you create a new draft it is marked as needing exporting (it's exported_at date hasn't been set).  When you publish that document the first thing that happens is the state is changed from draft to published, then we write out a final draft to the publishing-api; at which point if we ask if the document has ever been published to work out some logic about how to represent the document for exporting to publishing-api.  Without this change it'll say it has, when in fact it's in the middle of being published and so we were seeing the wrong content in drafts being sent to the publishing-api.  It's messy as all heck, yes, but that's par for the course for this app.

h-lame: To close the loop, once we write the final draft, then we send the publish message to publishing-api and then we set the exported_at timestamp so only at the end of the publishing process will the document say it is both `published?` and no longer `needs_exporting?`

h-lame: And, this particular problem is really only a problem for the first edition of a document.  In the past the UI would let you create a minor update for the first edition but we have to tell the publishing-api that it's a major update the first time it's published - regardless of what the model thinks.  The UI no longer allows this, but historical data still persists and it can break republish events if we don't cope with it.

cbaines: Ok. I don't have any concrete suggestions to offer, so I'm happy to go ahead with this.

PR #794, opened 2017-01-30
Send the correct public_updated_at to publishing-api for cbt manual

h-lame:
For: https://trello.com/c/StafARqt/98-manual-has-a-publication-timestamp-that-is-wrong

This manual somehow has a `first_published_at` that is after it's current
`public_updated_at` in the publishing-api/content-store.  We don't know
how it happened, but we can fix it with a republish event that sets the
`public_updated_at` correctly.

There are more notes in the migration.


NeilvB:
Looks good 👍

PR #795, opened 2017-01-31
Add a note to README about supported versions of bundler

floehopper:
The `cucumber` rake task which is run as part of the `default` rake task does not work with `bundler` versions of 1.13.0 onwards. The following exception occurs: `cannot load such file -- active_model/translation (LoadError)`. The full stack trace is recorded [here](https://gist.github.com/floehopper/79341ba0205a7d95fe0cd8ca369f8551).

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ If you are using the GDS development virtual machine then the application will b
 $ bundle exec rake
 ```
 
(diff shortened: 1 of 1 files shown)


chrisroos:
Looks good to me, @floehopper.

PR #796, opened 2017-01-31
Add note to README about supported versions of MongoDB

floehopper:
The insertion of a user (and the app in general) doesn't work with recent versions of MongoDB. v3.0.12 works OK; v3.4.1 does NOT work due to a problem with the `:w => 1` option no longer being supported at the top level, i.e. outside the Write Concern options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ In a rails console do:
 User.create!(name: "My Name", email: "my.email@somedomain.com", permissions: ["gds_editor"], organisation_slug: "government-digital-service", organisation_content_id: "af07d5a5-df63-4ddc-9383-6a666845ebe9")
 ```
 
-Note: This insert (and the app in general) doesn't work with recent versions of MongoDB. v3.0.12 works OK; v3.4.1 does NOT work due to a problem with the `:w => 1` option no longer being supported at the top level, i.e. outside the Write Concern options.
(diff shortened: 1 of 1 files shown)


h-lame:
It might be better to explicitly mention the current mongo (and bundler) versions that are pinned by govuk-puppet for this app / ruby version.

chrisroos:
Do you know how we can find the mongo version, @h-lame?

My best guess is that it's 2.6 based on [this comment in mongodb/manifests/server.pp](https://github.com/alphagov/govuk-puppet/blob/master/modules/mongodb/manifests/server.pp#L8). Although I can see that the [mongo version is explicitly set to 2.4.9 or 3.2.7 in a couple of places](https://github.com/alphagov/govuk-puppet/search?utf8=%E2%9C%93&q=%22mongodb%3A%3Aserver%3A%3Aversion%22&type=Code) but it's not clear to me whether these are being used on the server hosting manuals-publisher.

h-lame:
I think it's 2.4.9 because of https://github.com/alphagov/govuk-puppet/blob/f3614e33bcf037b218e0b9e816f0994786b41efb/hieradata/common.yaml#L1256 - the other things that set this variable are for ci machines or licensing machines, neither of which are relevant for this app (I'm not entirely sure how to point to a definition in puppet that would confirm that part though).

floehopper:
@h-lame:

> It might be better to explicitly mention the current mongo (and bundler) versions that are pinned by govuk-puppet for this app / ruby version.

I've added a sentence based on your reply to the question from @chrisroos. Does that address your concern? And would you prefer me to squash this commit into the first one?

h-lame:
Looks good now, thanks.

PR #798, opened 2017-02-01
Spike refactor the app

chrisroos:
I'm using this spike to explore whether it's worth trying to simplify the app.

In this branch I:

* Remove some unused code that I came across
* Introduce some class methods to replace the `ManualsPublisherWiring`
* Remove `ManualsPublisherWiring` and `DependencyContainer`

I'm pushing up this work in progress branch to get some feedback about the approach.


app/lib/publishing_api.rb
```diff
@@ -0,0 +1,9 @@
+class PublishingApi
+  def self.instance
+    GdsApi::PublishingApi.new(
```
floehopper: I think it would be better to be consistent about memo-izing the singleton instance as you did for `RummagerApi` in [this commit][1], i.e. as per the implementation of [`SingletonDependency#get`][2].

[1]: 0aabc2e51d8135b88aad91751bfa7f991f00e3fd
[2]: https://github.com/alphagov/manuals-publisher/blob/0500169731a70ff35b615e1a599637f86d991e11/lib/dependency_container.rb#L82

app/lib/publishing_api.rb
```diff
@@ -1,3 +1,5 @@
+require "gds_api/publishing_api"
+
```
floehopper: I assume you intend to squash this into the previous commit...?

app/lib/publishing_api_v2.rb
```diff
@@ -0,0 +1,10 @@
+require "gds_api/publishing_api_v2"
+
+class PublishingApiV2
+  def self.instance
+    GdsApi::PublishingApiV2.new(
```
floehopper: [Same comment as before](https://github.com/alphagov/manuals-publisher/pull/798/commits/fcb49fcebc6dbd8c90af89bc2945273780f0e600#r98912390).

app/lib/organisations_api.rb
```diff
@@ -0,0 +1,7 @@
+require "gds_api/organisations"
+
+class OrganisationsApi
+  def self.instance
+    GdsApi::Organisations.new(ORGANISATIONS_API_BASE_PATH)
```
floehopper: [Same comment as before](https://github.com/alphagov/manuals-publisher/pull/798/commits/fcb49fcebc6dbd8c90af89bc2945273780f0e600#r98912390).

app/lib/rummager_api.rb
```diff
@@ -0,0 +1,7 @@
+require "gds_api/rummager"
+
+class RummagerApi
+  def self.instance
+    @rummager_api ||= GdsApi::Rummager.new(Plek.new.find("search"))
+  end
+end
```
h-lame: A pattern that other gov.uk apps have been using is a Services module that has class methods on it for each gds-api used by the app.  Might be worth pushing this as far as that too.  For example: https://github.com/alphagov/policy-publisher/blob/0f2fa349a043ae7acea37e22bbb08b57154fb30e/lib/services.rb

app/lib/publishing_api_v2.rb
h-lame: I'd be 100% surprised if we need `publishing_api` and `publishing_api_v2`.  I suspect we can drop the former, and rename the latter to `publishing_api`

app/lib/specialist_document_renderer.rb
```diff
@@ -0,0 +1,19 @@
...
+require "govspeak_to_html_renderer"
+
+class SpecialistDocumentRenderer
+  def self.create
+    ->(doc) {
+      pipeline = [
+        MarkdownAttachmentProcessor.method(:new),
+        SpecialistDocumentHeaderExtractor.create,
+        GovspeakToHTMLRenderer.create,
+      ]
```
h-lame: It feels like it might be a bug that this and the `ManualDocumentRenderer` you deal with in the next few commits are subtly different; this one doesn't do `FootnoteSomethingSomething` in its pipeline, but the `ManualDocumentRenderer` does.  We should check obviously, but I think we can get rid of this one and replace it's single use with an instance of the `ManualDocumentRenderer` instead.

h-lame:
I think what you've done so far looks good.  My comments are mostly about where you could push it further next if you plan to keep iterating on this spike.  If you think you're done for now though I don't think it would hurt to merge (maybe wait until #799 is merged and deployed though).

chrisroos:
Thanks for looking at this @floehopper and @h-lame.

Do you think it's worth the effort of tidying this branch up and merging it? Does the approach in general leave the code in an improved-enough state to warrant more work?

chrisroos:
Note that some of the early commits in this branch (those that remove code) have been moved to PR #801.

chrisroos:
I've rebased this on master and force pushed now that a number of the other PRs have been merged.

chrisroos:
I've rebased this branch on master, addressed some of the concerns raised in the comments and opened PR #808 which supersedes this PR.

app/lib/publishing_api.rb
floehopper: Addressed in ece3882d4b7384a8c084f05565c6b1933594a450.

app/lib/publishing_api.rb
floehopper: Addressed in ece3882d4b7384a8c084f05565c6b1933594a450.

app/lib/publishing_api_v2.rb
floehopper: > Same comment as before.

Addressed in 8592ac5ccbcc8b07f8e08240a132ee9a8f60835c.

floehopper: > I'd be 100% surprised if we need `publishing_api` and `publishing_api_v2`. I suspect we can drop the former, and rename the latter to `publishing_api`

Addressed by opening #809.

app/lib/rummager_api.rb
floehopper: Addressed by opening #811.

app/lib/specialist_document_renderer.rb
floehopper: Addresses by opening #810.

PR #799, opened 2017-02-01
Simplify ui around originally published at

h-lame:
For: https://trello.com/c/hvHGxTuD/94-change-published-date-in-manuals

As a follow up to the UI in #793 we tweak the UI slightly to simplify it based on feedback from folk better acquainted with the app and its users.

Highlights:

* the state lozenges now include "published with new draft" to distinguish draft manuals that have never been published from those that have been published in a previous edition

manuals publisher

* the publish button now lives in a panel that includes text describing the outcome (major vs. minor, how the updated at timestamp will change) or explaining why the publish button isn't present

manuals publisher-1

* the create UI for setting the first published at remains the same, but the edit UI no longer has the first choice ("this manual is new" vs "this manual was published elsewhere") and the second choice has been reworded.

manuals publisher-2

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,13 +2,6 @@
 
 module ApplicationHelper
   def state(document)
-    state, classes = state_for_frontend(document)
-
-    content_tag(:span, state, class: classes).html_safe
-
-  end
-
-  def state_for_frontend(document)
     state = document.publication_state
 
     if %w(published withdrawn).include?(state) && document.draft?
@@ -20,10 +13,9 @@ module ApplicationHelper
     else
       classes = "label label-default"
     end
-    [state, classes]
-  end
 
-  module_function :state_for_frontend
+    content_tag(:span, state, class: classes).html_safe
+  end
 
   def show_preview?(item)
     if item.respond_to?(:documents)
@@ -96,10 +88,6 @@ module ApplicationHelper
 
   def content_preview_url(document)
     "#{Plek.current.find("draft-origin")}/#{document.slug}"
-  end
-
-  def finders_sorted_by_title
-    finders.sort_by {|_, value| value[:title] }
   end
 
   def publish_text(manual, slug_unique)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -102,4 +102,24 @@ module ApplicationHelper
     finders.sort_by {|_, value| value[:title] }
   end
 
+  def publish_text(manual, slug_unique)
+    if manual.state == "published"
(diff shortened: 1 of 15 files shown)


features/step_definitions/manual_steps.rb
```diff
@@ -716,3 +717,15 @@
     extra_attributes: manual_table_of_contents_attributes,
   )
 end
+
+Then(/^the manual is listed as (draft|published|published with new draft)$/) do |status|
+  visit manuals_path
+
+  # binding.pry if status == 'published with new draft'
```
NeilvB: ✂️  🙂

spec/lib/manual_update_type_spec.rb
```diff
@@ -0,0 +1,86 @@
+require "spec_helper"
+
+RSpec.describe ManualUpdateType do
```
NeilvB: Great, like how this refactor gives us more unit test coverage

app/helpers/application_helper.rb
```diff
@@ -102,4 +102,24 @@ def finders_sorted_by_title
     finders.sort_by {|_, value| value[:title] }
   end
 
+  def publish_text(manual, slug_unique)
+    if manual.state == "published"
+      text = "<p>There are no changes to publish.</p>"
+    elsif manual.state == "withdrawn"
+      text = "<p>The document is withdrawn. You need to create a new draft before it can be published.</p>"
```
NeilvB: Might it be better to refer to 'manual' rather than 'document' here?

app/helpers/application_helper.rb
```diff
@@ -102,4 +102,24 @@ def finders_sorted_by_title
...
+    elsif !current_user_can_publish?("manual")
+      text = "<p>You don't have permission to publish this document.</p>"
```
NeilvB: And here?

features/step_definitions/manual_steps.rb
h-lame: oops!

app/helpers/application_helper.rb
h-lame: Oh, yes.  I copied this almost directly from specialist-publisher but not with a critical enough eye.

features/step_definitions/manual_steps.rb
h-lame: done.

app/helpers/application_helper.rb
h-lame: done.

app/helpers/application_helper.rb
h-lame: done.

PR #800, opened 2017-02-01
Use Ruby keyword args instead of custom code

chrisroos:
A number of `initialize` methods were taking a hash parameter and then manually extracting required/optional parameters. I think it's clearer to use Ruby's keyword arguments to serve this same purpose.

This branch changes all but one instance (`Manual#initialize`) to use Ruby keyword args instead of custom code. I can't immediately see how to change `Manual#initialize` without also having to change the code that calls it.

We already had a mixture of regular args and keyword args in the code but making them more consistent is outside the scope of this PR.

### DIFF
--- a/app/lib/slug_generator.rb
+++ b/app/lib/slug_generator.rb
@@ -1,6 +1,6 @@
 class SlugGenerator
-  def initialize(dependencies)
-    @prefix = dependencies.fetch(:prefix)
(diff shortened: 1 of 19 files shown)


chrisroos:
Thanks @h-lame. I've rebased and force pushed in preparation for merging.

PR #802, opened 2017-02-01
Remove unused Cucumber steps

chrisroos:
I used `cucumber --format usage` to find and remove these unused steps.


chrisroos:
Thanks @h-lame!

PR #803, opened 2017-02-07
Remove redundant code related to SpecialistDocumentEdition#extra_fields

floehopper:
I came across this while trying to work out whether the `SpecialistDocumentEdition#extra_fields` field was being used. Although this PR is not as complete as I would like, I thought it was worth capturing my investigations so far. Feel free to close this PR if you think it needs to be more complete before it gets merged.

### Remove redundant DocumentMetadataDecorator

As far as I can see all this class' sub-classes were removed in the giant ["Remove specialist document models" commit][1]. I can't see any usage of this class or any of its methods and all the tests still pass.

### Remove usages of SpecialistDocumentEdition#extra_fields

It looks as if the means to edit this field via the UI was removed in the giant ["Remove specialist document models" commit][1]. As far as I can tell this field was only ever needed for documents which are now published via the Specialist Publisher app. I believe we could also remove the field itself. However, this is complicated by:

a) There are some instances of `SpecialistDocumentEdition` in the local data store which have the `extra_fields` attribute set. At first glance it looks as if these might all be orphans and in any case relate to content items which are now published via the Specialist Publisher app.

b) There are some existing migrations which rely on this field.

I have not removed the field nor addressed those issues in this pull request.

[1]: https://github.com/alphagov/manuals-publisher/commit/3162f0e72fd53bfd1cf87f5661bb198ff785940e


h-lame:
Needs a rebase but then we can merge.

PR #804, opened 2017-02-08
Rework republishing

h-lame:
For: https://trello.com/c/w8XMP7h2/73-investigate-potential-bug-with-republishing

We worried that republishing in manuals-publisher wasn't doing the right thing and might end up publishing drafts. Turns out it didn't, but neither did it do a full republish as it only worked on manuals that had been published and didn't have a new draft waiting.  If the intention of republishing is to resend the published version of a manual to the publishing-api we were missing out.  Looking at what some other apps did we realised what we should be doing is:

1. for draft manuals - resend the draft to the publishing api
2. for published manuals - write a new draft of the manual to the publishing api and then publish it
3. for published manuals with a new draft - write a draft of the published version of the manual to the publishing api and then publish it, finally resend the new draft to the publishing api.

If we don't do the last bit then the new draft is lost from the publishing-api and "preview draft" functionality will show the previously published version.

Turns out that extracting the previously published version of a manual is quite hard so we've had to add a new repository for doing this, and change the republishing service to take two sets of listeners because what we want to do with a published manual we are republishing is different to what we want to do with a draft manual we are "republishing".

I haven't tried a full republish of all the manuals locally, but I have tried a single republish of a published with a new draft manual and it all worked properly.  Which is nice. 

Probably best reviewed commit-by-commit.

### DIFF
--- a/app/exporters/manual_publishing_api_exporter.rb
+++ b/app/exporters/manual_publishing_api_exporter.rb
@@ -1,5 +1,5 @@
 class ManualPublishingAPIExporter
-  UPDATE_TYPES = %w(minor major republish).freeze
+  include PublishingAPIUpdateTypes
 
   def initialize(export_recipent, organisation, manual_renderer, publication_logs, manual, update_type: nil)
     @export_recipent = export_recipent
@@ -7,8 +7,8 @@ class ManualPublishingAPIExporter
     @manual_renderer = manual_renderer
     @publication_logs = publication_logs
     @manual = manual
-    raise ArgumentError, "update_type '#{update_type}' not recognised" if update_type.present? && !UPDATE_TYPES.include?(update_type)
     @update_type = update_type
+    check_update_type!(@update_type)
   end
 
   def call
@@ -22,7 +22,7 @@ private
     :organisation,
     :manual_renderer,
     :publication_logs,
-    :manual
+    :manual,
   )
 
   def base_path
--- a/app/exporters/manual_publishing_api_exporter.rb
+++ b/app/exporters/manual_publishing_api_exporter.rb
@@ -1,11 +1,14 @@
 class ManualPublishingAPIExporter
+  UPDATE_TYPES = %w(minor major republish).freeze
 
-  def initialize(export_recipent, organisation, manual_renderer, publication_logs, manual)
+  def initialize(export_recipent, organisation, manual_renderer, publication_logs, manual, update_type: nil)
     @export_recipent = export_recipent
     @organisation = organisation
     @manual_renderer = manual_renderer
     @publication_logs = publication_logs
     @manual = manual
+    raise ArgumentError, "update_type '#{update_type}' not recognised" if update_type.present? && !UPDATE_TYPES.include?(update_type)
+    @update_type = update_type
   end
 
   def call
@@ -69,6 +72,7 @@ private
   end
 
   def update_type
+    return @update_type if @update_type.present?
     ManualUpdateType.for(manual)
   end
 
--- a/app/repositories/versioned_manual_repository.rb
+++ b/app/repositories/versioned_manual_repository.rb
@@ -1,0 +1,111 @@
+require "manual_repository"
+
+class VersionedManualRepository
+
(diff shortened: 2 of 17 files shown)


app/exporters/manual_publishing_api_exporter.rb
```diff
@@ -1,11 +1,14 @@
 class ManualPublishingAPIExporter
+  UPDATE_TYPES = %w(minor major republish).freeze
```
NeilvB: We're using the same set of update types here in the exporter and as well in the section publisher and api publisher. Might it be worthwhile having them defined in just one place? Or are we keeping them separate because we might be expecting variation between publishing/exporting services?

spec/exporters/publishing_api_publisher_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+
+      it "asks the publishing api to publish the document" do
+        subject.call
+
+        expect(publishing_api).to have_received(:publish).with(document_id, nil)
+      end
+    end
+
+    context "when no explicit update_type is given" do
```
NeilvB: Looks like a typo - could go for "when an explicit update_type is given" ?

app/repositories/versioned_manual_repository.rb
```diff
@@ -0,0 +1,111 @@
...
+  def get_manual(manual_id)
+    manual_record = ManualRecord.find_by(manual_id: manual_id)
+    raise NotFoundError if manual_record.nil?
+
+    {
+      draft: get_current_draft_version_of_manual(manual_record),
+      published: get_current_published_version_of_manual(manual_record),
+    }
+  end
+
```
NeilvB: It looks like we can keep the public methods of this class to just the methods above this line.

app/repositories/versioned_manual_repository.rb
```diff
@@ -0,0 +1,111 @@
...
+  def get_current_draft_version_of_manual(manual_record)
+    return nil if manual_record.nil?
```
NeilvB: If this method isn't intended to be exposed anywhere else, I think we can remove this line as we already do checking for an nil manual in `get_manual`

app/repositories/versioned_manual_repository.rb
```diff
@@ -0,0 +1,111 @@
...
+    build_manual_for(manual_record, manual_record.latest_edition) do
+      {
+        documents: get_latest_version_of_documents(manual_record.latest_edition.document_ids),
+        removed_documents: get_latest_version_of_documents(manual_record.latest_edition.removed_document_ids),
+      }
+    end
+  end
+
+  def get_current_published_version_of_manual(manual_record)
+    return nil if manual_record.nil?
```
NeilvB: As above

app/exporters/manual_publishing_api_exporter.rb
h-lame: I don't think we would expect differences, no.  I'll think about what a good place to store this might be for all of them.

spec/exporters/publishing_api_publisher_spec.rb
h-lame: Oops, yes.  Didn't copy and paste this, no 😉.  Fixed in a rebase.

app/repositories/versioned_manual_repository.rb
h-lame: Yup, fixed in a rebase.

app/repositories/versioned_manual_repository.rb
h-lame: Yup, fixed in a rebase.

app/repositories/versioned_manual_repository.rb
h-lame: Yup, fixed in a rebase.

app/exporters/manual_publishing_api_exporter.rb
h-lame: Added a new module included in all 3 exporters/publishers in b0447a559c8811594454188ffd3b7f4d5352f9b2

PR #808, opened 2017-02-23
Remove ManualsPublisherWiring

chrisroos:
Supersedes PR #798.

This branch simplifies the app by removing the level of indirection introduced in the `ManualsPublisherWiring` class.

The changes in this branch and the feedback on PR #798 have revealed a number of further questions/improvements that we'll tackle separately:

* Investigate whether we need to call both v1 and v2 of the Publishing API (#809).
* Investigate the difference between rendering pipelines of `ManualDocumentRenderer` and `SpecialistDocumentRenderer` (#810).
* Introduce a `Services` module to hold singleton instances of various API clients (#811).


floehopper:
In [this comment](https://github.com/alphagov/manuals-publisher/pull/798#issuecomment-276694981) @h-lame recommended only merging this when #799 had been merged. That PR has since been merged, so there's no longer any need to hold off on this one.

floehopper:
I'm going to rebase this on master and force-push in preparation for merging.

PR #812, opened 2017-02-24
Remove SpecialistDocumentEditions with non-manual document_types

floehopper:
These non-manual SpecialistDocumentEditions should've been removed when the related specialist documents were moved to the newer [Specialist Publisher application][1] around the time of [this large commit][2].

I've tested this migration against a dump of the production database taken on 2017-02-24. I used a couple of temporary scripts to confirm that the migration was doing what I intended and that deleting the non-manual SpecialistDocumentEditions doesn't leave the database in an inconsistent state. See the individual commit notes for further details.

Given that it's no longer possible to add non-manual SpecialistDocumentEditions via the user interface, I think it's safe to rely on the results of the above test, even if the state of the production database changes before this migration is run.

It would be a lot of work to provide a downwards version of the migration, because data is lost in the upwards version. So, given that a downwards version is unlikely to be necessary, I decided not to write one and am raising an `IrreversibleMigration` exception instead.

[1]: https://github.com/alphagov/specialist-publisher
[2]: https://github.com/alphagov/manuals-publisher/commit/3162f0e72fd53bfd1cf87f5661bb198ff785940e


floehopper:

floehopper:

floehopper:
@steventux: You seemed to be suggesting that only the latest editions of specialist documents were sent to the Publishing API/Content Store when the Specialist Publisher app was split off? Is that correct? Do we need to try to recover the data which was deleted in this migration?

steventux:
@floehopper I know that at the end of separating specialist-publisher from manuals-publisher there was a discussion about backing up the specialist publisher workflow history. By this I mean superseded editions of specialist documents which predated the publishing API. The intent here is to use this data as the basis of a more complete content history (back-filling the publishing API and content store with all known workflow for a given document). The concern was that we preserve this data as there's a project focusing on Content History starting soon which would be looking to use this data.

floehopper:
@steventux: Thanks. I'll put the database backup I have somewhere safe.

PR #813, opened 2017-02-24
Remove specialist document related code

floehopper:
This is mostly removing code which has become unused since specialist documents were moved out of this app and into the Specialist Publisher. I'm sure there's more we can do along these lines, but I think it's a reasonable start.

I've also made some specs more realistic by using a "manual" document type instead of various non-"manual" types.

The removal of the `SpecialistDocumentEdition#extra_fields` field means it would probably be safest to merge this after the migration in #812 has been run in production, although I suspect it's not strictly necessary.


floehopper:
@chrisroos: Please can you mark this as reviewed - the review by @chrislo doesn't seem to count at the moment. Thanks.

chrisroos:
@floehopper is doing some further investigation based on the question from @chrislo. I'll wait until that's complete before marking this as approved.

floehopper:
I've rebased this against master, edited a commit message, and force-pushed.

floehopper:
@chrislo, @chrisroos:

> should this query have checked the opposite i.e that extra_fields is NOT nil or != [] ?

You were right to question this. I've [updated the commit note](https://github.com/alphagov/manuals-publisher/pull/813/commits/5f18aba14c9cbe60a82ecd8e0bcdb015de90a3fa) to list the results of some more relevant queries. Hopefully that makes a bit more sense.

Can you take a look?

chrislo:
@floehopper the new commit message is much clearer.

PR #814, opened 2017-02-27
Add script for PDF attachment report

NeilvB:
[Trello card](https://trello.com/c/JIT8o1Tn/90-pdf-report-for-natalie-baron-com)

This script produces a CSV of counts of published PDF attachments per
organisation owning the associated manual. This was produced for the
Content Operating Model team.

### DIFF
--- a/bin/content_model_pdf_report
+++ b/bin/content_model_pdf_report
@@ -1,0 +1,84 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../config/environment", __FILE__)
+require 'csv'
+
+POST_PUBLICATION_STATES = %w(published archived).freeze
+
+first_period_start_date = ENV.fetch('FIRST_PERIOD_START_DATE', Date.parse('2016-01-01'))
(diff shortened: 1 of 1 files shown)


chrislo:
@NeilvB - we booked in a deploy for Thursday, so feel free to merge this before then if you'd like it to go out as part of that deploy.

NeilvB:
Thanks @chrislo !

chrislo:
In a discussion on the GDS Slack (#manuals-publisher channel, not publicly archived) @NeilvB pointed out that this script may be used by the requester again in the future. @h-lame was concerned that if we delete it from the codebase after running it may make it more difficult for future people to find it. 

Given those two constraints, and that some time might pass before the script needs to be run again,  I think this script should have some tests added so that we can be more confident that our future refactorings don't break it.

chrislo:
@NeilvB could you also add some of the relevant context for this request to the PR from the non-public Trello card, just in case someone else needs to refer to it in the future?

chrisroos:
Note that tests were added for this script in PR #830.

PR #818, opened 2017-03-06
Change radio labels

keithiopia:
There's been some confusion over which option to choose. Made it clearer that the second radio button changes both the first published and updated dates. Connected to https://govuk.zendesk.com/agent/tickets/1901209

### DIFF
--- a/app/views/manuals/edit_original_publication_date.html.erb
+++ b/app/views/manuals/edit_original_publication_date.html.erb
@@ -15,10 +15,10 @@
         <%= f.datetime_select :originally_published_at, { include_blank: true, start_year: Date.today.year, end_year: 1945 }, { class: 'date form-control' } %>
       </div>
       <div class="checkbox add-vertical-margins">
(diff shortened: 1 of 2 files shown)


h-lame:
@keithiopia - good call, the confusion is pretty clear.  Double so as your changes are actually the wrong way around!  

Choosing either option will always change the "First published at".  Choosing the first button (originally labelled `Change the GOV.UK publication date.`) will also update the "Public updated at", choosing the last button (originally labelled `Keep the original GOV.UK publication date.`) will not update the "public updated at".

So your new labels need to be flipped: the first should be `Change the first published and last updated date.` and the last should be `Change the first published date.`.

Of course, if what I've said is not the behaviour that the app is exhibiting we need to investigate a bit more as what I've described is the intention of the buttons.

keithiopia:
@h-lame Bugger. I'm going to double check in integration - I think my change reflects what I saw in there, but...

keithiopia:
@h-lame You were right, just a test of course. You passed!
(I've swapped them around)

floehopper:

PR #819, opened 2017-03-06
Remove `bin/fix_inconsistent_edition_states`

chrisroos:
We noticed this script when renaming
`specialist_document_editions`. As far as we can tell it was
introduced in c3fa17e56 and the behaviour hasn't changed since. The
embedded tests at the bottom of the file do not run on current master,
and are not part of the automated build.


chrisroos:
@h-lame: Do you happen to know whether we've had to use this script recently? If not, is there someone else that we should ask before deleting it?

h-lame:
@chrisroos / @chrislo nah - I can't remember being asked to run this.  Maybe @steventux remembers?  I'd say it's certainly safe to delete though.  Any inconsistent states we find in the data would be between the app and publishing-api so would be solved in a different way.

PR #821, opened 2017-03-07
Rename SpecialistDocument js to MarkdownPreview

chrisroos:
This JavaScript is used to add a markdown preview function to the edit
pages for Manuals and Sections. It feels like `MarkdownPreview` is a
better name for it than `SpecialistDocument`.

We came across this because we're naming `SpecialistDocument` to
`Section` in another branch. When it came to renaming this we felt
that giving it a more generic name made more sense.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -4,7 +4,7 @@
 //= require govuk_toolkit
 //= require ajax_setup
 //= require length_counter
-//= require specialist_documents
+//= require markdown_preview
 //= require toggle_display_with_checked_input
 
 function initPrimaryLinks(){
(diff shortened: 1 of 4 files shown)


chrisroos:
> I'm not totally convinced about this namespace rename, because the addPreviewFeature function (a) duplicates part of the new namespace and (b) implies that the intention is/was to have multiple "features" within a single namespace.
> 
> I might be more comfortable with it if the function was also renamed from addPreviewFeature to init or initialise.
> 
> Anyway, I'm going to approve this and let you decide whether to do anything about my comments.

We've had a look at the other JavaScript in the app and they don't appear to share a consistent approach/style. We think the rename in this PR is an improvement so we'll merge it as it is and come back to the style if/when it becomes a problem.

Issue #816, opened 2017-03-06
Tests fail with new bundler on CI


tijmenb:
We're seeing these errors on master:

```
LoadError: cannot load such file -- active_model/translation
/var/lib/jenkins/bundles/manuals-publisher/deployed-to-production/ruby/2.2.0/gems/activemodel-3.2.22.3/lib/active_model/validations.rb:49:in `block in <module:Validations>'
/var/lib/jenkins/bundles/manuals-publisher/deployed-to-production/ruby/2.2.0/gems/activesupport-3.2.22.3/lib/active_support/concern.rb:121:in `class_eval'
/var/lib/jenkins/bundles/manuals-publisher/deployed-to-production/ruby/2.2.0/gems/activesupport-3.2.22.3/lib/active_support/concern.rb:121:in `append_features'
```

This has also been noted in https://github.com/alphagov/manuals-publisher/commit/0a674b5352d7a9de7df04aefd449535abf65a210.

tijmenb:
https://github.com/alphagov/govuk-content-schemas/pull/555 should be reverted after this is fixed.

h-lame:
Don't really know what the cause is but it works with bundler 1.12.6 (latest 1.12 series) and fails with bundler 1.13.0 (first in bundler 1.13 series).  This might help with tracking it down - although nothing obvious jumps out at me from the bundler changelog.

h-lame:
You can trigger it by running `bundle exec rake spec:javascript` or `bundle exec rake cucumber`.  `bundle exec rake spec` works, `bundle exec rake test` runs the specs and then fails when it tries to run the cucumber features.  The problem may lurk in the cucumber and jasmine rakefiles and how they require things.

floehopper:
I assume this has been triggered by the upgrade of Bundler for Ruby v2.2.4 from v1.10.6 to v1.14.5 in alphagov/govuk-puppet@857083c4431ced07b910578d50a9450c39fbea95.

I think the problem is to do with the `mongoid_rails_migrations` gem; more specifically because the gem calls `Bundle.require` when it is required by Bundler - see [this line](https://github.com/adacosta/mongoid_rails_migrations/blob/460a92dcf6fcca8b0cf82f6bd6457e4a4fe31442/lib/mongoid_rails_migrations.rb#L3).

[...]

PR #822, opened 2017-03-07
Use fork of mongoid_rails_migrations to fix build

floehopper:
* This is an attempt to fix the problem described in #816. It uses a forked version of mongoid_rails_migrations. The branch in that fork was branched off the v1.0.0 tag, which is the version of the gem that was specified in our `Gemfile`. This is so that we keep the changes to our app to a minimum.

* This fix is a bit of a bodge - the proper fix would be along the lines of adacosta/mongoid_rails_migrations#3, but given that the failing build is a significant blocker, it feels as if this is worth a try. If it works OK, we can work on getting the proper fix in place.

* I installed v1.14.5 of Bundler which is the version [now specified in govuk-puppet][1] for Ruby v2.2.4. Hence the change to `BUNDLED WITH` in the `Gemfile`.

* I've run a migration locally to ensure that the gem still works OK even with this patch in place.

[1]: https://github.com/alphagov/govuk-puppet/blob/e2d61705aaf6da180a781b6f46f83235fd890df0/modules/govuk_rbenv/manifests/all.pp#L46

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem "govuk_admin_template", "~> 4.0" # higher versions require rails 4
 gem "kaminari", "0.16.1"
 gem "logstasher", "0.4.8"
 gem "mongoid", "2.5.2"
-gem "mongoid_rails_migrations", "1.0.0"
+gem "mongoid_rails_migrations", git: "https://github.com/alphagov/mongoid_rails_migrations", branch: "avoid-calling-bundler-require-in-library-code"
 gem "multi_json", "1.10.0"
 gem "plek", "1.12.0"
 gem "quiet_assets", "1.0.3"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,14 @@
+GIT
+  remote: https://github.com/alphagov/mongoid_rails_migrations
+  revision: 4db234408a335d336a45f840e0f72c585031859b
+  branch: avoid-calling-bundler-require-in-library-code
+  specs:
+    mongoid_rails_migrations (1.0.0)
+      activesupport (>= 3.2.0)
+      bundler (>= 1.0.0)
+      rails (>= 3.2.0)
+      railties (>= 3.2.0)
+
 GEM
   remote: https://rubygems.org/
   specs:
@@ -185,11 +196,6 @@ GEM
       activemodel (~> 3.1)
       mongo (~> 1.7)
       tzinfo (~> 0.3.22)
-    mongoid_rails_migrations (1.0.0)
-      activesupport (>= 3.2.0)
-      bundler (>= 1.0.0)
-      rails (>= 3.2.0)
-      railties (>= 3.2.0)
     multi_json (1.10.0)
     multi_test (0.1.2)
     multi_xml (0.6.0)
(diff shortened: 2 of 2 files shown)


floehopper:
Note that the newer version of Bundler seems to cause a bunch of warnings when installing the bundle from scratch:

```
Unable to use the platform-specific () version of activesupport (3.2.22.3) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of sprockets (2.2.3) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of cucumber (2.2.0) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of railties (3.2.22.3) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of cucumber-rails (1.4.5) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of domain_name (0.5.20161129) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of faraday (0.9.0) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of rest-client (2.0.0) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of oauth2 (1.3.0) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of omniauth (1.4.2) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of rubocop (0.39.0) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of jquery-rails (3.1.4) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of jasmine-rails (0.14.1) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Fetching https://github.com/freerange/mongoid_rails_migrations
Unable to use the platform-specific () version of quiet_assets (1.0.3) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of redis-namespace (1.5.0) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of rspec-expectations (3.2.1) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of rspec-mocks (3.2.1) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of rspec-rails (3.2.3) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Unable to use the platform-specific () version of sinatra (1.4.4) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
```

However, as far as I can tell those warnings don't seem to be causing any problems at the moment.

floehopper:
The build is passing. Yay! 🎉

PR #828, opened 2017-03-08
Fix typo in export_recipent -> export_recipient

floehopper:
This typo is present in many, but not *all* cases which made things doubly confusing.


floehopper:
@chrislo: Thanks.

PR #831, opened 2017-03-09
Remove factory argument from ManualRepository.new

chrislo:
Although this method is written to use a factory to build objects, in practice it only ever builds instances of Manual, so we can replace the use of a factory with a call to `Manual.new`


chrislo:
Rebase and force push before merging.

PR #835, opened 2017-03-09
Use class instances instead of anonymous functions

chrisroos:
PR #808 introduced a number of class methods that return anonymous functions.

This PR removes those class methods in favour of using instances of the class instead. I think this makes the code easier to follow.


chrisroos:

PR #834, opened 2017-03-09
Remove unused RepublishDocumentWorker

floehopper:
This has not been used since [this commit][1].

[1]: https://github.com/alphagov/manuals-publisher/commit/43e2e7534da179f92bed2ce7e73bc3f3c4fdfb06

PR #836, opened 2017-03-09
Remove redundant condition from case statement in ManualsPublisher.document_services

floehopper:
Note that this branch is based off the branch in #834.

As far as I can see `ManualsPublisher.document_services` is now only ever invoked from the `bin/republish_withdrawn_document` script. The value of the `document_type` argument passed into it is always obtained from a call to `SpecialistDocumentEdition#document_type` which will never be 'manual_document'; only 'manual' or one of the Specialist Document document types, e.g. 'cma_case'. Indeed since [this commit][1], it will only ever be 'manual'. Thus it's safe to remove this condition and simplify the code.

[1]: https://github.com/alphagov/manuals-publisher/commit/723ce320c9845003bf1aa584d4ea726b587c1891

### DIFF
--- a/app/lib/manuals_publisher.rb
+++ b/app/lib/manuals_publisher.rb
@@ -2,10 +2,9 @@ module ManualsPublisher
   extend self
 
   def document_services(document_type)
-    builder = ManualBuilder.create
     AbstractDocumentServiceRegistry.new(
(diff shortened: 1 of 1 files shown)


floehopper:
@chrislo: Thanks. Rebasing against master and force-pushing in preparation for merging.

PR #837, opened 2017-03-10
Rename specialist document

chrislo:
This PR replaces the use of the term "specialist document" with the term "section". The publishing of what are known as Specialist Documents was moved to a separate application [some time ago](). However the name still remains in this codebase and seems to be used to refer to a section of a manual. We've gone with `Section` as opposed to something like `ManualSection` as the application currently only has responsibility for manuals and their sections and this seems to match the language used in the UI. 

This PR does not currently touch any data - the underlying mongodb collection for `SectionEdition` is still `specialist_document_editions`. I'd like someone to help check that this continues to work with a recent copy of a production-like database. 

I think we'll rename the collection in a subsequent commit after this one has landed as we need to do some investigation about whether that would involve application downtime.


chrisroos:
> This PR does not currently touch any data - the underlying mongodb collection for SectionEdition is still specialist_document_editions. I'd like someone to help check that this continues to work with a recent copy of a production-like database.

I took a recent dump of production data and checked out this branch locally. I clicked around the manuals-publisher UI and can confirm that everything appears to work as expected.

chrislo:
Thanks @chrisroos - I've rebased this against master so I think it's ready for review.

chrisroos:
I've added issue #848 to prompt us to rename the Mongo collection once this has been merged.

spec/repositories/versioned_manual_repository_spec.rb
```diff
@@ -38,19 +38,19 @@
...
       end
 
-      it "has the first draft of the specialist document editions as Specialist Document instances attached" do
+      it "has the first draft of the section editions as Section instances attached" do
         documents = subject.documents.to_a
         expect(documents.size).to eq 2
 
         document_1 = documents[0]
-        expect(document_1).to be_a ::SpecialistDocument
+        expect(document_1).to be_a ::Section
```
floehopper: Do you know why the `::` prefix is needed here and elsewhere in this file?

app/models/document_factory_registry.rb
```diff
@@ -6,7 +6,7 @@
 require "builders/manual_document_builder"
 require "manual_with_documents"
 require "slug_generator"
-require "specialist_document"
+require "section"
```
floehopper: Is this require statement still necessary now that the classname matches the filename?

app/models/attachment.rb
```diff
@@ -7,7 +7,7 @@ class Attachment
   field :file_id, type: String
   field :file_url, type: String
 
-  embedded_in :specialist_document_edition
+  embedded_in :section_edition
```
floehopper: I think it would be useful to mention why this change does not require any database migration, i.e. can you confirm that none of the existing attachments store a reference to their parent edition using this key. I'm guessing it's because this is an embedded document...?

app/repositories/repository_registry.rb
```diff
@@ -1,13 +1,12 @@
 require "specialist_document_repository"
 require "manual_repository"
-require "specialist_document_edition"
+require "section_edition"
```
floehopper: Is this require statement necessary now that the class name matches the filename?

floehopper:
There are a bunch of classes named `*ManualSection*`. Given the choice to rename `SpecialistDocument` to `Section`, we will want to rename them to `*Section*`.

chrisroos:
I'm going to rebase this on master given all the recent changes.

chrisroos:
I've rebased and force pushed.

I'm going to hold off merging until I can pair with @chrislo to address @floehopper's comments and to double check that we're happy with the name `Section` instead of `ManualSection`.

app/models/attachment.rb
chrislo: Yep, that's correct. I've addressed this comment in an improved commit message on b0b409ad6.

chrislo:
I've raised https://github.com/alphagov/manuals-publisher/issues/867 to make sure we keep track of the suggestion to rename `ManualSection` to `Section`

spec/repositories/versioned_manual_repository_spec.rb
chrislo: No, it struck us as odd too - will investigate after merging.

app/models/document_factory_registry.rb
chrislo: Possibly not - I think we'll take a look at this after we merge this PR, there's probably a few unnecessary requires scattered around.

chrislo:
Rebasing against master and force pushing before merging.

PR #844, opened 2017-03-10
Remove CMA and DCLG factories

chrisroos:
There doesn't appear to be anything special about writers/editors from CMA or DCLG. We can make that explicit by using the `generic_writer`, `generic_editor` and `gds_editor` factories instead.

### DIFF
--- a/features/step_definitions/access_control_steps.rb
+++ b/features/step_definitions/access_control_steps.rb
@@ -41,7 +41,7 @@ Then(/^I see manuals created by all organisations$/) do
 end
 
 Given(/^I am logged in as a writer$/) do
-  login_as(:cma_writer)
+  login_as(:generic_writer)
   stub_organisation_details(GDS::SSO.test_user.organisation_slug)
 end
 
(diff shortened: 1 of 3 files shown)


features/step_definitions/access_control_steps.rb
```diff
@@ -9,13 +9,13 @@
 end
 
 Given(/^there are manuals created by multiple organisations$/) do
-  login_as(:cma_editor)
+  login_as(:gds_editor)
```
floehopper: I think it might be better to introduce a `generic_editor_from_another_organisation` factory or similar to avoid overloading the use of `gds_editor` and I think this would make the intended behaviour clearer.

chrisroos:
Thanks for the review, @floehopper. While trying to address your feedback I've noticed a few other tests that appear to be CMA specific. I'm going to see whether it makes sense to remove them in this PR too.

chrisroos:
I'm closing this in favour of PR #864 that contains an expanded version of the changes in this branch.

PR #845, opened 2017-03-10
Remove broken republish_withdrawn_document script

floehopper:
This script relies on `ManualsPublisher#document_services` which is broken in a couple of different ways:

* The `AbstractDocumentServiceRegistry` class was removed in [this commit][1].
* The `ManualsPublisher#observer_registry` method was removed in [this commit][2].

Writing about this script in [this PR comment][3], @h-lame said:

> Suspect this could be got rid of and we re-write as/when requested via a support ticket.

So I think it's safe to remove the script. It's then sensible to remove the `ManualsPublisher` module which at that point is unused as well as broken.

[1]: https://github.com/alphagov/manuals-publisher/commit/43e2e7534da179f92bed2ce7e73bc3f3c4fdfb06
[2]: https://github.com/alphagov/manuals-publisher/commit/9bbf2d34128fc53c5ead049240cc64d732b652f9
[3]: https://github.com/alphagov/manuals-publisher/pull/832#issuecomment-285417474


floehopper:
@h-lame: Yes. Thanks - I've just been fixing it. I've also added a new commit to remove another method.

PR #847, opened 2017-03-10
Remove unused ManualPublishingApiBulkDraftExporter

floehopper:
I've created this PR based on the less controversial commits from #832.

The two places which are using v1 of the Publishing API have been broken since [this commit][1], because they rely on methods which were [removed from gds-api-adapters in v38.0.0][2]; the application is currently using v39.2.0. This PR removes a bunch of code around the invocations of these non-existent methods on the basis that the code is currently useless.

I'm very confident that `ManualPublishingApiBulkDraftExporter` and its associated code can be removed, because the rake task that was using it was removed in [this commit][3].

Quoting from [a comment](https://github.com/alphagov/manuals-publisher/pull/832#issuecomment-285357242) by @h-lame on #832:

> I'm :+1: on the 2nd half of removing the bulk exporter as I don't recall that being used recently.

Note that the first commit in this PR leaves an exception being raised in `PublishingAPIRedirecter#call`, but I think that's still an improvement, because the code was already broken - it's just that now the root cause is made clearer.

[1]: https://github.com/alphagov/manuals-publisher/commit/11c9cfe237ea3150a3ead96da1c1a349b53d2323
[2]: https://github.com/alphagov/gds-api-adapters/blob/master/CHANGELOG.md#3800
[3]: https://github.com/alphagov/manuals-publisher/commit/a739e78e2d0e7c926ad110ba9e57181cb9e40f34


floehopper:
Does anyone know why the snyk check is failing for this PR, but is passing for others?

chrisroos:
> Does anyone know why the snyk check is failing for this PR, but is passing for others?

@floehopper: I saw this happen on one of my PRs last week. I had to force push the branch a couple of times to force snyk to re-check the code.

I also asked about the problem in the govuk-developers Slack channel last week and was told that we don't have to worry too much about snyk problems. Apparently it'll sometimes complain about things for which there isn't currently a fix!

floehopper:
Rebasing against master and force pushing in the hope that that fixes the snyk build failure and allows me to merge the PR.

PR #849, opened 2017-03-13
Remove SpecialistDocumentEdition#document_type

floehopper:
Since [this commit][1] all editions have had a `document_type` value of 'manual' and so the field had become redundant.

Unfortunately the string literal, 'manual', appeared rather a lot throughout the code, so some of the early commits extract these occurrences out into a number of different constants in order to better express their meaning. A couple of these which internal to the app are then gradually refactored away, leaving only ones relating to the Publishing API and Rummager.

For simplicity I've decided not to take the final step of unsetting the `document_type` field on all existing persisted instances of `SpecialistDocumentEdition` within this pull request, but I would expect to do it very shortly.

[1]: https://github.com/alphagov/manuals-publisher/commit/723ce320c9845003bf1aa584d4ea726b587c1891

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,8 +20,7 @@ class ApplicationController < ActionController::Base
   helper_method :current_user_can_publish?
 
   def current_user_can_withdraw?
-    format = PermissionChecker::MANUAL_FORMAT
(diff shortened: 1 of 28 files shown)


floehopper:
@chrisroos: Thanks.

PR #864, opened 2017-03-13
Remove references to CMA and DCLG

chrisroos:
This app used to be responsible for publishing CMA (and possibly DCLG?) content. CMA content has now moved to specialist-publisher. I've removed references to CMA and DCLG to make it clear that there's nothing special about those organisations in this app.

This is an expanded version of PR #844.

### DIFF
--- a/spec/lib/duplicate_draft_deleter_spec.rb
+++ b/spec/lib/duplicate_draft_deleter_spec.rb
@@ -8,7 +8,7 @@ describe DuplicateDraftDeleter do
   it "deletes duplicate editions that aren't present in Publishing API" do
     original_content_id = SecureRandom.uuid
     FactoryGirl.create(:section_edition,
-      slug: "cma-cases/a-case",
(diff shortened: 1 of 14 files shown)


spec/lib/duplicate_draft_deleter_spec.rb
```diff
@@ -8,20 +8,20 @@
   it "deletes duplicate editions that aren't present in Publishing API" do
     original_content_id = SecureRandom.uuid
     FactoryGirl.create(:specialist_document_edition,
-      slug: "cma-cases/a-case",
+      slug: "guidance/slug",
```
floehopper: Since this factory returns an edition of a manual *section*, I think a more realistic path would be `guidance/manual-slug/section-slug`, would it not? Same comment for other occurrences in this commit.

chrisroos: Good idea, @floehopper. I'll make that change.

chrisroos:

PR #865, opened 2017-03-13
Inline service registries into ManualDocumentAttachmentsController

floehopper:
The overall aim is to reduce the levels of indirection in the code to make it easier to follow. In some cases, this means temporarily introducing duplication, but we're confident that in the long run this will make it easier to identify better abstractions to remove that duplication.

The net effect is to inline most of the methods from `AbstractManualDocumentAttachmentServiceRegistry` and its two sub-classes, `ManualDocumentAttachmentServiceRegistry` & `OrganisationalManualDocumentAttachmentServiceRegistry` into `ManualDocumentAttachmentsController`.


floehopper:
I've now talked this through with @chrislo and we're happy that it's a sensible approach. So @chrisroos / @chrislo can you review this PR for me? Thanks.

chrislo:
I think this is a good approach - it moves us towards a slightly more conventional pattern in the controllers. As we discussed earlier, there's effectively two branches the code can take based on the `if current_user_is_gds_editor?` conditional and that decision is already made in the controller already - I think removing the additional indirection makes things much easier to understand.

app/controllers/manual_document_attachments_controller.rb
```diff
@@ -1,6 +1,17 @@
...
+require "update_manual_document_attachment_service"
+require "show_manual_document_attachment_service"
+require "new_manual_document_attachment_service"
+
 class ManualDocumentAttachmentsController < ApplicationController
   def new
-    manual, document, attachment = services.new(self).call
+    service = NewManualDocumentAttachmentService.new(
+      repository,
+      # TODO: This be should be created from the document or just be a form object
```
chrislo: Small typo in this comment, but perhaps we should move it to an issue?

app/controllers/manual_document_attachments_controller.rb
```diff
@@ -41,21 +64,21 @@ def update
...
+  def gds_editor_repository
+    RepositoryRegistry.create.manual_repository
   end
 
-  def organisational_services
-    OrganisationalManualDocumentAttachmentServiceRegistry.new(
-      organisation_slug: current_organisation_slug,
-    )
+  def organisational_repository
+    manual_repository_factory = RepositoryRegistry.create
```
chrislo: There's probably some additional indirection we can remove here, but I think we'll see it if we refactor all of the controllers to use this pattern.

app/controllers/manual_document_attachments_controller.rb
floehopper: Yeah. I think this should be dealt with separately - I think it's clearer if this code matches the code which it was taken from as closely as possible, i.e. the typo in the original is exactly replicated here.

app/controllers/manual_document_attachments_controller.rb
floehopper: Agreed.

floehopper:
I've expanded the description to better explain what's going on.

floehopper:
I'm going to rebase against master and force-push in preparation for merging.

PR #868, opened 2017-03-14
Remove migrations that have already run on production

chrislo:
Comparing `master` with `deployed-to-production` (with `git diff
master..deployed-to-production db/migrate`) shows that all the
migrations in `db/migrate` have already been run on production.

Many of these migrations will no longer run because they refer to code
that has been removed or renamed, and leaving them around makes
renaming code more difficult.

What do you think about doing this?


h-lame:
Not my place to argue, but I'm 50:50 on this.  On the one hand, migrations are records of state at a single point in time and we shouldn't spend too much effort maintaining them (doubly so on a schemaless db like mongo) so deleting them now they've been run makes sense.  On the other though, I know that when doing one-off data massaging it's been super useful to look in this folder to find the kinds of things we've done in the past for "inspiration", so deleting them could be detrimental to future devs wanting a reference.  On the other other hand, the refactoring you're doing will mean these will be woefully out of date as a reference so deleting them makes sense.

I may have just typed myself into being 90:10 in favour now.  As you were.

chrislo:
On the contrary @h-lame it's great to get your input - I didn't request a review from you as I know you're busy with other projects, but do jump in at anytime. 

I agree with you that some of the older migrations are so out-of-date now that they're probably not that useful as a reference. Perhaps if we fix up and maintain some of the "scripts" they will serve that purpose. Perhaps I should add a reference to the commit sha that removes these migrations to the README to make it easier for people to find them in the future if they need them?

chrislo:
Rebasing and force-pushing before merging.

Issue #867, opened 2017-03-14
Rename `ManualSection` to `Section`


chrislo:
Arrises from: https://github.com/alphagov/manuals-publisher/pull/837#issuecomment-286146673

There are a number of classes with `ManualSection` in their name, for example `ManualSectionIndexableFormatter`. [As we've settled on](https://github.com/alphagov/manuals-publisher/pull/837) `Section` as the name of the parts that make up a manual, we should rename these classes to be consistent.

PR #869, opened 2017-03-14
Rename manual section

chrislo:
Fixes: #867 

This PR removes references to "manual section" and replaces them with "section". I've left the use of `manual_section` in place where this app calls the PublishingApi as that API uses this namespacing. I've also done the same with `SectionIndexableFormatter::RUMMAGER_DOCUMENT_TYPE` so that the schema of the search database does not change. 

I haven't made this rename in migrations as all of them have already been run against production (see #868).


chrislo:
Rebasing/force-pushing to keep branch up-to-date.

chrislo:
I've updated the description and merged. Thanks @floehopper.

PR #870, opened 2017-03-14
Fail fast if the Asset Manager isn't available

chrisroos:
I was seeing an exception in development when trying to save a Section
with an attachment. I eventually realised that it was because I didn't
have the Asset Manager running and therefore the attachment couldn't be
uploaded. There was no indication of this problem in the user interface.
It looks as though the call to `errors.add` was intended to do something
sensible in the case of an error but it doesn't appear to have any
effect. The result is that errors are essentially swallowed and the user
is led to believe that the attachment has been uploaded successfuly.

In order to add a test for the new behaviour I've made the
`#upload_file` method public.

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -46,12 +46,12 @@ class Attachment
         response = AttachmentApi.client.update_asset(file_id, file: @uploaded_file)
       end
       self.file_url = response["file_url"]
+    rescue GdsApi::HTTPNotFound => e
+      raise "Error uploading file. Is the Asset Manager service available?\n#{e.message}"
     rescue StandardError
       errors.add(:file_id, "could not be uploaded")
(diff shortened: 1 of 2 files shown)


chrisroos:
> Minor: I wonder whether the wording of the exception is correct. Is it possible for the AssetManager service to be available, but for the API to return a 404 error status? If so, it might be better for the exception message to say something more like "Error uploading file to Asset Manager service".

You're quite right, @floehopper. It is possible for the Asset Manager service to be available and for the API to respond with a 404. I observed this behaviour in the Rails console locally when calling `AttachmentApi.client.update_asset` with a file id that doesn't exist. I've updated the exception message so that it only offers the Asset Manager being down as a possibility rather than it being a certainty.

Issue #851, opened 2017-03-13
Fix or remove reslug_manual_section script


floehopper:
We should decide whether or not it's worth keeping this script. If we do decide to keep it, we'll need to fix it and ensure that it remains fixed by adding the appropriate test coverage.

From [this comment][1] by @floehopper:
> The two places which are using v1 of the Publishing API have been broken since [this commit][2], because they rely on methods which were [removed from gds-api-adapters in v38.0.0][3]; the application is currently using v39.2.0. This PR removes a bunch of code around the invocations of these non-existent methods on the basis that the code is currently useless.
>
> [...]
>
> I'm slightly less confident that the `bin/reslug_manual_section` script and its associated code can be removed, but given that it is currently broken, if someone wants to keep it, it will need to be fixed. I would also argue that it should only be kept if we write a test for it to ensure that it stays in a working state.

From [this comment][4] by @h-lame:
> I think you could replace the `put_content_item` call in `PublishingAPIRedirecter` with a call to `put_content` on a publishing_api_v2 adapter.  I'm fairly sure these are equivalent.
> 
> The `bin/reslug_manual_section` script is something we wrote last year in response to a support ticket, and I think we've had to run it more than once (but obvs not since I broke it with a gem bump).  We get a lot of requests to do this kind of thing to HMRC manuals (which live in a different app), and it's conceivable we'd get other requests to do it on these, as you can't change the slug once it's published (I think).
>
> I think it's probably 50:50 on whether we drop that script as you've done already, or fix the underlying thing and add tests.  If we chose to fix it I suspect it would be worth looking at the `ManualRelocator` in lib.  This object works across an entire manual rather than a single section, but it's more recent, and has some tests.  Maybe a good approach would be to remove the `PublishingAPIRedirecter`, and replace it with some functionality extracted from `ManualRelocator`?  I reckon it's up to you as current custodians though, as that sounds like a lot of work.

[1]: https://github.com/alphagov/manuals-publisher/pull/832#issue-213028165
[2]: https://github.com/alphagov/manuals-publisher/commit/11c9cfe237ea3150a3ead96da1c1a349b53d2323
[3]: https://github.com/alphagov/gds-api-adapters/blob/master/CHANGELOG.md#3800
[4]: https://github.com/alphagov/manuals-publisher/pull/832#issuecomment-285357242

brenetic:
This is a useful script for second line on Gov.uk. Would be nice to keep it around. 

[...]

PR #872, opened 2017-03-15
Inline service registries into SectionsController

floehopper:
This is similar to #865, but for the `SectionsController` instead of the `SectionAttachmentsController`. The overall aim is to reduce the levels of indirection in the code to make it easier to follow. In some cases, this means temporarily introducing duplication, but we're confident that in the long run this will make it easier to identify better abstractions to remove that duplication.

The net effect is to inline most of the methods from `AbstractSsectionServiceRegistry` and its two sub-classes, `SectionServiceRegistry` & `OrganisationalSectionServiceRegistry` into `SectionsController`. I've also extracted three anonymous procs returned by methods in `AbstractSectionServiceRegistry` into new named classes to avoid excessive duplication:

* `PublishingApiDraftManualExporter`
* `PublishingApiDraftSectionExporter`
* `PublishingApiDraftSectionDiscarder`

Although I've applied the refactorings to the `TidyUpRemovedDocuments` migration, I haven't tested these changes. However, since we're hoping to remove the commits in #868, I think this is OK.

Although I've applied the refactorings to `SectionReslugger`, this class is not tested. However, since the script which uses this class is already broken (see #851), I think this is OK.

### DIFF
--- a/app/controllers/sections_controller.rb
+++ b/app/controllers/sections_controller.rb
@@ -1,8 +1,14 @@
+require "show_section_service"
+
 class SectionsController < ApplicationController
   before_filter :authorize_user_for_withdrawing, only: [:withdraw, :destroy]
 
   def show
-    manual, section = services.show(self).call
+    service = ShowSectionService.new(
+      services.manual_repository,
+      self,
+    )
+    manual, section = service.call
 
     render(:show, locals: {
       manual: manual,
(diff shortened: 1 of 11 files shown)


floehopper:
Rebasing against master and force-pushing to keep the branch up-to-date.

floehopper:

floehopper:
Note that rebasing against master was a bit painful due to the renaming of `ManualDocument` to `Section` in #887. However, I think I've caught everything - including editing the commit notes!

Issue #811, opened 2017-02-23
Introduce a `Services` module to hold singleton instances of various API clients


chrisroos:
As per [@h-lame's comment on PR #798](https://github.com/alphagov/manuals-publisher/pull/798#discussion_r98921717).

__NOTE.__ This should ideally wait until PR #808 has been merged.

> A pattern that other gov.uk apps have been using is a Services module that has class methods on it for each gds-api used by the app. Might be worth pushing this as far as that too. For example: https://github.com/alphagov/policy-publisher/blob/0f2fa349a043ae7acea37e22bbb08b57154fb30e/lib/services.rb

chrisroos:
This was added/merged in PR #875.

PR #875, opened 2017-03-15
Introduce Services module

chrisroos:
Using a `Services` module to keep all the GdsApi clients together is a pattern used in a number of other GOV.UK apps.

As suggested by @h-lame.

Fixes issue #811.


chrisroos:
> I wonder whether it's worth also moving the `ORGANISATIONS_API_BASE_PATH` constant from its initializer into the new `Services` module. It might be worth seeing what other applications do.

I had a quick look at how other apps are instantiating `GdsApi::Organisations` and can see that they appear to be using `Plek.new.find('whitehall-admin')` directly. Manuals Publisher uses `ORGANISATIONS_API_BASE_PATH` so that it can specify an alternative API endpoint in development. This can already be achieved by using an environment variable to control `Plek`'s behaviour (we made a similar [change to Smart Answers](https://github.com/alphagov/smart-answers/commit/9c01ec59eb2a715cfdf641a7db85e1b84d03677c)). I'm going to leave `ORGANSATIONS_API_BASE_PATH` where it is in this branch but then remove it completely (and rely on Plek) in another branch.

PR #876, opened 2017-03-16
Remove fast_spec_helper

chrisroos:
Remove fast_spec_helper

The use of the fast_spec_helper was making it hard to run some tests in
isolation.

Removing it doesn't seem to have any real effect on the time
taken to run the specs so I'm happy that we're not making anything worse
with this change. I ran the specs (`time bundle exec rspec`) with and
without the fast_spec_helper:

== With fast_spec_helper

real  0m21.615s
real  0m21.809s
real  0m21.661s

== Without fast_spec_helper

real  0m21.622s
real  0m21.576s
real  0m21.796s

I initially tried to add the relevant `require` statements to
ensure that each test would run in isolation while continuing to use the
fast_spec_helper but that was proving problematic.

I used the following Ruby script to run each test in isolation and
confirm that they all pass after this change:

    Dir['./spec/**/*_spec.rb'].each do |spec_file|
      puts "Running #{spec_file}"
      `bundle exec rspec #{spec_file}`
    end


chrisroos:
Thanks, @floehopper.

Issue #854, opened 2017-03-13
Check or remove republish_manuals script


floehopper:
We should decide whether or not it's worth keeping this script. If we do decide to keep it, we'll need to check that it works and ensure that it remains working by adding the appropriate test coverage.

From [this comment][1] by @h-lame:
> `republish_manuals` is useful for when all the content needs to be resent to the publishing-api when we change the schema or so-on (for example we might need to use this if we implement https://trello.com/c/Q01pcyd3/103-send-more-details-to-rummager-from-manuals-publisher)

[1]: https://github.com/alphagov/manuals-publisher/pull/832#issuecomment-285417474

floehopper:
Since #877 this has been fairly well tested by `spec/features/republishing_manuals_spec.rb`. It sounds as if it's still useful if we need to make changes to the content schema and I can see that even a "fully migrated" app like Service Manual Publisher has [similar functionality](https://github.com/alphagov/service-manual-publisher/blob/b0915aed50fe02555b625268b5613eb8037afdf6/lib/tasks/republish.rake). So I'm happy to keep this script around. Closing.

Issue #852, opened 2017-03-13
Check or remove delete_draft_manual & withdraw_manual scripts


floehopper:
We should decide whether or not it's worth keeping these scripts. If we do decide to keep them, we'll need to ensure they work and ensure that they remain working by adding the appropriate test coverage.

From [this comment][1] by @h-lame:
> `delete_draft_manual` and `withdraw_manual` work around functionality missing in the UI - you can't get rid of manuals via the UI even though the functionality exists in the system (we might need to check that these actually do the right thing wrt publishing-api).  We can probably get rid of these once the functionality is in the UI - see https://trello.com/c/PnoMy4LK/73-discard-manual for a ticket about `delete_draft_manual`.

[1]: https://github.com/alphagov/manuals-publisher/pull/832#issuecomment-285417474

floehopper:
Since the changes in #877, `bin/withdraw_manual` has been fairly well tested by `withdrawing-a-manual.feature`. Given that it is also a workaround for functionality missing in the UI, I'm happy to keep it around until we're in a position to move that functionality to the UI.

floehopper:
I'm happy that both scripts are tested and worth keeping. Closing.

PR #877, opened 2017-03-16
Fix and add tests for a couple of scripts

floehopper:
This fixes a couple of obvious problems in the `republish_manuals` script and adds basic test coverage, but only to the extent that no un-handled exceptions are raised, i.e. there may be more subtle problems with the script. I believe this script is used when changes are made to the content schemas, so we'll probably need to keep it around for a while.

It also improves the testing of the `withdraw_manual` script in that it tests more of the code used in the script itself by extracting it out into a class, as opposed to testing code duplicated from the script which is what was happening previously (with the potential for the two to get out of sync). While it is currently possible to withdraw an individual section of a manual via the user interface it isn't possible to withdraw a whole manual. Thus I think we're going to need to keep this script around for a while too.

While this isn't perfect, I feel as if it's enough of a step forward to do some other refactoring with confidence. What's more, we already have #852 & #854 to take the improvements to the scripts a bit further.

### DIFF
--- a/bin/republish_manuals
+++ b/bin/republish_manuals
@@ -1,26 +1,11 @@
 #!/usr/bin/env ruby
 
 require File.expand_path("../../config/environment", __FILE__)
-require "manual_service_registry"
-require "marshallers/document_association_marshaller"
+require "manuals_republisher"
 require "logger"
 
 logger = Logger.new(STDOUT)
 logger.formatter = Logger::Formatter.new
 
-manual_records = ManualRecord.all
-count = manual_records.count
-
-logger.info "Republishing #{count} manuals..."
-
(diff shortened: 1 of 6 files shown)


floehopper:
Since @chrislo reviewed this, I've discovered an RSpec feature that was testing some similar (but not identical) code as that used by the `republish_manuals` script. I've just edited my commits so that instead of introducing a new Cucumber feature, I've changed the existing RSpec feature to use the extracted `ManualsRepublisher` class.

@chrislo or @chrisroos: I'd be grateful if one of you could review it again.

floehopper:
I've just force-pushed again with a small change removing some Cucumber step definitions that were no longer needed.

floehopper:

PR #878, opened 2017-03-16
Remove ManualRelocator

chrislo:
The last reference to this class was removed in f8698ce22. It is
mentioned in #851 as a recent example of useful script that has tests
so we may want to use it in the future to base a similar migration
on. But for now it cannot be used externally via a script or rake task
so I think it is safe to remove it.


h-lame:
You might want to keep this around - it's a complicated task and used specifically in response to support queries that come in.  Responding to support queries may not always be dealt with by committed scripts (although they probably should be).

chrislo:
Thanks for pointing that out @h-lame - if this is still used, I might add and document a small script in `/bin` that allows it to be used from the command line. That way there'll be a reference to this class somewhere in the codebase to remind us that it is used.

PR #880, opened 2017-03-16
Remove params from SectionPublishingAPIExporter constructor

chrisroos:
We weren't using the flexibility of being able to construct a `SectionPublishingAPIExporter` with different `export_recipient`s or `document_renderer`s. Constructing them in the class makes this more explicit.

### DIFF
--- a/app/exporters/section_publishing_api_exporter.rb
+++ b/app/exporters/section_publishing_api_exporter.rb
@@ -4,10 +4,10 @@ class SectionPublishingAPIExporter
   PUBLISHING_API_SCHEMA_NAME = "manual_section".freeze
   PUBLISHING_API_DOCUMENT_TYPE = "manual_section".freeze
 
-  def initialize(organisation, document_renderer, manual, document, update_type: nil)
+  def initialize(organisation, manual, document, update_type: nil)
     @export_recipient = Services.publishing_api_v2.method(:put_content)
     @organisation = organisation
-    @document_renderer = document_renderer
+    @document_renderer = ManualDocumentRenderer.new
     @manual = manual
     @document = document
     @update_type = update_type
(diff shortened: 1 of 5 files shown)


app/services/abstract_manual_document_service_registry.rb
```diff
@@ -125,9 +125,7 @@ def publishing_api_draft_manual_document_exporter
       ).call
 
       SectionPublishingAPIExporter.new(
-        publishing_api_v2.method(:put_content),
```
h-lame: You mean we're no longer passing methods around forcing us to use `.call` everywhere and making things harder to understand?

👏 👏 👏 
➕ 💯

app/exporters/section_publishing_api_exporter.rb
```diff
@@ -113,7 +112,7 @@ def update_type
   end
 
   def rendered_document_attributes
-    @rendered_document_attributes ||= document_renderer.call(document).attributes
+    @rendered_document_attributes ||= ManualDocumentRenderer.new.call(document).attributes
```
floehopper: I'm slightly less convinced about this change, i.e. chaining the calls to `ManualDocumentRenderer.new` and `ManualDocumentRenderer#call`, but perhaps it's OK.

spec/exporters/section_publishing_api_exporter_spec.rb
```diff
@@ -1,4 +1,4 @@
-require "fast_spec_helper"
+require "spec_helper"
```
floehopper: Is this change supposed to be here?

chrisroos: Nope. It'll disappear once I rebase on master now that PR #876 has been merged.

app/exporters/section_publishing_api_exporter.rb
chrisroos: I think I'm happy with it for now given that I suspect it'll change in future commits.

PR #883, opened 2017-03-16
Remove params from ManualPublishingAPIExporter

chrisroos:
We weren't using the flexibility of being able to construct a `ManualPublishingAPIExporter` with different `export_recipient`s, `manual_renderer`s or `publication_log`s. Constructing them in the class makes this more explicit.


chrisroos:

PR #884, opened 2017-03-16
Remove params from ManualPublishingAPILinksExporter

chrisroos:
We weren't using the flexibility of being able to construct a `ManualPublishingAPILinksExporter` with different `export_recipient`s. Constructing it in the class makes this more explicit.


chrisroos:

PR #887, opened 2017-03-16
Rename manual document to section

### LINKED ISSUE #810: Investigate the difference between rendering pipelines of `ManualDocumentRenderer` and `SpecialistDocumentRenderer`
As per [@h-lame's comment on PR #798](https://github.com/alphagov/manuals-publisher/pull/798#discussion_r98923301).

> It feels like it might be a bug that this and the ManualDocumentRenderer you deal with in the next few commits are subtly different; this one doesn't do FootnoteSomethingSomething in its pipeline, but the ManualDocumentRenderer does. We should check obviously, but I think we can get rid of this one and replace it's single use with an instance of the ManualDocumentRenderer instead.

chrislo:
This set of commits removes references to `ManualDocument` and `manual_document` throughout the codebase, replacing them instead with `Section` and `section`. 

It's mostly fairly tedious, but along the way in d88b389 the rename revealed that we could combine `SectionRenderer` and `ManualDocumentRenderer` which fixes #810. It might be a good idea to check that I haven't missed anything obvious there. 

The most complicated commit is 1d69ee0 where the routes containing `manual_document` are renamed. This showed that some of the JS is quite coupled to how Rails generates HTML using its helpers. I will add a couple of small view specs to help catch this in the future as it turned out to be quite hard to debug. 

This PR doesn't leave the language used in the code in an entirely consistent state - there are still a lot of references to `document`, `document_id` and so on. But I think it's a step in the right direction. In d69ec30 I at least try to make the language used in the "English" bits of the feature specs consistent. 

Let me know if you think I should press on and fix all references to "document" too, or if I can land this as-is.


chrisroos:
I've just been through each commit relatively carefully and think it all looks good. I've also checked the branch out and clicked around the UI locally. Everything seems to be working as expected.

I don't think you should do any more renaming in this PR and think we should get it merged ASAP to avoid more merge/rebase pain.

chrislo:
Thanks @chrisroos - I'm going to merge this on the command line so I can add a bit of context to the merge commit.

floehopper:
😭

PR #888, opened 2017-03-16
Simplify OrganisationFetcher

chrisroos:
I think this is easier to understand. Previously the `instance` class
method returned an anonymous function that had to be `call`ed with an
organisation slug.


chrisroos:

PR #890, opened 2017-03-17
Inline service registries into ManualsController

floehopper:
This is similar to #865 & #872, but for the `ManualsController` instead of the `SectionAttachmentsController` or `SectionsController`. The overall aim is to reduce the levels of indirection in the code to make it easier to follow. In some cases, this means temporarily introducing duplication, but we're confident that in the long run this will make it easier to identify better abstractions to remove that duplication.

The net effect is to inline most of the methods from `AbstractManualServiceRegistry` and its two sub-classes, `ManualServiceRegistry` & `OrganisationalManualServiceRegistry` into `ManualsController`.

Note that I improved the test coverage of a couple of scripts in #877 in order to give me more confidence I hadn't broken them in this PR. Although I've also applied the refactorings to `SectionReslugger`, this class is not tested. However, since the script which uses this class is already broken (see #851), I think this is OK.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -2,6 +2,8 @@ require "list_manuals_service"
 require "show_manual_service"
 require "create_manual_service"
 require "update_manual_service"
+require "queue_publish_manual_service"
+require "publish_manual_worker"
 
 class ManualsController < ApplicationController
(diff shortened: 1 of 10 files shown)


floehopper:
@chrislo: Thanks.

floehopper:
Rebasing & force-pushing again now that CI builds seem to be working again.

PR #891, opened 2017-03-20
Remove params from PublishingAPIWithdrawer

chrisroos:
We weren't using the flexibility of being able to construct a `PublishingAPIWithdrawer` with different `publishing_api`s. Constructing it in the class makes this more explicit.


chrisroos:
Thanks, @floehopper. I've rebased and force pushed in preparation for merging.

PR #892, opened 2017-03-20
[WIP] Inline listeners in services

chrislo:
I'm looking at inlining the `listeners` that are injected into the various `Service` classes used in the controller actions. In all cases I've looked at so far, the listeners that are injected are always the same so it feels like this change helps to remove some indirection. 

It does make the unit testing (where it exists) slightly more verbose as we cannot inject a stub listener and assert that methods are called on it. On the other hand, this at least makes it obvious that calling a service results in calls being made to the API. 

What do you think @chrisroos @floehopper ?

### DIFF
--- a/app/controllers/sections_controller.rb
+++ b/app/controllers/sections_controller.rb
@@ -154,10 +154,6 @@ class SectionsController < ApplicationController
     service = RemoveSectionService.new(
       manual_repository,
       self,
-      listeners: [
-        PublishingApiDraftManualExporter.new,
-        PublishingApiDraftSectionDiscarder.new
-      ]
     )
     manual, section = service.call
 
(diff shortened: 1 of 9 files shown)


chrisroos:
This looks good to me, @chrislo.

The only thing I'm not completely sure about is the name of the methods you've introduced. Instead of:

* `call_publishing_api_draft_manual_exporter`
* `call_publishing_api_draft_section_discarder`
* `call_publishing_api_draft_section_exporter`

I wonder whether something like these might be a bit better?

* `export_manual`
* `discard_section`
* `export_section`

chrislo:
Thanks both. Those method names are much better, I'll continue with this refactor using those names.

floehopper:
I agree with what @chrisroos said - it's definitely an improvement. Although as you say the unit testing gets more verbose, it makes things more explicit and I'm hopeful that in time changes like this will allow us to see better abstractions which can then in turn make the unit testing simpler.

PR #895, opened 2017-03-20
Extract SectionFactory

chrisroos:
This allows me to remove the `DocumentFactoryRegistry#section_factory_factory` and, in my opinion, leaves the code in a more understandable state.


chrisroos:

Issue #896, opened 2017-03-21
Fix incompatibility between mongoid_rails_migrations and mongoid versions


floehopper:
It's currently impossible to use `ActiveRecord::ConnectionAdapters::SchemaStatements` methods, e.g. `#rename_table`, from within a migration, because `mongoid_rails_migrations` relies on a later version of `mongoid`. More specifically, `Mongoid::Migration#connection` relies on `Mongoid.default_session` rather than `Mongoid.database`.

Although this change isn't specifically mentioned in the changelog, it looks as if it was made in [this commit][1]. Running the following commands suggest that it the first release containing this commit was v3.0.0.

    $ git clone git@github.com:mongodb/mongoid.git
    $ cd mongoid
    $ git tag --contains b5040fc196d0cf4552b4608278a14c72e2f9656f | sort | head -n1
    v3.0.0

Note that it may be more appropriate to use a better alternative to the `mongoid_rails_migrations` gem rather than attempt to fix this.

[1]: https://github.com/mongodb/mongoid/commit/b5040fc196d0cf4552b4608278a14c72e2f9656f

floehopper:
It might be worth considering the [`mongoid_migration`](https://github.com/computadude/mongoid_migration) gem instead of `mongoid_rails_migrations`. Or even rolling our own.

floehopper:
I'm now less convinced that it was ever intended that you would be able to use `ActiveRecord::ConnectionAdapters::SchemaStatements` methods via `mongoid_rails_migrations`.

However, the specific incompatibility mentioned has been resolved by the recent upgrades to Mongoid in combination with the upgrade to `mongoid_rails_migrations` in #1154. More specifically, as [I mentioned in this commit](https://github.com/alphagov/manuals-publisher/commit/598f6692a0e805dbc0625a5bf4329a298c728d42), it's now possible to use `Mongoid::Migration.connection` from within a migration.

Closing.

Issue #893, opened 2017-03-20
Use separate Mongo database for this app (stop sharing with other apps)


floehopper:
Conversation from Slack govuk-developers channel:

@floehopper:
> It looks as if `contentapi`, `manuals_publisher`, `publisher`, `specialist_publisher` and `travel_advice_publisher` apps all share the same mongo *database*.
>
> Have I understood this correctly? And if so, is there any reason not to use separate databases for each app? Or do some of the apps "communicate" via the database?

@danielroseman:
> content-api and publisher (and panopticon, now dead) did indeed communicate via the database. But content-api is being killed.
>
> And travel-advice-publisher is being split out into its own db any day now
>
> specialist-publisher only stores User information in its database, although I did think it had its own. Everything else is stored directly in the publishing-api.

@floehopper:
> ok - that’s useful to know - it sounds as if (at least in the short term) it might make sense to use a separate database for `manuals_publisher` too

In general it seems like a bad idea to share a database with others applications which presumably have write access.

See also #839.

thomasleese:
The process I'm going to follow for [Travel Advice Publisher (TAP)][tap] is as follows.

1. Take TAP down (book a slot of downtime and alert users).
2. Take a dump of the database: `mongodump -d govuk_content_production --out dump`
3. Restore the dump to a new database: `mongorestore --db travel_advice_publisher_production dump/govuk_content_production `
4. Deploy [a puppet change][puppet] to change the database URL.
5. Take TAP back up and alert the users that it is available to use again.
6. Update [the development configuration][config] to match the production database name one day later (to ensure replication from integration is possible).
7. Delete unused collections from the new database that were copied from the old one.

[tap]: https://github.com/alphagov/travel-advice-publisher
[puppet]: https://github.com/alphagov/govuk-puppet/pull/5642
[config]: https://github.com/alphagov/travel-advice-publisher/pull/182

floehopper:
@thomasleese: Thanks!

PR #897, opened 2017-03-21
Rename SectionEdition's database collection from specialist_document_editions to manual_section_editions

### LINKED ISSUE #848: Rename the specialist_document_editions Mongo collection
PR #837 renames `SpecialistDocumentEdition` to `SectionEdition` but the data is still stored in the `specialist_document_editions` Mongo collection. 

We should rename the collection to `section_editions` once PR #837 has been merged.

floehopper:
* The new name more closely matches the model class name and removes one of the last vestiges of specialist documents in the app.

* I've prefixed the new collection name with "manual" rather than just using "section_editions", because it turns out that this app is sharing a database with a number of other apps, e.g. Publisher, and I felt as if "section_editions" was too generic and might lead to confusion.

* I've used methods on `Mongoid::DB` rather than `ActiveRecord::ConnectionAdapters::SchemaStatements#rename_table`, because of an incompatibility (see #896) between `mongoid_rails_migrations` and the version of `mongoid` we are using which results in the following exception:

```
NoMethodError: undefined method `default_session' for Mongoid:Module
```

* I've had to include code to drop the target collection if it exists and is empty before doing the rename, because it appears that something (presumably Mongoid) is creating the collection on application startup based on the existence of the `SectionEdition` class and its call to `store_in` with the new collection name. This means that an empty `manual_section_editions` collection is created before the migration is run and if we didn't drop the collection, the rename would fail. I'm only doing this if the collection is empty, on the off-chance that the collection really exists and contains some important data.

Closes #848.

### DIFF
--- a/app/models/section_edition.rb
+++ b/app/models/section_edition.rb
@@ -4,7 +4,7 @@ class SectionEdition
   include Mongoid::Document
   include Mongoid::Timestamps
 
-  store_in "specialist_document_editions"
+  store_in "manual_section_editions"
 
   field :document_id,          type: String
   field :version_number,       type: Integer, default: 1
--- a/db/migrate/20170320160403_rename_specialist_document_editions_collection_to_section_editions.rb
+++ b/db/migrate/20170320160403_rename_specialist_document_editions_collection_to_section_editions.rb
@@ -1,0 +1,21 @@
+class RenameSpecialistDocumentEditionsCollectionToSectionEditions < Mongoid::Migration
+  def self.up
+    rename_collection('specialist_document_editions', 'manual_section_editions')
+  end
+
+  def self.down
+    rename_collection('manual_section_editions', 'specialist_document_editions')
+  end
+
+  def self.rename_collection(source, target)
+    db = Mongoid.database
+    if db.collection_names.include?(target)
+      if db.collection(target).count == 0
+        db.drop_collection(target)
+      else
+        raise "Unexpected non-empty collection: #{target}"
+      end
+    end
+    db.rename_collection(source, target)
+  end
+end
--- a/spec/models/section_edition_spec.rb
+++ b/spec/models/section_edition_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe SectionEdition do
   subject { SectionEdition.new }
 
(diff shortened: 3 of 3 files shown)


chrislo:
@floehopper This looks good - I agree with the choice of collection name. It feels odd that this app shares a database with other apps but hopefully this will make it more obvious where this collection is managed from to anyone looking at that DB. Are you reasonably confident that none of the other apps rely on this collection being named `specialist_document_editions`? As-in other apps don't read or write from this collection?

floehopper:
@chrislo:

Thanks for reviewing.

> It feels odd that this app shares a database with other apps but hopefully this will make it more obvious where this collection is managed from to anyone looking at that DB.

I've created #893 to investigate using a separate DB.

> Are you reasonably confident that none of the other apps rely on this collection being named `specialist_document_editions`? As-in other apps don't read or write from this collection?

Good point. I'm fairly confident. Historically I think `govuk_content_api` would have used it via `govuk_content_models`. However, when Specialist Documents began to be published via the Publishing API, this will've become unnecessary. The following PRs are relevant:

* alphagov/govuk_content_models#172
* alphagov/govuk_content_models#324
* alphagov/govuk_content_api#229

So all in all, I'm happy to go ahead and merge this.

floehopper:
I've just noticed that alphagov/govuk_content_models#324 did not remove manuals and that there is still a [`RenderedManual` model](https://github.com/alphagov/govuk_content_models/blob/0607ac79d4955f39428e2cf170403b2b0e4710eb/app/models/rendered_manual.rb) in that repo. However, the fact that it's a `RenderedManual` and not a `Manual` and is presumably mapped to a `rendered_manuals` collection suggests that it is not relying on the `manual_records` collection provided by Manuals Publisher.

Looking at my copy of the production mongo db, it appears that both `rendered_specialist_documents` and `rendered_manuals` collections exist. My guess is that they are no longer used, but I'm going to ask in Slack to confirm.

floehopper:
As far as I can see, [manuals-frontend](https://github.com/alphagov/manuals-frontend) only uses the content store and not the content api and so I can't see how any other apps can be depending on the `specialist_document_editions` collection name. I plan to merge this PR in the morning.

PR #900, opened 2017-03-21
Inline listeners in services

chrislo:
In all the places I could find, when a Service was called from a Controller with a "listener" argument the value passed in did not vary from call site to call site. I've inlined these listener arguments into the constructor of each Service. 

In a couple of cases I've given the `notify_listeners` method a more meaningful name. In some others, where the listeners were fetched from the `ManualObserversRegistry` I haven't. I think instead I'd like to try and remove the `ManualObserversRegistry` and either inline the methods it calls into the Services themselves, or give it a better name.

### DIFF
--- a/app/services/remove_section_service.rb
+++ b/app/services/remove_section_service.rb
@@ -65,7 +65,7 @@ private
   end
 
   def export_draft_manual_to_publishing_api
-    PublishingApiDraftManualExporter.new.call(document, manual)
+    PublishingApiDraftManualExporter.new.call(manual)
   end
 
   class ManualNotFoundError < StandardError; end
--- a/app/services/remove_section_service.rb
+++ b/app/services/remove_section_service.rb
@@ -2,10 +2,6 @@ class RemoveSectionService
   def initialize(manual_repository, context)
     @manual_repository = manual_repository
     @context = context
-    @listeners = [
-      PublishingApiDraftManualExporter.new,
-      PublishingApiDraftSectionDiscarder.new
-    ]
   end
 
   def call
@@ -19,7 +15,8 @@ class RemoveSectionService
 
       remove
       persist
-      notify_listeners
+      export_draft_manual_to_publishing_api
+      discard_section_via_publishing_api
     end
 
     [manual, document]
@@ -27,7 +24,7 @@ class RemoveSectionService
 
 private
 
-  attr_reader :manual_repository, :context, :listeners
+  attr_reader :manual_repository, :context
 
   def remove
     manual.remove_document(document_id)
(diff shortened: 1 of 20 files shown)


spec/services/remove_section_service_spec.rb
```diff
@@ -70,9 +74,14 @@
       expect(manual).not_to have_received(:draft)
     end
 
-    it "does not notifies its listeners" do
+    it "does not export a manual" do
+      begin; service.call; rescue; end
```
floehopper: Do you know why we need this `rescue` block? It's a bit horrible! However, it looks as if you've just copied it from an existing spec. If so, it would be nice to explain that in the commit note and perhaps to create a new issue to investigate removing these `rescue` blocks.

app/services/remove_section_service.rb
```diff
@@ -63,10 +60,12 @@ def change_note_params
...
-  def notify_listeners
-    listeners.each do |listener|
-      listener.call(document, manual)
-    end
+  def discard_section
+    PublishingApiDraftSectionDiscarder.new.call(document, manual)
+  end
+
+  def export_manual
+    PublishingApiDraftManualExporter.new.call(document, manual)
```
floehopper: Minor: Keep the order of method definitions consistent, i.e. `export_manual` first.

floehopper: Also see earlier comment about method names.

app/services/republish_manual_service.rb
```diff
@@ -1,7 +1,7 @@
 class RepublishManualService
-  def initialize(published_listeners: [], draft_listeners: [], manual_id:)
-    @published_listeners = published_listeners
-    @draft_listeners = draft_listeners
+  def initialize(manual_id:)
+    @published_listeners = ManualObserversRegistry.new.republication
+    @draft_listeners = ManualObserversRegistry.new.update
```
floehopper: Minor: One difference is that we're now instantiating `ManualObserversRegistry` twice. I'm guessing there's no significant cost to doing this, but it might be worth noting in the commit note. Alternatively the instance could be extracted into a local variable in the constructor.

spec/services/remove_section_service_spec.rb
chrislo: I thought they looked odd too. I've raised #901 to investigate.

app/services/republish_manual_service.rb
chrislo: I've fixed this by assigning `ManualObserversRegistry.new` to a variable.

chrislo:
I've addressed the renaming of the methods (and fixing the typo with the method names) and force pushed. Thanks for the feedback @floehopper.

app/services/remove_section_service.rb
```diff
@@ -16,15 +15,16 @@ def call
 
       remove
       persist
-      notify_listeners
+      export_draft_manual_to_publishing_api
+      discard_section
```
floehopper: Maybe this should be something more like `discard_section_via_publishing_api`.

chrislo: I changed that method name to `discard_section_via_publishing_api`

chrislo:
Rebasing and force pushing.

PR #902, opened 2017-03-22
Use 'rummager' to find Rummager service

chrisroos:
As per the commit note for
https://github.com/alphagov/gds-api-adapters/commit/ddae94d1656f59f7778ca794a0cce356fc75e7e2.

This change was released in version 32.0.0[1] of gds-api-adapters and we're
currently using 39.2.0.

This allows me to use the GdsApi::TestHelpers::Rummager methods (e.g.
`stub_any_rummager_post`) in our tests.

[1]:
https://github.com/alphagov/gds-api-adapters/blob/master/CHANGELOG.md#3200

PR #903, opened 2017-03-22
Fix reslug_section script

chrisroos:
This branch aims to fix the `bin/reslug_section` script that had been neglected. It was prompted by issue #851.

I've run the script locally and observed that it now updates manual section slugs. I haven't looked too closely at the actual behaviour but am hoping that someone familiar with the script will be able to use it to determine whether it's still doing the right thing.

This branch includes:

* Fixes for a number of small problems in `SectionReslugger`.
* Updates to `PublishingAPIRedirector` to use v2 of the Publishing API and added a spec to help us avoid breaking it again in future.
* A test for `SectionReslugger` to help us avoid breaking it again in future.

### DIFF
--- a/lib/section_reslugger.rb
+++ b/lib/section_reslugger.rb
@@ -44,7 +44,7 @@ private
 
   def validate_current_section_in_content_store
     raise Error.new("Manual Section does not exist in content store") if current_section_in_content_store.nil?
-    raise Error.new("Manual Section already withdrawn") if current_section_in_content_store.format == "gone"
+    raise Error.new("Manual Section already withdrawn") if current_section_in_content_store['format'] == "gone"
   end
 
   def validate_new_section
--- a/lib/section_reslugger.rb
+++ b/lib/section_reslugger.rb
@@ -1,5 +1,4 @@
 require "gds_api/content_store"
-require "manual_service_registry"
 require "services"
 require "update_section_service"
 
--- a/lib/section_reslugger.rb
+++ b/lib/section_reslugger.rb
@@ -65,7 +65,7 @@ private
 
   def redirect_section
     PublishingAPIRedirecter.new(
-      publishing_api: Services.publishing_api,
+      publishing_api: Services.publishing_api_v2,
       entity: current_section_edition,
       redirect_to_location: "/#{full_new_section_slug}"
     ).call
--- a/lib/section_reslugger.rb
+++ b/lib/section_reslugger.rb
@@ -107,10 +107,8 @@ private
   end
 
   def publish_manual
-    observers = ManualObserversRegistry.new
     service = PublishManualService.new(
       manual_repository: RepositoryRegistry.new.manual_repository,
-      listeners: observers.publication,
       manual_id: manual_record.manual_id,
       version_number: manual_version_number,
     )
--- a/lib/section_reslugger.rb
+++ b/lib/section_reslugger.rb
@@ -60,6 +60,7 @@ private
   def validate_new_section_in_content_store
     section = section_in_content_store(full_new_section_slug)
     raise Error.new("Manual Section already exists in content store") if section
+  rescue GdsApi::ContentStore::ItemNotFound # rubocop:disable Lint/HandleExceptions
   end
 
   def redirect_section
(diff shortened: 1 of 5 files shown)


chrisroos:
The first commit in this branch is currently open as PR #902. I'll remove the commit from this branch once that PR's been merged.

lib/section_reslugger.rb
```diff
@@ -109,10 +109,8 @@ def change_note
   end
 
   def publish_manual
-    observers = ManualObserversRegistry.new
     service = PublishManualService.new(
       manual_repository: RepositoryRegistry.create.manual_repository,
-      listeners: observers.publication,
```
chrislo: Sorry @chrisroos I just noticed I'd missed this in 2fd96ea98! I also have a commit on my current working branch that fixes this, but I'll get rid of that.

lib/section_reslugger.rb
```diff
@@ -58,8 +58,11 @@ def validate_new_section_in_database
   end
 
   def validate_new_section_in_content_store
-    section = section_in_content_store(full_new_section_slug)
-    raise Error.new("Manual Section already exists in content store") if section
+    begin
+      section = section_in_content_store(full_new_section_slug)
+      raise Error.new("Manual Section already exists in content store") if section
+    rescue GdsApi::ContentStore::ItemNotFound # rubocop:disable Lint/HandleExceptions
+    end
```
floehopper: Minor: The `begin` & `end` statements are redundant within a method definition.

floehopper: Minor: The exception mentioned in the commit note doesn't match the rescued exception: `ItemNotFound` vs `NotFoundError`.

floehopper: It seems pretty nasty to use exception handling in this way. Is there no way to ask the Content Store whether the section exists and obtain a return value rather than an exception?

lib/services.rb
```diff
@@ -1,4 +1,5 @@
 require "gds_api/asset_manager"
+require "gds_api/content_store"
```
floehopper: I now realise that a `Services.content_store` method already makes the Content Store available. Thus I think my last comment isn't quite right. I think it would be clearer to amend the commit note to say that `lib/services.rb` should explicitly require all its dependencies. I think mentioning the Rails console is confusing.

chrisroos: I've updated the commit note to remove any reference of Rails console.

lib/section_reslugger.rb
chrisroos: I've removed the `begin` and `end` statements.

chrisroos: I've updated the commit message to correct the name of the exception.

chrisroos: > It seems pretty nasty to use exception handling in this way. Is there no way to ask the Content Store whether the section exists and obtain a return value rather than an exception?

I agree it's a bit nasty. Unfortunately, there doesn't appear to a way to ask the Content Store whether a piece of content exists. At least not via the GdsApi gem - https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/content_store.rb.

chrisroos:
> This looks good to me, but since I'm not familiar with the script I'm not going to approve the PR in the hope that you can find someone more familiar with it to check your changes.

Thanks for the feedback, @floehopper. My plan was to get this merged and then ask someone to use it in integration to see whether we can correctly rename the sections mentioned in the Zendesk ticket linked to from #851. Are you happy to approve it on that basis?

chrisroos:
> The first commit in this branch is currently open as PR #902. I'll remove the commit from this branch once that PR's been merged.

I've removed this commit now that #902 has been merged.

chrisroos:
The first commit in this branch is open as PR #910. I'll remove the commit from this branch once that PR's been merged.

floehopper:
@chrisroos:

> My plan was to get this merged and then ask someone to use it in integration to see whether we can correctly rename the sections mentioned in the Zendesk ticket linked to from #851. Are you happy to approve it on that basis?

Yes. I'll do it now.

chrisroos:

PR #904, opened 2017-03-22
Rename ManualWithDocuments

chrisroos:
This branch renames `ManualWithDocuments` and its methods to reflect the language of the domain.

Note that there are still lots of variables (in code and test) that refer to 'documents' instead of 'sections'. I'm intentionally not changing them in this branch as it's a big enough change as it is.

### DIFF
--- a/app/observers/manual_observers_registry.rb
+++ b/app/observers/manual_observers_registry.rb
@@ -69,7 +69,7 @@ private
         )
       end
 
-      manual.removed_documents.each do |doc|
+      manual.removed_sections.each do |doc|
         next if doc.withdrawn?
         next if doc.minor_update?
 
@@ -100,7 +100,7 @@ private
         )
       end
 
-      manual.removed_documents.each do |section|
+      manual.removed_sections.each do |section|
         indexer.delete(
           SectionIndexableFormatter.new(section, manual),
         )
@@ -146,7 +146,7 @@ private
         document.mark_as_exported! if action != :republish
       end
 
-      manual.removed_documents.each do |document|
+      manual.removed_sections.each do |document|
         next if document.withdrawn? && action != :republish
         begin
           publishing_api_v2.unpublish(document.id, type: "redirect", alternative_path: "/#{manual.slug}", discard_drafts: true)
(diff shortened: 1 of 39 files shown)


spec/models/manual_with_documents_spec.rb
```diff
@@ -131,13 +131,13 @@
     let(:document_c) { double(:document, id: "c") }
 
     it "removes the document from #documents" do
```
chrislo: Perhaps update the description of the test too.

spec/models/manual_with_documents_spec.rb
```diff
@@ -131,13 +131,13 @@
...
-      manual_with_documents.remove_document(document_a.id)
+      manual_with_documents.remove_section(document_a.id)
 
       expect(manual_with_documents.documents.to_a).to eq([document_b])
     end
 
     it "adds the document to #removed_documents" do
```
chrislo: This description could also be updated

app/observers/manual_observers_registry.rb
```diff
@@ -69,7 +69,7 @@ def publication_logger
         )
       end
 
-      manual.removed_documents.each do |doc|
+      manual.removed_sections.each do |doc|
```
chrislo: I might be tempted to use `|section|` rather than `|doc|` here

app/observers/manual_observers_registry.rb
```diff
@@ -146,7 +146,7 @@ def publishing_api_publisher
         document.mark_as_exported! if action != :republish
       end
 
-      manual.removed_documents.each do |document|
+      manual.removed_sections.each do |document|
```
chrislo: and here

app/repositories/marshallers/document_association_marshaller.rb
```diff
@@ -29,12 +29,12 @@ def dump(manual, record)
       document_repository.store(document)
     end
 
-    manual.removed_documents.each do |document|
+    manual.removed_sections.each do |document|
```
chrislo: and here

chrisroos:
Thanks @chrislo! I'm going to hold off renaming other instances of `document` and `doc` in this branch as I'm confident we'll get to them soon enough.

PR #905, opened 2017-03-22
Add missing require to section_factory.rb

chrisroos:
This avoids the following error when trying to view a Manual in
development:

    uninitialized constant SectionFactory::ChangeNoteValidator

This is similar to the problem we saw in
01e4ab5af2fdb2a3811a4ea02fff9a2ceb1b93e7. As noted in that commit it
doesn't seem to be possible to write a test to protect us from these
sort of problems.

### DIFF
--- a/app/lib/section_factory.rb
+++ b/app/lib/section_factory.rb
@@ -1,3 +1,6 @@
+require 'validators/change_note_validator'
+require 'validators/section_validator'
+
 class SectionFactory
   def initialize(manual)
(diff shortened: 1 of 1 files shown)


chrisroos:
Thanks @floehopper!

chrisroos:
> I wonder if it would be safer to make the validators auto-loadable...?

I've added issue #907 to capture this idea, @floehopper.

PR #906, opened 2017-03-22
Remove manuals observers registry

chrislo:
This PR removes the `ManualObserversRegistry` and inlines the functionality into the Service classes. This allows us to remove all of the references to "listeners" and instead use more imperative method names in the Service classes. 

I've struggled a little bit to come up with names for some of these methods and classes that I have extracted and I think this points to the fact that their functionality is duplicated in part elsewhere in the application. In particular: 

- NewPublishingApiPublisher <-> PublishingAPIPublisher
- NewPublishingApiWithdrawer <-> PublishingAPIWithdrawer
- PublishingApiDraftExporter <-> PublishingApiDraftManualExporter <-> PublishingApiDraftSectionExporter

If we're happy with these changes, I might raise issues to capture those three areas as places where we might be able to simplify the code.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -6,7 +6,6 @@ require "queue_publish_manual_service"
 require "publish_manual_worker"
 require "preview_manual_service"
 require "builders/manual_builder"
-require "manual_observers_registry"
 
 class ManualsController < ApplicationController
   before_filter :authorize_user_for_publishing, only: [:publish]
--- a/app/exporters/publishing_api_draft_manual_with_sections_exporter.rb
+++ b/app/exporters/publishing_api_draft_manual_with_sections_exporter.rb
@@ -1,0 +1,31 @@
+class PublishingApiDraftManualWithSectionsExporter
+  def call(manual, action = nil)
+    update_type = (action == :republish ? "republish" : nil)
+
+    organisation = organisation(manual.attributes.fetch(:organisation_slug))
+
+    ManualPublishingAPILinksExporter.new(
+      organisation, manual
+    ).call
+
+    ManualPublishingAPIExporter.new(
+      organisation, manual, update_type: update_type
+    ).call
+
+    manual.sections.each do |document|
+      next if !document.needs_exporting? && action != :republish
+
+      SectionPublishingAPILinksExporter.new(
+        organisation, manual, document
+      ).call
+
+      SectionPublishingAPIExporter.new(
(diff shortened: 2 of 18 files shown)


chrislo:
Github is showing these commits in a slightly different order to how I have them - in the branch they are grouped 'Extract -> Inline -> Rename'

chrisroos:
I'm still going through this branch but just wanted to add a note to apologise for making your life  harder by merging PR #904. I can see that you've extracted classes containing calls to `ManualWithDocuments#documents` which will need to be changed to `ManualWithSections#sections`. Sorry and I hope it's not too painful!

chrisroos:
The extracted classes all seem to operate on a ManualWithSections. I wonder whether we can use that to come up with more specific (albeit more verbose) names?

* `PublishingApiDraftExporter` -> `PublishingApiDraftManualWithSectionsExporter`
* `NewPublishingApiPublisher` -> `PublishingApiManualWithSectionsPublisher`
* `RummagerExporter` -> `RummagerManualWithSectionsExporter`
* `RummagerWithdrawer` -> `RummagerManualWithSectionsWithdrawer`
* `NewPublishingApiWithdrawer` -> `PublishingApiManualWithSectionsWithdrawer`

This might help clarify the differences between, for example:

* `PublishingApiDraftManualsWithSectionExporter`
* `PublishingApiDraftManualExporter`
* `PublishingApiDraftSectionExporter`

I also wonder whether these extracted classes should also live in app/exporters given that there are already other *_exporter and *_withdrawer classes in there?

chrisroos:
Replacing the comments with assertions in "Inline ManualObserversRegistry#publication" is a great improvement :-)

chrisroos:
In "Inline ManualObserversRegistry#update" it might be worth updating the commit note to mention that you'll handle `#update_original_publication_date` in a separate commit. It wasn't immediately obvious to me why the alias had been removed and the method name changed.

chrisroos:
Despite my comments I think this is a definite improvement. It makes it much easier to see what the various *_manual_service classes actually do. Good work!

chrisroos:
The only thing preventing me from accepting the PR at the moment is the naming of some of the extracted classes. What do you think about either the names I've suggested or an alternative that slightly better describes their purpose, @chrislo?

chrislo:
Thanks @chrisroos - I've made the class name and file name changes you suggested, and added some more details in the commit message for ffee3735. I've rebased the changes back in the history and force pushed.

chrislo:
Force pushing in preparation for merging.

PR #908, opened 2017-03-22
Improvements to RemoveSectionService spec

### LINKED ISSUE #901: Remove `begin; service.call; rescue; end` from `remove_section_service_spec.rb`
Arises from: https://github.com/alphagov/manuals-publisher/pull/900/files/92040afe8491a41d9dab4f32177b2f9c6d074551..b3cae20cb3f9dde4c65901f9f0fba0c1efc4a5b8#diff-de762617bb0811292f68975b2a3674da

It's not clear why this rescue is required and it may hide an spec failure when we're refactoring, so it would be good to remove it.

floehopper:
* Only ignore a specific exception class
* Make it more explicitly that we are deliberately ignoring the exception
* Reduce duplication

Fixes #901

### DIFF
--- a/spec/services/remove_section_service_spec.rb
+++ b/spec/services/remove_section_service_spec.rb
@@ -63,25 +63,33 @@ RSpec.describe RemoveSectionService do
       }
     end
 
-    it "raises a SectionNotFound error" do
+    it "raises a SectionNotFoundError" do
       expect {
         service.call
       }.to raise_error(described_class::SectionNotFoundError, document_id)
     end
 
-    it "does not mark the manual as a draft" do
-      begin; service.call; rescue; end
-      expect(manual).not_to have_received(:draft)
-    end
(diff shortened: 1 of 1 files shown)


spec/services/remove_section_service_spec.rb
```diff
@@ -69,19 +69,27 @@
       }.to raise_error(described_class::SectionNotFoundError, document_id)
     end
 
-    it "does not mark the manual as a draft" do
-      begin; service.call; rescue; end
-      expect(manual).not_to have_received(:draft)
-    end
+    context "when SectionNotFoundError is ignored" do
```
chrisroos: I wonder whether this would be better named as "when SectionNotFoundError is raised"?

floehopper: Agreed. I'll make this change and amend the commit.

floehopper:
@chrisroos: Thanks.

Issue #907, opened 2017-03-22
Auto load validators and builders


chrisroos:
We've had two problems caused by missing `require` statements. One for a builder (01e4ab5af2fdb2a3811a4ea02fff9a2ceb1b93e7) and another for a validator (eea94ce9fa7a6af50ef7007ff302b3ed986c567a).

I _think_ that everything in app/*.rb is auto-required/loaded so it might make sense to extend that to app/**/*.rb so that it includes builders and validators too.

PR #909, opened 2017-03-22
Add missing require statements

chrisroos:
I needed to add these in order to be able to navigate to /manuals/new in
development.

They've been needed since 30a72a16dadde70c86a669ae23f715f20554b7fd.
Prior to that commit, constructing a `RepositoryRegistry` (which happens
indirectly in `ManualsController#index`) would result in a
`DocumentFactoryRegistry` being created which in turn would result in
the necessary builders and validators being required.

As of this commit I appear to be able to do everything through the user
interface as expected.

A longer term fix might be to autoload/require the builders and
validators as suggested in issue #907.

### DIFF
--- a/app/models/builders/manual_builder.rb
+++ b/app/models/builders/manual_builder.rb
@@ -1,4 +1,6 @@
 require "securerandom"
+require "validators/manual_validator"
+require "validators/null_validator"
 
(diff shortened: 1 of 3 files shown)


chrisroos:
Thanks @floehopper!

Issue #855, opened 2017-03-13
Check or remove delete_documents_marked_for_delete, delete_duplicate_drafts & find_duplicate_documents scripts


floehopper:
We should decide whether or not it's worth keeping these scripts. If we do decide to keep them, we'll need to check they work and ensure that they remain working by adding the appropriate test coverage.

From [this comment][1] by @h-lame:
> `delete_documents_marked_for_delete`, `delete_duplicate_drafts` and `find_duplicate_documents` are to help working around the fact that sometimes normal use of the app can result in duplicate manuals or sections being created.  `delete_documents_marked_for_delete` is used when the users flag the duplicates so they are easy to remove (prefix the title with an xx I think), and the others try to find things programmatically.  Getting to the bottom of the actual problem should mean we can get rid of all three of these (particularly if we also provide delete/withdraw functionality). See https://trello.com/c/R3kQxY9Z/74-duplication-bug-in-manuals for some thoughts on what the cause might be

[1]: https://github.com/alphagov/manuals-publisher/pull/832#issuecomment-285417474

chrisroos:
I've opened two PRs to improve and add test coverage to the `delete_documents_marked_for_delete` (PR #913) and `find_duplicate_documents` (PR #915) scripts.

The `delete_duplicate_drafts` already appears to work and has test coverage.

Given these scripts appear to be working around functionality that's missing from the UI I'm happy for them to remain. I believe that if/when the two PRs are merged we can close this issue.

floehopper:
Given that #913 and #915 have been merged, @chrislo & I think we can close this. Closing.

PR #913, opened 2017-03-23
Improve delete_documents_marked_for_delete script

chrisroos:
This addresses one of the three scripts mentioned in issue #855.

This branch:

* Renames the script from `delete_documents_marked_for_delete` to `delete_sections_marked_for_delete`.
* Extracts the `MarkedSectionDeleter` class.
* Adds some basic tests for the `MarkedSectionDeleter` class.

There are definite improvements that can be made to the code in `MarkedSectionDeleter` but I think this is enough of an improvement for now.


lib/marked_section_deleter.rb
```diff
@@ -0,0 +1,60 @@
+require "services"
+
+class MarkedSectionDeleter
+  def initialize(logger = STDOUT)
```
floehopper: Minor: It feels slightly odd/confusing calling this a "logger" when it's actually just an output stream.

chrisroos: Hmm, good point. How about I rename it to `io`? Unless you have any other suggestions?

chrisroos: I've renamed the param to `io`.

chrisroos:

PR #914, opened 2017-03-23
Rename document to section (part 1)

floehopper:
Lots of the code is still referring to documents when it means sections. This is a remnant of when the app supported Specialist Documents as well as Manuals and their Sections.

This PR does the rename across a chunk (but not all) of the codebase. I decided it was worth getting these changes merged as soon as possible before doing any more to avoid unnecessary merge conflicts with other PRs.

Note that I've intentionally avoided renaming anything to do with the `SectionEdition#document_id` and `ManualRecord#document_ids` fields which are in are the Mongo database.


floehopper:
@chrisroos: Thanks.

PR #915, opened 2017-03-23
Extract DuplicateDocumentFinder

chrisroos:
This addresses one of the three scripts mentioned in issue #855.

And add a spec to test the behaviour.

The spec isn't ideal but I think it's probably good enough to give us
some confidence that we won't break this script while we continue to
refactor the app.

### DIFF
--- a/bin/find_duplicate_documents
+++ b/bin/find_duplicate_documents
@@ -1,18 +1,6 @@
 #!/usr/bin/env ruby
 
 require File.expand_path("../../config/environment", __FILE__)
+require 'duplicate_document_finder'
 
-slug_hash = {}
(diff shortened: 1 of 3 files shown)


chrisroos:
> In adding this test we are implicitly saying that this script is worth keeping at least for the moment. I think it might be worth making that decision explicit, perhaps as a comment on #855.

Did you see my comment below on #855, @floehopper? Does this cover what you mean?

> Given these scripts appear to be working around functionality that's missing from the UI I'm happy for them to remain. I believe that if/when the two PRs are merged we can close this issue.

chrisroos:
Thanks for reviewing, @floehopper.

PR #918, opened 2017-03-24
Improve consistency of ordering in Mongoid queries

chrisroos:
These discrepancies were highlighted when I upgraded this app to Mongoid 3. I've split these changes into their own PR as I think they're useful in their own right, and merging them will reduce the size of the branch that upgrades Mongoid.

### DIFF
--- a/lib/attachment_reporting.rb
+++ b/lib/attachment_reporting.rb
@@ -24,7 +24,7 @@ class AttachmentReporting
       # the editions of these sections in version order to find unique PDF attachments and their
       # publication times.
       all_unique_document_ids_for_manual(manual).each do |document_id|
-        section_editions = SectionEdition.where(document_id: document_id).order(:version_number)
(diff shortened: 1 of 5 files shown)


lib/attachment_reporting.rb
```diff
@@ -24,7 +24,7 @@ def create_organisation_attachment_count_hash
       # the editions of these sections in version order to find unique PDF attachments and their
       # publication times.
       all_unique_document_ids_for_manual(manual).each do |document_id|
-        section_editions = SectionEdition.where(document_id: document_id).order(:version_number)
+        section_editions = SectionEdition.where(document_id: document_id).order([:version_number, :asc])
```
floehopper: The commit message implies that the correct syntax is `order(version_number: :asc)`, but here it has been changed to `order([:version_number, :asc])`. I'm not sure which is correct.

chrisroos: Good spot. I'll update the commit note.

chrisroos:

PR #919, opened 2017-03-27
Rename document to section (part 2)

floehopper:
This follows on from #914.

I think this changes most of the remaining occurrences of "document" to "section". The main exceptions are:

*  `SectionEdition#document_id`, `ManualRecord#document_ids` & `ManualRecord#removed_document_ids` fields.
* A few places where an object can be either a `Manual` **or** a `Section`.
* HTML element IDs / CSS classes & selectors.

I'm sorry that there are a lot of commits, but they are mostly pretty small and the tests should pass after every commit.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -70,9 +70,9 @@ module ApplicationHelper
     end
   end
 
-  def preview_path_for_section(manual, document)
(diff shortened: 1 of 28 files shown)


floehopper:
@chrislo: Many thanks. I think I'll get it merged then!

PR #923, opened 2017-03-28
Remove DocumentFactoryRegistry

floehopper:
This is one of the last references to the word "document" and it doesn't feel as if the abstraction it offers is very useful or intention-revealing.

### DIFF
--- a/app/repositories/repository_registry.rb
+++ b/app/repositories/repository_registry.rb
@@ -7,6 +7,7 @@ require "manual_publish_task"
 require "manual_with_publish_tasks"
 require "manual"
 require "manual_record"
+require "manual_with_sections"
 
 class RepositoryRegistry
   def organisation_scoped_manual_repository_factory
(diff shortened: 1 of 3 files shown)


app/repositories/repository_registry.rb
```diff
@@ -27,7 +28,15 @@ def scoped_manual_repository(collection)
         SectionAssociationMarshaller.new(
           section_repository_factory: section_repository_factory,
           decorator: ->(manual, attrs) {
-            DocumentFactoryRegistry.new.manual_with_sections.call(manual, attrs)
+            ManualValidator.new(
```
chrislo: Could you go a step further here and inline this decorator argument into the `SectionAssociationMarshaller`? I don't think the marshaller is ever called with a different value for this decorator.

floehopper: I probably could, but my goal was simply to remove one of the last classes/modules including the word "Document" and these changes achieve that. I think I'd prefer to limit the scope of this PR to just that.

floehopper:
@chrislo:

> I think this looks good - I'm in the process of removing the RepositoryRegistry altogether, but I think I should be able to cope with rebasing my changes on top of these.

I'm sorry about that. This was something I had hanging over in my to-do list for getting rid of occurrences of the word "document".

chrislo:
@floehopper no problem!

floehopper:
Force-pushing to encourage Jenkins to pick up this branch which was pushed earlier this morning before the problems with Jenkins.

PR #925, opened 2017-03-28
Remove DocumentHelpers

floehopper:
This is another of the remaining classes/modules including the word "document" in its name.

There are already methods in `ManualHelpers` which relate to both manuals and sections, so it doesn't seem worth keeping `DocumentHelpers` around given how few useful methods it contains.

My preference would be to make this change and then to extract the relevant methods out of `ManualHelpers` into a new `SectionHelpers` module if appropriate.


floehopper:
Force-pushing to encourage Jenkins to pick up this branch which was pushed earlier this morning before the problems with Jenkins.

floehopper:

floehopper:
@chrisroos: Thanks.

PR #926, opened 2017-03-28
Remove repository registry

chrislo:
This PR removes the `RepositoryRegistry` class, hopefully removing one layer of indirection when other code requires reference to a repository. 

In doing so I've introduced another Repository called `ScopedManualRepository` after the method name it was extracted from. Long-term I don't think we'll want to have repositories at all, but this refactor does reduce the indirection in RepositoryRegistry. 

I'm not sure about the name `ScopedManualRepository` - looking at it, I wonder if it should be `ManualWithSectionsRepository` to reflect some of the other changes we have made?


chrislo:
Thanks @chrisroos! 

Rebasing and force pushing before merging.

PR #927, opened 2017-03-28
Remove unused options from mongoid.yml

chrisroos:
As far as I can tell these options have no effect. They're not options in Mongoid and don't appear to be options in the underlying mongo gem either.

These options were added in c371b95 but the commit note doesn't explain anything about them. In the comments on that commit, the original author has explained that the setting was copied from alphagov-deployment and that there's no explanation in that repo either.

Removing them should make it easier to upgrade Mongoid in a separate branch.

### DIFF
--- a/config/mongoid.yml
+++ b/config/mongoid.yml
@@ -14,7 +14,6 @@ production:
   database: govuk_content_production
   refresh_mode: :sync
   logger: false
-  refresh_interval: 120
   use_activesupport_time_zone: true
   hosts:
     - <%= ENV['MONGODB_NODES'] %>
--- a/config/mongoid.yml
+++ b/config/mongoid.yml
@@ -12,7 +12,6 @@ test:
 
 production:
   database: govuk_content_production
-  refresh_mode: :sync
   logger: false
   use_activesupport_time_zone: true
   hosts:


config/mongoid.yml
```diff
@@ -14,7 +14,6 @@ production:
   database: govuk_content_production
   refresh_mode: :sync
   logger: false
-  refresh_interval: 120
```
floehopper: I think this might in fact be "used" at least in some sense:

* [The production config sets a value against the `hosts` key](https://github.com/alphagov/manuals-publisher/blob/7decf07c3452a24ab07850130a352bba72aea6bd/config/mongoid.yml#L19-L20)
* This causes [`Mongoid::Config#configure_databases`](https://github.com/mongodb/mongoid/blob/v2.5.2/lib/mongoid/config.rb#L286-L292) to instantiate a `Mongoid::Config::ReplsetDatabase` and call `#configure` on it
* This in turn [instantiates a `Mongo::ReplSetConnection`](https://github.com/mongodb/mongoid/blob/v2.5.2/lib/mongoid/config/replset_database.rb#L22)
* The [`Mongo::ReplSetConnection`](https://github.com/mongodb/mongo-ruby-driver/blob/1.12.1/lib/mongo/legacy.rb#L59-L70)
* The latter is a sub-class of [`Mongo::MongoReplicaSetClient`](https://github.com/mongodb/mongo-ruby-driver/blob/1.12.1/lib/mongo/mongo_replica_set_client.rb) which does accept a [`refresh_interval` option](https://github.com/mongodb/mongo-ruby-driver/blob/1.12.1/lib/mongo/mongo_replica_set_client.rb#L105-L106).

However, maybe the production config for Mongoid in the repo isn't actually used in production...?

Also note that the option was renamed to `heartbeat_frequency` in [this commit](https://github.com/mongodb/mongo-ruby-driver/commit/f81a2b566516a2ea81596aed86b981e2dc025603).

config/mongoid.yml
```diff
@@ -12,7 +12,6 @@ test:
 
 production:
   database: govuk_content_production
-  refresh_mode: :sync
```
floehopper: This appears to be [another option associated with `Mongo::MongoReplicaSetClient`](https://github.com/mongodb/mongo-ruby-driver/blob/1.12.1/lib/mongo/mongo_replica_set_client.rb#L101-L104) as per my comment on the previous commit.

config/mongoid.yml
chrisroos: > However, maybe the production config for Mongoid in the repo isn't actually used in production...?

It is indeed used, @floehopper. Thanks for investigating so thoroughly. I'll come back to this later.

chrisroos:
Thanks for the thorough investigation, @floehopper. I'm going to close this PR and deal with these options as part of the upgrade to Mongoid 3.

PR #928, opened 2017-03-28
Upgrade to Mongoid 2.8.1

chrisroos:
I relaxed the constraint on Mongoid so that it can go up to the latest
version in the 2.x line.

I ran `bundle update mongoid` to generate the updated Gemfile.lock. All
the tests pass.

I create a migration (`rails g migration upgrading-mongoid`) and used it
to print the number of items in the manual_records collection (`p
ManualRecord.count`). This gives me confidence that the
mongoid_rails_migrations gem still works as expected after this upgrade.

Upgrading to the latest 2.x release might make it easier to upgrade to
3.x in future.


chrisroos:

PR #929, opened 2017-03-28
Upgrade Mongoid to 3

chrisroos:
This branch upgrades Mongoid from 2.8.1 to 3.1.7. It's another small step on the path toward upgrading to the latest version of Mongoid.

I upgrade Mongoid in an early commit and then fix the problems that this upgrade causes in the subsequent commits. This means that the app is essentially broken until the final commit in this branch. While not ideal, I think this is preferable to having the upgrade and all the fixes in a single commit.

The mongoid_rails_migrations gem appears to continue to work so I've not upgraded that as part of this PR.

A number of the existing migrations in the project fail to run after this upgrade. They're calling `Collection#update` which no longer exists. While I'm sure we could fix them I think my preference would be to delete them as we have with the other migrations that have already run on production.


chrisroos:
I've created issue #930 to investigate switching to `MONGODB_URI`.

chrisroos:
> * Work out what to do about the `MONGODB_NODES` in the [Puppet recipe for manuals_publisher](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/manuals_publisher.pp). 
>   * The nodes are stored in a string that's inserted into the mongoid.yml config file. This string contains significant whitespace and won't work with the updated mongoid.yml file that we need for 3.x.
>   * Option 1: Update the string in manuals_publisher.pp so that it contains the correct amount of whitespace.
>   * Option 2: Update mongoid.yml so that it handles different amounts of whitespace (e.g. iterate over the lines in `MONGODB_NODES` instead of inserting the string directly.
>   * Option 3: Try to switch to using `MONGODB_URI` as per the discussion in https://github.com/alphagov/govuk-puppet/commit/0d03bfd5c6e5e875906c9b1a9f44969f6dd2c330.

I've opened PR #931 as a solution to this problem.

chrisroos:
I plan to pick this up again next week. @floehopper / @chrislo: Feel free to pick it up in my absence if you're interested.

chrisroos:
I _think_ this is now ready for review. @chrislo/@floehopper: Can one of you take a look when you have a moment, please?

app/models/section_edition.rb
```diff
@@ -39,9 +39,9 @@ class SectionEdition
 
   scope :with_slug_prefix, ->(slug) { where(slug: /^#{slug}.*/) }
 
-  index "section_id"
-  index "state"
-  index "updated_at"
+  index section_id: 1
+  index state: 1
+  index updated_at: 1
```
floehopper: Is it worth extracting these magic numbers into an intention-revealing named constant? That is assuming we know what the magic number means!

app/models/manual_record.rb
```diff
@@ -12,7 +12,7 @@ class ManualRecord
     autosave: true
 
   def self.find_by(attributes)
-    first(conditions: attributes)
+    where(attributes).first
```
floehopper: If I'm understanding the documentation correctly, I think we might be able to remove this method entirely, because it's [implemented in Mongoid](http://www.rubydoc.info/gems/mongoid/3.0.0/Mongoid/Finders#find_by-instance_method).

app/models/section_edition.rb
chrisroos: I'm fairly happy to leave these magic numbers as they are. The Mongoid and MongoDB references about indexes talk about the index direction in terms of numbers (1 = ascending / -1 = descending). I've also explained in the commit note that these are set to 1 as that matches what was happening previously.

config/mongoid.yml
```diff
@@ -81,3 +81,13 @@ test:
...
+
+production:
+  sessions:
+    default:
+      uri: <%= ENV['MONGODB_URI'] %>
+      options:
+        logger: false
+        use_activesupport_time_zone: true
+        refresh_interval: 120
+        refresh_mode: :sync
```
chrisroos: I've added issue #958 to prompt us to investigate removing this configuration option. I think I prefer that to adding a comment to the config.

app/models/manual_record.rb
chrisroos: Great! I'm going to assume that we'll notice this again in future so am not going to do anything about it at the moment.

chrisroos:
> Do you think there's any need to explicitly create any extra indexes now that they are not being auto-created?

My guess is that we'll need to do this for any new indexes we add but that any indexes previously created will remain. I'm not planning to investigate this any further right now.

chrisroos:
> I found the commits where you generated an example configuration and then gradually modified it a bit confusing. I wondered whether it would be better to generate an example config and rename it to something else, e.g. config/mongoid_example.yml. And then gradually copy bits of it into the existing config...?

I agree that those commits are confusing. I'm not going to redo them now but will consider your suggested approach if I have to do something similar in future.

chrisroos:
Thanks for the review, @floehopper. I'm going to get this merged now!

Issue #930, opened 2017-03-28
Use MONGODB_URI instead of MONGODB_NODES in mongoid config


chrisroos:
It looks as though the standard way of defining multiple MongoDB hosts is to use the `MONGODB_URI` environment variable ([see other alphagov apps using this variable](https://github.com/search?q=org%3Aalphagov+mongodb_uri&type=Code)). Manuals Publisher uses a custom `MONGODB_NODES` environment variable instead because Ruby 2.1 couldn't parse multiple hosts from the `MONGODB_URI`. See the commit note and comments in https://github.com/alphagov/govuk-puppet/commit/0d03bfd5c6e5e875906c9b1a9f44969f6dd2c330 for more information.

Manuals Publisher is now using Ruby 2.2 (see a0aa4406d156ac1f28c9bc4a1f026f28d262c8c4) so it should hopefully be possible to remove `MONGODB_NODES` and switch to using `MONGODB_URI` instead.

floehopper:
I've opened a [PR on govuk-puppet to remove the `MONGODB_NODES` env var](https://github.com/alphagov/govuk-puppet/pull/5698).

PR #931, opened 2017-03-28
Iterate over MONGODB_NODES in mongoid.yml

chrisroos:
This reduces the coupling between this app and the [manuals_publisher
Puppet recipe in govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk/manifests/apps/manuals_publisher.pp).

The `MONGODB_NODES` variable contains a string like this:

```
- node1:27017
      - node2:27017
      - node3:27017
```

The whitespace is important as it allows this string to be inserted into
mongoid.yml and for the result to be valid (see commit https://github.com/alphagov/govuk-puppet/commit/06acc5ce08e11bee1d05d11e47c6afacb073a7f0 for
more info). The mongoid.yml file ends up containing something like:

```
production:
  hosts:
    - - node1:27017
      - node2:27017
      - node3:27017
```

I'm in the process of upgrading to Mongoid 3 which requires a change to
the format of mongoid.yml. Significantly the indentation of the hosts
setting changes which means that inserting `MONGODB_NODES` no longer
results in valid YAML.

The update in this commit treats `MONGODB_NODES` as an array and strips
the leading whitespace from each host so that we don't have to update
the Puppet recipe if/when this file changes.

I tested the change locally by setting `MONGODB_NODES` to the example
above and adding some debug output to `Mongoid::Config.load!` in the
mongoid gem. I confirmed that I get the `settings` hash yielded in the
`load!` method before and after this change.

I think the proper solution is to switch to using `MONGODB_URI` (as
documented in issue #930) but I think this is a good enough to allow me
to proceed with the Mongoid upgrade for now.

### DIFF
--- a/config/mongoid.yml
+++ b/config/mongoid.yml
@@ -10,6 +10,9 @@ test:
   database: govuk_content_manuals_publisher_test
   autocreate_indexes: true
 
+<% mongo_nodes = (ENV['MONGODB_NODES'] || '').split("\n") %>
+<% first_mongo_node = mongo_nodes.shift %>
+
 production:
   database: govuk_content_production
   refresh_mode: :sync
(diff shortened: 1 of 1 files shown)


floehopper:
I think this has been made redundant by the change to using `MONGODB_URI` in #936. Closing.

PR #932, opened 2017-03-28
Simplify scoped manual repository

chrislo:
This PR inlines the arguments of `SectionAssociationMarshaller` and `ManualPublishTaskAssociationMarshaller` making the code a bit easier to follow. It's still not really clear to me why these marshallers exist but hopefully reducing some of the indirection will eventually let us work out where their behaviour really belongs.

### DIFF
--- a/spec/repositories/marshallers/section_association_marshaller_spec.rb
+++ b/spec/repositories/marshallers/section_association_marshaller_spec.rb
@@ -15,7 +15,7 @@ describe SectionAssociationMarshaller do
     )
   }
 
-  let(:entity) { double(:entity) }
+  let(:manual) { double(:manual) }
   let(:record) {
     double(
       :record,
@@ -37,7 +37,7 @@ describe SectionAssociationMarshaller do
   let(:removed_sections) { [removed_section] }
 
   before do
-    allow(SectionRepository).to receive(:new).with(manual: entity).and_return(section_repository)
+    allow(SectionRepository).to receive(:new).with(manual: manual).and_return(section_repository)
   end
 
   describe "#load" do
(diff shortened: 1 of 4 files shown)


spec/repositories/marshallers/section_association_marshaller_spec.rb
```diff
@@ -64,16 +57,22 @@
...
+      allow(NullValidator).to receive(:new)
+      allow(ManualValidator).to receive(:new)
+
       marshaller.load(entity, record)
 
-      expect(decorator).to have_received(:call).
-        with(entity, sections: sections, removed_sections: removed_sections)
+      expect(ManualWithSections).to have_received(:new).with(:section_builder, entity, sections: [section], removed_sections: [removed_section])
+      expect(NullValidator).to have_received(:new)
+      expect(ManualValidator).to have_received(:new)
```
floehopper: I'm not convinced this is the best way to test this - I find it very difficult to follow.

Although it might introduce some duplication, I think it might be worth experimenting with testing the emergent behaviour of the chain of delegators, rather than attempting any mocking at all. Let me know if this doesn't make sense and I can try to explain in Slack or in person tomorrow.

floehopper: Also I might be missing something, but I'm not sure the `NullValidator` does anything useful at all, does it?

chrislo:
Thanks @floehopper! I've addressed your feedback (and the discussion we had in person) in 	aa8e2ad.

chrislo:
I accidentally merged aa8e2ad which I had intended to fixup into 02c2361. I don't think it's a big enough problem to warrent rewriting `master`, but apologies!

PR #933, opened 2017-03-29
Rename document_id to section_id

floehopper:
This PR continues the renaming theme from #914 & #919.

The first 4 commits fix some things that were missed in those PRs.

The next 3 commits alias the relevant Mongoid fields to the new name (using the `as` option) which has the effect of renaming the accessor-like methods. Calls to those methods are renamed within the same commit.

The last 3 commits remove the field aliases and rename the underlying fields in Mongo for all the existing records. Use of the underlying field names (e.g. in queries) are renamed within the same commit.

The tests all pass after each commit.

### DIFF
--- a/app/models/manual_record.rb
+++ b/app/models/manual_record.rb
@@ -76,7 +76,7 @@ private
     field :state, type: String
     field :version_number, type: Integer
     field :document_ids, type: Array, as: :section_ids
-    field :removed_document_ids, type: Array, as: :removed_section_ids
+    field :removed_section_ids, type: Array
     field :originally_published_at, type: DateTime
     field :use_originally_published_at_for_public_timestamp, type: Boolean
 
--- a/app/models/manual_record.rb
+++ b/app/models/manual_record.rb
@@ -75,7 +75,7 @@ private
     field :body, type: String
     field :state, type: String
     field :version_number, type: Integer
-    field :document_ids, type: Array, as: :section_ids
+    field :section_ids, type: Array
     field :removed_section_ids, type: Array
     field :originally_published_at, type: DateTime
     field :use_originally_published_at_for_public_timestamp, type: Boolean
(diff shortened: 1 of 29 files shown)


app/models/manual_record.rb
```diff
@@ -76,7 +76,7 @@ class Edition
     field :state, type: String
     field :version_number, type: Integer
     field :document_ids, type: Array
-    field :removed_document_ids, type: Array
+    field :removed_document_ids, type: Array, as: :removed_section_ids
```
chrislo: I tried to find the documentation for the `as` option just in case it changes/goes away when #929 lands. I don't think we need to worry about it as this change goes away from 335688 onwards.

app/repositories/section_repository.rb
```diff
@@ -103,7 +103,7 @@ def only_section_ids_for(collection)
     collection.all
       .order_by([:updated_at, :desc])
       .only(:document_id, :updated_at)
```
chrislo: Does this use of `document_id` need to change?

db/migrate/20170329103706_rename_manual_record_editions_removed_document_ids_to_removed_section_ids.rb
```diff
@@ -0,0 +1,17 @@
+class RenameManualRecordEditionsRemovedDocumentIdsToRemovedSectionIds < Mongoid::Migration
+  def self.up
+    ManualRecord::Edition.collection.update(
```
chrislo: Is it worth defining a `ManualRecord::Edition` class within this migration so that it can run if that class name changes? I don't think the project has been very consistent with that in the past, so I'm not concerned if we don't.

floehopper: I did consider it, but given that we've already set the precedent of deleting migrations that have run in production, I decided it wasn't worth it. Also it would be easy to add if it turns out we need it.

floehopper:
@chrislo: Thanks.

floehopper:
I've rebased this against master again and tested the migrations with the Mongoid changes in place and it all appears to work OK. Force pushing in preparation for merging.

PR #935, opened 2017-03-29
Revert "Upgrade to Mongoid 2.8.1"

floehopper:
Reverts alphagov/manuals-publisher#928

@chrislo & I think this must be what is causing the [asset pre-compilation to fail on integration](https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/11779/consoleFull).


chrislo:
It looks like all the changes introduced in PR #928 are reversed in this PR 👍

PR #936, opened 2017-03-30
Use MONGODB_URI environment variable in Mongoid config and upgrade Mongoid to v2.8.1

### LINKED ISSUE #930: Use MONGODB_URI instead of MONGODB_NODES in mongoid config

floehopper:
Now that the application is using Ruby v2.2.4, it seems as if Mongoid can parse URIs containing multiple comma-separated hosts. This means we can use the (standard) `MONGODB_URI` environment variable in the Mongoid configuration instead of the (non-standard) `MONGODB_NODES`.

We've tested these changes on the integration box where `MONGODB_URI` is set to:

`mongodb://mongo-1.backend,mongo-2.backend,mongo-3.backend/govuk_content_production`

as opposed to `MONGODB_NODES` which is set to:

```
- mongo-1.backend:27017
- mongo-2.backend:27017
- mongo-3.backend:27017
```

We believe the two configurations are equivalent (`27017` is the default port and `govuk_content_production` is separately specified using the `database` config key), both indicating a replica set made up of three nodes. The deployment succeeded and the app seemed to behave correctly.

This then allowed us to re-apply the changes in #928 to upgrade Mongoid from v2.5.2 to v2.8.1 which we previously had to revert in #935 due to a problem in the integration deployment - specifically during the pre-compilation of assets which happens to be the first time the production Mongoid configuration was used.

Once this change has been deployed to production, we ought to be able to close #930 and remove the generation of the `MONGODB_NODES` environment variable in [puppet](https://github.com/alphagov/govuk-puppet/blob/06acc5ce08e11bee1d05d11e47c6afacb073a7f0/modules/govuk/manifests/apps/manuals_publisher.pp#L96-L118).


steventux:
:+1: This looks good to me, glad to see the back of `MONGODB_NODES`

floehopper:
@steventux: Thanks!

PR #938, opened 2017-03-31
Rename record arg -> edition for *Marshaller#load & #dump

floehopper:
As far as I can tell, these methods are only being called from `ManualRepository#store` and `ManualRepository#build_manual_for`. In both of those cases, the 2nd argument is an instance of a `ManualRecord::Edition` and not a `ManualRecord` as the name "record" seems to imply.

Using the name "edition" is less confusing.


floehopper:
@chrislo: Thanks.

Merging.

Issue #937, opened 2017-03-31
Consider using Pundit for authorization c.f. Specialist Publisher


floehopper:
https://github.com/elabs/pundit

https://github.com/alphagov/specialist-publisher/tree/master/app/policies

boffbowsh:
Or https://github.com/ryanb/cancan. We use cancan in support and maslow, and pundit in signon.

PR #939, opened 2017-03-31
Extract User#manual_records to DRY up controller methods

floehopper:
The logic for scoping the manual records to be used by the various repositories was repeated in each of the three controllers.

I paired with @chrislo on this change which extracts the duplicate logic into `User#manual_records`. While `User#manual_records` may not be the best long-term home for this logic, we think it's definitely an improvement. On a similar note, it's not ideal that we're using the `PermissionChecker` inside `User#manual`, but we think it'll do for now.

One benefit of this change is that we've been able to write a unit test for the extracted logic.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -228,23 +228,11 @@ private
   end
 
   def repository
-    if current_user_is_gds_editor?
-      ScopedManualRepository.new(ManualRecord.all)
-    else
-      ScopedManualRepository.new(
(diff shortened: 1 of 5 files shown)


floehopper:
@chrislo: Thanks. I've just re-worked this so the new method is called `User#manual_records` instead of `User#manuals`, because I think the former is clearer. I've force-pushed my changes. Also I opened #937 to remind us to aim to use Pundit in the future.

floehopper:

PR #940, opened 2017-04-03
Collapse validators and ManualWithSection into Manual

floehopper:
In general we're aiming to move towards a more standard Rails application. Wrapping instances of `Manual` with classes like `ManualWithSections`, `NullValidator` & `ManualValidator` rather than implementing the relevant behaviour directly on `Manual` is not very standard in a Rails app. This PR aims to collapse the behaviour in these other classes into `Manual` as a step towards making the app easier to understand. I'm hopeful that the behaviour added to `Manual` in this PR can later be further simplified using standard Rails mechanisms, e.g. model associations.

The actual motivation to do this came from trying to get rid of the various manual repositories. This proved difficult, because it was never very clear whether the code was dealing with an instance of `Manual` or of `ManualWithSections`.

The approach I've taken is as follows:

1. Instantiate a `ManualWithSections` instance inside each `Manual` and delegate the relevant methods to it. Instantiate a `Manual` instance in places where we were previously instantiating a `ManualWithSections` instance.
2. Gradually move methods in `ManualWithSections` up into `Manual`.
3. Remove `ManualWithSections` class once all methods are removed.
4. Collapse `NullValidator` into `Manual`.
5. Collapse `ManualValidator` into `Manual`.

See the individual commit notes for more details.


floehopper:
@chrisroos: Thanks.

Merging.

PR #941, opened 2017-04-03
Use methods on Manual vs manual repositories classes from ManualsController services

floehopper:
I paired with @chrislo on the beginnings of this PR. It follows on from #940 with the idea of making the app more "standard". The idea here is for the `ManualsController` services to call methods on `Manual` instead of the various manual repository classes. Although for the moment the new methods introduced on `Manual` delegate to instances of the manual repository classes, I'm hoping that by hiding this behind the more Rails-y API of `Manual`, we can more easily convert `Manual` to be a more standard Rails model class.

We found that to do this we had to pass the current user to most of the new methods on `Manual`. However, we're pretty confident that we can deal with this by using [Pundit scopes](https://github.com/elabs/pundit/blob/72c6c1a6060e020d2f8cad999374b6221a27ec4a/README.md#scopes) as suggested by #937.

Furthermore, in hindsight, I wish I had only passed in the current user and not the entire controller context to each service. However, once I'd started, it felt simpler (and not too awful) to keep things consistent.

Also in hindsight it might've been slightly better to extract each new `Manual` method across all services in a single commit, rather than the service-by-service approach which I've taken.

Anyway, I'm pretty happy with the result and I think we ought to be able to apply the same overall approach to the other two controllers and hence eventually get rid of the manual repositories altogether.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -11,7 +11,6 @@ class ManualsController < ApplicationController
 
   def index
     service = ListManualsService.new(
-      manual_repository: associationless_repository,
       context: self,
     )
     all_manuals = service.call
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -50,6 +50,7 @@ class ManualsController < ApplicationController
       manual_repository: repository,
       manual_builder: ManualBuilder.create,
       attributes: create_manual_params,
+      context: self,
     )
     manual = service.call
     manual = manual_form(manual)
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -120,7 +120,6 @@ class ManualsController < ApplicationController
 
   def publish
     service = QueuePublishManualService.new(
-      repository: repository,
       manual_id: manual_id,
       context: self,
     )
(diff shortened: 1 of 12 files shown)


floehopper:
Rebasing against master and force-pushing to remove initial temporary squashed merge commit.

chrisroos:
This is an observation rather than a request for changes.

I think this PR would've been slightly easier to review if there were commit comments explaining why some/all of the changes have been made. I found myself having to go back and forward through the individual commits to get enough context to help me understand the changes. Taking the first three commits, for example:

1. Extract Manual.all
2. Remove unused ListManualsService constructor argument
3. Remove unused ManualsController#associationless_repository

I would've found these easier to understand if commit 2 explained that the constructor argument was no longer used because of the change in commit 1. And if commit 3 explained that `#associationless_repository` was no longer used because of the change in commit 2.

chrisroos:
From the PR description:

> Furthermore, in hindsight, I wish I had only passed in the current user and not the entire controller context to each service. However, once I'd started, it felt simpler (and not too awful) to keep things consistent.

Presumably the consistency you mention here is with the other services that are already instantiated with a controller context? Rather than consistency just within this branch?

I first understood it to be the latter of those two options (consistency within the branch) but am now taking it to be the former (consistency with existing code).

floehopper:
@chrisroos: Thanks for your feedback. Having just discussed it in person, we've agreed that in this case it's not worth the time it would take to re-write/restructure the commits. However, I will bear it in mind for future pull requests.

Merging.

PR #943, opened 2017-04-03
Remove RepositoryPaginator

chrislo:
The last reference to this class was removed in 43e2e753.


chrislo:
Force pushing before merging.

PR #944, opened 2017-04-03
Use methods on Manual vs manual repositories from SectionsController services

floehopper:
This follows on from #941, but makes the changes for the `SectionsController` services rather than the `ManualsController` services.

I paired with @chrislo on this.

@chrisroos: I'm afraid it was too late to learn the lessons of #941 and so it uses a similar style, but we can address those concerns if you like.


floehopper:
@chrisroos: Thanks.

Rebasing against master in preparation for merging.

PR #947, opened 2017-04-04
Use methods on manual instead of repository in attachment services

chrislo:
This PR removes the dependency on a ManualRepository from the SectionAttachment services (in a similar fashion to #941 and #944. 

I've also added a couple of commits to remove the builder argument from NewSectionAttachmentService.

Note - there are no unit tests for the controller or services modified in this PR. The functionality is covered by a feature spec (in `features/attachments.feature`) that didn't require changing as part of this refactor. I'm reasonably confident that those feature tests show that the behaviour hasn't changed as part of this refactor, and I don't plan to add the missing unit tests at this stage.


chrislo:
@floehopper thank you! I've addressed the comment about tests in the PR note.

PR #950, opened 2017-04-04
Simplify ManualBuilder

chrisroos:
I've simplified `ManualBuilder` by removing unnecessary flexibility and indirection and by moving all the behaviour to the `#call` method. I find the resulting class easier to understand.

### DIFF
--- a/app/models/builders/manual_builder.rb
+++ b/app/models/builders/manual_builder.rb
@@ -12,17 +12,7 @@ class ManualBuilder
   end
 
   def call(attrs)
-    @attrs = attrs
-
-    factory.call(defaults.merge(attrs))
-  end
-
-private
-
-  attr_reader :slug_generator, :factory, :attrs
-
-  def defaults
-    {
+    default_attrs = {
       id: SecureRandom.uuid,
       slug: slug_generator.call(attrs.fetch(:title)),
       summary: "",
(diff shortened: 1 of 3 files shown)


app/models/builders/manual_builder.rb
```diff
@@ -1,10 +1,6 @@
 require "securerandom"
 
 class ManualBuilder
-  def self.create
-    ManualBuilder.new
-  end
-
```
floehopper: Minor: In the commit message, I would say you are "in-lining" the method, not just removing it.

app/models/builders/manual_builder.rb
```diff
@@ -24,10 +18,14 @@ def call(attrs)
       use_originally_published_at_for_public_timestamp: true,
     }
 
-    factory.call(default_attrs.merge(attrs))
+    attrs = default_attrs.merge(attrs)
```
floehopper: Minor: I'd have a slight preference for not re-assigning `attrs` here which I find a bit confusing, i.e. in-line this local variable into the call to `Manual.new` on the next line.

app/models/builders/manual_builder.rb
chrisroos: I've updated the commit message to read "Inline ManualBuilder.create".

app/models/builders/manual_builder.rb
chrisroos: I've changed the code to `manual_attrs = default_attrs.merge(attrs)` to avoid the confusion.

I haven't inlined `default_attrs.merge(attrs)` in `Manual.new` because that would mean we'd be reading the `sections` and `removed_sections` from `attrs` but instantiating the object with `default_attrs.merge(attrs)`. While this wouldn't actually change the behaviour it would mean the code was slightly different than the code I took from the factory that I'm inlining.

PR #951, opened 2017-04-04
Remove SectionValidator

chrisroos:
The SectionValidator class was being used to add validation to the
Section model. Adding the validation to the model directly is more
Rails-like and results in code that's easier to understand.

I temporarily removed these validations and observed the following two
cucumber scenarios in creating-and-editing-a-manual.feature failing:

* Try to create an invalid section
* Previewing a manual with invalid HTML

I'd normally expect unit test coverage around this validation but I'm
not planning to add that at the moment.


chrisroos:

PR #952, opened 2017-04-04
Rename entity to section in ChangeNoteValidator

chrisroos:
We only ever initialize a ChangeNoteValidator with a Section(ish) thing
(e.g. SectionValidator). Renaming `entity` to `section` makes this more
explicit.


chrisroos:

PR #953, opened 2017-04-05
Reduce number of references to manual repositories outside Manual class (part 1)

floehopper:
This continues the work we've done in #941, #944 & #947 in services not used by controllers and some other places. The hope is that by continuing this work, eventually we'll be able to remove the manual repositories.

### DIFF
--- a/app/models/manual.rb
+++ b/app/models/manual.rb
@@ -26,7 +26,7 @@ class Manual
   end
 
   def self.all(user)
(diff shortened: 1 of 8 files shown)


chrislo:
Looks good to me! I think I might have been tempted to squash bfead58 c7e8008 and e0ef9e9 together, but I don't think you need to do that.

PR #955, opened 2017-04-05
Remove scoped manual repository

chrislo:
This PR removes the ScopedManualRepository since it was primarily delegating all of its methods to ManualRepository. The only difference was the addition of the "marshallers" which we have inlined into ManualRepository in ccfac12.

### DIFF
--- a/app/models/manual.rb
+++ b/app/models/manual.rb
@@ -22,11 +22,11 @@ class Manual
   attr_accessor :sections, :removed_sections
 
   def self.find(id, user)
-    ScopedManualRepository.new(user.manual_records).fetch(id)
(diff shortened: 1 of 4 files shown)


lib/manual_withdrawer.rb
```diff
@@ -7,7 +7,8 @@ def initialize(logger)
 
   def execute(manual_id)
     service = WithdrawManualService.new(
-      manual_repository: ScopedManualRepository.new(ManualRecord.all),
+      manual_repository:
```
floehopper: Very minor: I'm not sure this newline should be here.

chrislo:
Force pushing before merging.

PR #954, opened 2017-04-05
Reduce number of references to manual repositories outside Manual class (part 2)

floehopper:
This follows on from #953 and is a further effort at reducing the number of places which depend on the various manual repositories. There are still some left to deal with, notably `VersionedManualRepository`.

PR #956, opened 2017-04-05
Reduce number of references to manual repositories outside Manual class (part 3)

floehopper:
This follows on from #953 & #954 and is a further effort at reducing the number of places which depend on the various manual repositories. In this case, I'm specifically dealing with `VersionedManualRepository`. There are still some occurrences left to deal with, e.g. `ManualRepository::NotFoundError`.

### DIFF
--- a/app/models/manual.rb
+++ b/app/models/manual.rb
@@ -47,6 +47,11 @@ class Manual
     ManualRepository.new(user.manual_records).store(self)
   end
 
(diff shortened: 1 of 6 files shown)


floehopper:
@chrisroos: Thanks. I'll do as you suggest and add the PR description text to the commit note.

floehopper:

PR #959, opened 2017-04-05
Update Jenkinsfile to use govuk.buildProject() function

### LINKED ISSUE #898: Intermittent test failures on CI
We feel as though we've seen intermittent failures on CI so I'm opening this issue to allow us to record information about when they happen.

floehopper:
The `Jenkinsfile` contained a lot of custom code that is now available within the [`govuk_jenkinslib.groovy` shared library](https://github.com/alphagov/govuk-puppet/blob/c439fe32a2c0aebf2b0941f3cab4960d2f183ae5/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy). By changing our `Jenkinsfile` to rely on the `govuk.buildProject()` function within this library, we can take advantage of the latest developments and be more consistent with other projects.

I've made the changes to the `Jenkinsfile` in a step-by-step fashion in order to use the commit notes to document why particular bits of code are no longer necessary. Thus the `Jenkinsfile` is not left in a working state until the penultimate commit.

The change in this PR should mean that the "Run tests" stage is locked via the "Lockable Resources" Jenkins plugin so that only one build can run its tests on a given CI agent at one time. See the commit titled "Remove ThrottleJobProperty from Jenkinsfile" and [this PR](https://github.com/alphagov/govuk-puppet/pull/5641) for details. Hopefully this should fix #898.

Note that I haven't resolved the Ruby linting TODO added in the "Remove call to govuk.rubyLinter() from Jenkinsfile" commit, but I'm not sure it's worth delaying getting this PR merged. I think we could address it separately.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,6 @@
 #!/usr/bin/env groovy
 
 REPOSITORY = 'manuals-publisher'
-DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
 
 def runTests() {
   echo 'Running tests'
@@ -10,11 +9,6 @@ def runTests() {
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-
-    govuk.initializeParameters([
-      'IS_SCHEMA_TEST': 'false',
-      'SCHEMA_BRANCH': DEFAULT_SCHEMA_BRANCH,
-    ])
 
     stage('Checkout') {
       checkout scm
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,10 +17,6 @@ node {
       'SCHEMA_BRANCH': DEFAULT_SCHEMA_BRANCH,
     ])
 
-    if (!govuk.isAllowedBranchBuild(env.BRANCH_NAME)) {
-      return
-    }
-
     stage('Checkout') {
       checkout scm
       govuk.cleanupGit()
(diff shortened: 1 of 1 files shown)


chrisroos:
> Note that I haven't resolved the Ruby linting TODO added in the "Remove call to govuk.rubyLinter() from Jenkinsfile" commit, but I'm not sure it's worth delaying getting this PR merged. I think we could address it separately.

I'm happy to postpone making this change but I think it might be better to capture this in a GitHub issue rather than as a TODO in the code. What do you think, @floehopper?

chrisroos:
Have you compared the output of the Jenkins job in master versus this branch? Are you confident that we're not losing anything (aside from the slight difference in Ruby linting) by making this change?

floehopper:
@chrisroos:

> I'm happy for you to merge it if you're confident that we're not losing anything by making this change.

The only thing I know we're losing is lint-ing of some files/directories as per my comment in the PR description:

* Before the change: `rubyLinter("app bin config features Gemfile lib spec")`
* After the change: `rubyLinter("app lib spec test")`

So the following files/directories would no longer be Ruby lint-ed in CI:

* `bin`
* `config`
* `features`
* `Gemfile`

There doesn't seem to be an obvious way to override this, which is why I've left it for now.

Having said that, I've just noticed that we have a `lint.rake` file which defines a "lint" rake task which is run as part of the "default" rake task which is run in CI. This rake task runs the linter against the following: `app bin config features Gemfile lib spec`, and so maybe we're not losing anything after all.

floehopper:

PR #960, opened 2017-04-06
Revert "Merge pull request #929 from alphagov/upgrade-mongoid-to-3"

chrisroos:
This reverts commit 5eb24945c170ee9ad12df8e0ce3aaba4727dd191, reversing
changes made to a73ed988c8d9f5087aa4694a7a3bbb0f4a02243f.

I ran `git revert -m1 5eb24945c170ee9ad12df8e0ce3aaba4727dd191` to
reverse the changes that were merged in PR #929.

This is because we've realised that the upgrade to Mongoid 3 is causing
problems with the app in integration: the /manuals index page is now
very slow to load and there seems to be a problem with creating a new
manual in that we see an error saying "Manual not found".

We're reverting while to give us more time to investigate these
problems.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,7 @@ gem "rails", "3.2.22.3"
 
 # Alphabetical order please :)
 gem "airbrake", "~> 4.3"
+gem "bson_ext", "1.12.1"
 gem "faraday", "0.9.0"
 gem "fetchable", "1.0.0"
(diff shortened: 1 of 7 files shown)


chrisroos:
I'm closing this now that PR #963 has been merged. I'm hopeful that we can avoid reverting the upgrade to Mongoid 3 but it's easy enough to do in future if necessary.

PR #961, opened 2017-04-06
Inline manual repository marshallers

chrislo:
This PR removes the ManualPublishTaskAssociationMarshaller and the SectionAssociationMarshaller by inlining their functionality into the only place that calls them, ManualRepository. 

I've had to re-write the unit tests that were on the SectionAssociationMarshaller to fit with the setup of the ManualRepository specs. This has left the spec file in a bit of a mess, but I wanted to try to make it as clear as possible the provenance of these specs. I've decided to defer refactoring / tidying up ManualRepository for now, I think that can be a separate PR. 

The build, and in particular the integration tests, passes for each commit.


spec/repositories/manual_repository_spec.rb
```diff
@@ -163,20 +161,18 @@
         .with(arguments)
     end
 
-    it "returns the built manual" do
-      expect(repo[manual_id]).to be(manual)
-    end
-
```
floehopper: I assume the justification for removing this is:

> The `it returns the built manual` spec is now double-tested so it has also been removed.

However, I don't completely understand what you mean by "double-tested". Can you elaborate?

app/repositories/manual_repository.rb
```diff
@@ -82,10 +82,14 @@ def build_manual_for(manual_record)
...
 
-    manual_with_sections = SectionAssociationMarshaller.new.load(base_manual, edition)
+    manual_with_sections = add_sections_to_manual(base_manual, edition)
     add_publish_tasks_to_manual(manual_with_sections)
   end
 
+  def add_sections_to_manual(manual, edition)
+    SectionAssociationMarshaller.new.load(manual, edition)
+  end
+
```
floehopper: Minor: It would be nice to give a brief summary of why you are doing this in the commit note.

spec/repositories/manual_repository_spec.rb
```diff
@@ -139,10 +139,19 @@
...
       allow(record_collection).to receive(:find_by).and_return(manual_record)
       allow(manual_record).to receive(:latest_edition).and_return(edition)
       allow(Manual).to receive(:new).and_return(manual)
+      allow(SectionRepository).to receive(:new).with(manual: manual).and_return(section_repository)
+      allow(manual).to receive(:'sections=')
+      allow(manual).to receive(:'removed_sections=')
+      allow(edition).to receive(:section_ids).and_return([:section_id])
+      allow(edition).to receive(:removed_section_ids).and_return([:removed_section_id])
+      allow(section_repository).to receive(:fetch).with(:section_id).and_return(:section)
+      allow(section_repository).to receive(:fetch).with(:removed_section_id).and_return(:removed_section)
```
floehopper: Minor: Having this number of stubs is a bit of an indicator that the code under test may not be in the best shape. However, I imagine this is what you mean when you wrote the following in the PR description:

> I've had to re-write the unit tests that were on the SectionAssociationMarshaller to fit with the setup of the ManualRepository specs. This has left the spec file in a bit of a mess, but I wanted to try to make it as clear as possible the provenance of these specs. I've decided to defer refactoring / tidying up ManualRepository for now, I think that can be a separate PR.

One way I can imagine making this simpler is to use an actual `Manual` instance with real sections, rather than stubbing everything. However, I'm happy that this can be done separately later.

spec/repositories/manual_repository_spec.rb
```diff
@@ -131,10 +146,25 @@
...
-    it "calls dump on the section association marshaller with the manual domain object and edition" do
+    it "saves associated sections and removed sections" do
+      repo.store(manual)
+
+      expect(section_repository).to have_received(:store).with(section)
+      expect(section_repository).to have_received(:store).
+        with(removed_section)
+    end
+
+    it "updates associated document ids on the edition" do
```
floehopper: Minor: "associated document ids" -> "associated section ids" in spec name.

spec/repositories/manual_repository_spec.rb
```diff
@@ -131,10 +146,25 @@
...
+      repo.store(manual)
+
+      expect(edition).to have_received(:section_ids=).with([section.id])
+    end
+
+    it "updates associated removed document ids on the edition" do
```
floehopper: Minor: "associated removed document ids" -> "associated removed section ids" in spec name.

spec/repositories/manual_repository_spec.rb
chrislo: Yes - that's a good point. This style of testing is common to both the ManualRepository and SectionAssociationMarshaller specs before this PR, so I decided to move them across with as few changes as I could. But I think a good next step would be to make them less brittle (or to test their functionality in a different way if we move it to Manual)

chrislo:
Thanks for the feedback @floehopper. I've fixed up your suggestions and am rebasing/force-pushing in preparation for merging.

PR #962, opened 2017-04-06
Reduce number of references to manual repositories outside Manual class (part 4)

floehopper:
This follows on from #953, #954 & #956 and is a further effort at reducing the number of places which depend on the various manual repositories.

After these changes, I think the various manual repositories are only referred to from within `Manual` or within the repository classes themselves.


floehopper:
@chrislo: Thanks.

> In the commit message for 7b540d2 should it read "the behaviour is *now* just standard Ruby behaviour"? 

I'll fix that up and get this merged.

PR #968, opened 2017-04-10
Inline manual repository into manual

chrisroos:
This PR removes the ManualRepository class. We do that in three steps: 

- First simplify and refactor the methods on ManualRepository to avoid moving redundant code
- Move the methods from ManualRepository onto Manual (which is the only class that calls them). We keep the specs as close as possible to the ones in ManualRepository at this point by isolating them within their own context in `manual_spec.rb`
- Refactor the Manual specs, dropping the heavy mocking and stubbing and relying on some FactoryGirl generated database objects to ensure the new methods on Manual are interfacing correctly with ManualRecord.

There's probably some more refactoring we could do for the Manual specs but I think they can be left for later.


chrislo:
I've rebased this on master and force pushed. I think it's ready for a review.

chrisroos:
This all looks good to me, @chrislo.

I have some minor suggestions about 4 commit messages but I'll leave it to you to decide whether you think the changes are worth making:

* Commit: "Inline ManualRepository#[] into Manual#find"
  * Reference commit "Inline ManualRepository#store into Manual#save" in the commit message to help someone looking at this commit in future.

* Commit "Inline ManualRepository#all into Manual#all"
  * Reference commit "Inline ManualRepository#store into Manual#save" in the commit message to help someone looking at this commit in future.

* Commit: "Move remaining ManualRepository methods to Manual"
  * Maybe rename commit subject to "Move ManualRepository#build_manual_for to Manual" and add a note about also adding the associated private methods.

* Commit: "Improve Manual#save tests"
  * Maybe copy the message from either the commit with subject "Improve Manual.find spec" or from the commit with subject "Improve specs for Manual.all".

chrisroos:
@chrislo: I don't seem to be able to review/approve this PR - presumably because I created it. I suggest you approve it and get it merged.

chrislo:
Thanks @chrisroos - I've improved the commit messages as you suggested. 

Force pushing before merging.

chrislo:
I'm going to wait until #976 is merged before fixing up this PR.

chrislo:
I'm going to remove the commits where I was experimenting with getting the build to pass, rebase, and force-push.

spec/models/manual_spec.rb
```diff
@@ -318,6 +318,117 @@
...
+        record = ManualRecord.where(manual_id: manual.id).first
+        edition = ManualRecord::Edition.where(
+          manual_record_id: record.id
+        ).first
+
+        expect(edition.title).to eq(manual.title)
+        expect(edition.summary).to eq(manual.summary)
+        expect(edition.body).to eq(manual.body)
+        expect(edition.state).to eq(manual.state)
+        # TODO: something better than `to_i` to compare times?
```
floehopper: It would be nice to get rid of this "TODO" as soon as possible.

Issue #861, opened 2017-03-13
Unexpected updates for The Highway Code on 1 Mar 2017


brenetic:
This problem was originally reported in [Zendesk ticket 1901209][zendesk-1901209] on 1 Mar 2017.

There are unexpected [updates listed for The Highway Code][highway-code-updates] against 1 March 2017.

We expect to see "Annex 5 Penalties" with a single change described as "Updated the penalty table to increase penalty points for using a hand-held mobile phone when driving from 3 to 6.".

We're also seeing:

* Road users requiring extra care (204 to 225)
  * Updated Rule 209 to add an image of the school bus road sign.
* Signals to other road users
  * New section added.
* Signals by authorised persons
  * New section added.
* Road markings
  * New section added.
* Vehicle markings
  * New section added.
* Annex 8. Safety code for new drivers
  * Updated the Pass Plus email address to passplus@dvsa.gov.uk

For all the updates on 1 March 2017 we're seeing two identical change notes, one with a timestamp of 2017-03-01T09:49:18Z and the other with a timestamp of 2017-03-01T11:03:18Z.

[highway-code-updates]: https://www.gov.uk/guidance/the-highway-code/updates
[zendesk-1901209]: https://govuk.zendesk.com/agent/tickets/1901209

chrisroos:
@brenetic: Do you know what's happening with this issue? I can see that the problem is also mentioned in [this Trello card](https://trello.com/c/JGuGBZD2/118-fix-update-date-in-manuals-publisher) but that the card's been archived today. Do you know whether I can close this issue?

chrisroos:
I've added a [comment to the Trello card to ask Liz L if this is being handled elsewhere](https://trello.com/c/JGuGBZD2/118-fix-update-date-in-manuals-publisher#comment-58e3c1b93e73079175ebc081).

brenetic:
@chrisroos I'm afraid I have not been keeping track of it. I had thought that Survey and Support were doing it? I can ask Liz L tomorrow

chrisroos:
I've spoken to Antonia (whose team is looking after manuals-publisher) and agreed to investigate this issue.

chrisroos:
I've updated the description of the issue now that I understand more about the problem.

chrisroos:
[...]

PR #973, opened 2017-04-13
Add migration to fix Highway Code change note data

chrisroos:
This is the first part of the fix for the problem described in issue
 #861.

There are unexpected updates listed for The Highway
Code[1] against 1 March 2017.

We expect to see "Annex 5 Penalties" with a single change described as
"Updated the penalty table to increase penalty points for using a
hand-held mobile phone when driving from 3 to 6.".

We're also seeing:

* Road users requiring extra care (204 to 225)
  * Updated Rule 209 to add an image of the school bus road sign.
* Signals to other road users
  * New section added.
* Signals by authorised persons
  * New section added.
* Road markings
  * New section added.
* Vehicle markings
  * New section added.
* Annex 8. Safety code for new drivers
  * Updated the Pass Plus email address to passplus@dvsa.gov.uk

For all the updates on 1 March 2017 we're seeing two identical change
notes, one with a timestamp of 2017-03-01T09:49:18Z and the other with a
timestamp of 2017-03-01T11:03:18Z.

The problem is due to duplicate "major update" `SectionEdition`s and
their change notes. The presence of these SectionEditions result in
"duplicate" `PublicationLog`s which are sent to the Publishing API as
change notes.

This migration updates the affected `SectionEdition`s to make them minor
changes and removes their changes notes.

We still need to rebuild the `PublicationLog`s and update the content in
the Publishing API but they will happen separately.

[1]: https://www.gov.uk/guidance/the-highway-code/updates

### DIFF
--- a/db/migrate/20170413112512_fix_change_note_data_for_the_highway_code.rb
+++ b/db/migrate/20170413112512_fix_change_note_data_for_the_highway_code.rb
@@ -1,0 +1,62 @@
+class FixChangeNoteDataForTheHighwayCode < Mongoid::Migration
+  def self.reset_change_note(section_slug, created_at)
+    SectionEdition.where(
+      slug: 'guidance/the-highway-code/' + section_slug,
+      created_at: created_at
+    ).each do |section_edition|
+      section_edition.update_attributes!(change_note: nil, minor_update: true)
+      edition_attributes = [
+        section_edition.id,
+        section_edition.title,
+        section_edition.created_at
+      ]
+      puts "Updated SectionEdition with attributes: #{edition_attributes.join(', ')}"
+    end
+  end
+
+  def self.up
+    created_at_0925 = Time.parse('2017-03-01 09:25:48')
+    created_at_1102 = Time.parse('2017-03-01 11:02:33')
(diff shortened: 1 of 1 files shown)


chrislo:
This looks good to me and I think it should work fine on production-like data. 

In development however I ran the migration and got:

```
13:17 $ be rake db:migrate
==  FixChangeNoteDataForTheHighwayCode: migrating =============================
rake aborted!
StandardError: An error has occurred, 20170413112512 and all later migrations canceled:

Only expected to find a single match, found 0
```

Which is caused by line 7 of the migration returning 0 results. Perhaps it could fail more gracefully in this situation?

chrislo:
I think it also might be worth mentioning in the commit note why there is no down migration - I think it would be possible to write one, but given that the data is bad, I'm not sure it's needed.

chrisroos:
Thanks @chrislo. I've updated the migration based on your feedback. Can you take another look when you have a moment, please?

chrislo:
Looks great!

chrisroos:
Thanks @chrislo! Merging now.

PR #974, opened 2017-04-13
WIP: Investigating build failures on CI

chrisroos:
Anecdotally it feels as though we've seen an increase in build failures recently.

My guess is that it's related to PR #959 that switched us to using the standard Jenkinsfile.

The standard Jenkinsfile only drops/creates/loads the database schema if config/database.yml exists (see [line 155][line-155]. I _think_ this means that our Mongo database is hanging around between tests.

I don't understand why we're only seeing some builds fail. I think I'd expect to see them all fail if the database really is persisting between runs.

This is a work in progress as I need to revisit the final change (dropping the database in `beforeTest`) to see whether that's the right/best thing to do.

[line-155]: https://github.com/alphagov/govuk-puppet/blob/c439fe32a2c0aebf2b0941f3cab4960d2f183ae5/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L155

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ node {
     beforeTest: {
       govuk.setEnvar("TEST_COVERAGE", "true")
       stage("Set up the database") {
-        runRakeTask("db:drop db:create db:schema:load")
+        runRakeTask("db:drop")
       }
     }
   )
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,9 @@ node {
     sassLint: false,
     beforeTest: {
       govuk.setEnvar("TEST_COVERAGE", "true")
+      stage("Set up the database") {
+        runRakeTask("db:drop db:create db:schema:load")
+      }
     }
   )
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ node {
     beforeTest: {
       govuk.setEnvar("TEST_COVERAGE", "true")
       stage("Set up the database") {
-        sh("bundle exec rake db:drop")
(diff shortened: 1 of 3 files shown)


chrisroos:
This most recent failure might be interesting to investigate - https://ci.integration.publishing.service.gov.uk/job/manuals-publisher/job/investigating-build-failures-on-ci/14/console.

chrislo:
I left the following comments on slack:

```
We've been having some problems with intermittent build failures on CI on the `manuals-publisher` project. For example build https://ci.integration.publishing.service.gov.uk/job/manuals-publisher/job/inline-manual-repository-into-manual/14/ and https://ci.integration.publishing.service.gov.uk/job/manuals-publisher/job/inline-manual-repository-into-manual/15/ build the same (I *think*) code but one build passed, the other failed.

The failure was caused by what looks like stale data in the mongo database.

[3:05] 
We're using "database cleaner" which should prevent this - but I wondered if anyone else has seen the issue before? The only difference I could see is that #14 ran on `ci-agent-2` whereas #15 ran on `ci-agent-4` - but that could be coincidental.
```

```
Chris Lowis [3:26 PM] 
I've just seen the same issue triggering the build manually. https://ci.integration.publishing.service.gov.uk/job/manuals-publisher/job/inline-manual-repository-into-manual/18/ ran on `ci-agent-3` and passed. https://ci.integration.publishing.service.gov.uk/job/manuals-publisher/job/inline-manual-repository-into-manual/17/ ran on `ci-agent-4` and failed. Any ideas?
```

chrislo:
I believe the build failures should have been fixed by #976

chrisroos:
Good work @chrislo! I'm going to close this PR.

PR #977, opened 2017-04-20
Inline SectionBuilder into Manual

chrisroos:
I don't think the `SectionBuilder` was adding any real value and that removing it makes the code easier to understand.


chrisroos:
Thanks @floehopper. I'll get this merged.

PR #978, opened 2017-04-20
Inline ManualBuilder into Manual

chrisroos:
I don't think the ManualBuilder was adding any real value and think that removing it makes the code easier to understand.


chrisroos:
Thanks @chrislo!

Issue #809, opened 2017-02-23
Remove Services.publishing_api


chrisroos:
As per [@h-lame's comment on PR #798](https://github.com/alphagov/manuals-publisher/pull/798#discussion_r98922123):

> I'd be 100% surprised if we need publishing_api and publishing_api_v2. I suspect we can drop the former, and rename the latter to publishing_api

---

Update 4 Apr 17

We're no longer using v1 of the Publishing API in manuals-publisher so it can be removed. Once removed we should probably rename `Services.publishing_api_v2` to `.publishing_api`.

h-lame:
I am willing to say that the answer here is "no, we don't need to" and (without cheking) that the code is only using v2.  v1 is effectively, if not officially deprecated.

floehopper:
I attempted to remove all use of v1 Publishing API in #832, but did not completely succeed because of #851. If we can fix the latter or remove the relevant script then we should be able to close this issue.

chrisroos:
PR #903 addresses issue #851 and, in doing so, removes the last occurrences of v1 of the Publishing API. If/when PR #903 if merged we should be able to resolve this issue by removing `Services#publishing_api`.

chrisroos:
PR #903 has been merged so I think we can now remove all occurrences of Publishing API v1.

@floehopper: You're currently assigned to this issue. Are you planning to do the work or is it up for grabs for anyone to do?

chrisroos:
@floehopper has confirmed in Slack that he's not actively working on this so I'm going to remove him from this issue.

chrisroos:
I've updated the title now that we know we can remove v1 of the Publishing API from this app.

chrisroos:
I've opened PR #979 to address this issue.

chrisroos:
I've just merged PR #979 which resolves this issue.

PR #979, opened 2017-04-20
Remove Publishing API v1 client

chrisroos:
This removes `Services.publishing_api` and renames `Services.publishing_api_v2` to `Services.publishing_api` now that we're only using a single version of the Publishing API.

This address issue #809.


chrisroos:
Thanks @floehopper. I'll rebase and get it merged.

PR #980, opened 2017-04-20
Move service classes into sub-directories / namespaces

floehopper:
Most of the service classes are only used in a single controller, but all the services were in a single directory and it was hard to see just the services used by a single controller.

This PR moves all of the service classes into sub-directories / namespaces based on the model class which they manipulate. Initially I used the controller name prefixes for the sub-directory names, but that didn't feel quite right. I also contemplated getting rid of the `Service` class name suffix on all the service classes, but I thought that made the code a bit harder to read/understand.

I'm not 100% sure the way I've done it is ideal, but I do think it's an improvement having the service classes in separate directories like this.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -4,7 +4,7 @@ class ManualsController < ApplicationController
   before_filter :authorize_user_for_publishing, only: [:publish]
 
   def index
-    service = ListManualsService.new(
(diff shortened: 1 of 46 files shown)


chrislo:
I also loaded these changes in development to make sure the autoloading was working.

floehopper:
@chrislo: Thanks for reviewing. I'll get it merged shortly.

PR #982, opened 2017-04-21
Add task to sync section slugs with titles

steventux:
https://trello.com/c/sigxg9uu/914-1908077-fix-numbering-of-manual-chapters-in-urls

Adds a rake task to help curate manual sections where the slug is "out of sync" with the section title.

This scenario is particularly obvious when manual sections are titled using a numerical prefix, the slug is initially generated from the title so `1. Introduction` generates `1-introduction`.
If the section entitled `1. Introduction` is changed to `2. Introduction` the slug appears incorrect.

The root cause is simply that updating the title does not update the slug, it's arguable that this should happen in the application but for the few issues we have raised in zendesk I've added a simple set of rake tasks to report on and re-slug the offending sections.

There's no attempt to resolve conflicts ie. re-slugging a section to an already occupied path, but these conflicts are reported by the `sections:report` task.

See https://govuk.zendesk.com/agent/tickets/1908077 for some context of the problems experienced by publishers.


lib/section_slug_synchroniser.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    [amendments, conflicts]
+  end
+
+  def section_slug(slug)
+    slug.split("/").last
+  end
+
+  def user
+    OpenStruct.new(manual_records: manual_records)
```
kevindew: Not sure if there's anything we can do to fix it, but this is a bit confusing - We're having to create like a proxy object here to pretend to be a User? and that needs to have every manual_record in the system?

lib/section_slug_synchroniser.rb
```diff
@@ -0,0 +1,75 @@
...
+    if conflicts.any?
+      log "The following sections cannot be reslugged:"
+      conflicts.each do |k, v|
+        log "'#{section_slug(k)}' would change to '#{section_slug(v)}' but this is already in use."
+      end
+    end
+  end
+
+  def synchronise
+    amendments, = analyse_sections
```
kevindew: might be an improvement to have this as `analyse_sections.first` or similar, the trailing comma looks a bit like a typo

lib/section_slug_synchroniser.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def analyse_sections
+    amendments = {}
+    conflicts = {}
+    section_slugs = sections.map(&:slug)
+
+    sections.each do |section|
+      slugified_title = SlugGenerator.new(prefix: manual.slug).call(section.title)
+      # Does the title match the slug?
+      unless slugified_title == section.slug
```
kevindew: might be cleaner to have a `next unless` to save an indentation?

lib/section_slug_synchroniser.rb
steventux: It's tricky because [`Manual#find`](https://github.com/alphagov/manuals-publisher/blob/master/app/models/manual.rb#L26) takes a user as the second parameter. I agree it's not ideal creating this proxy, it's something we also do in other utility classes but this could lead to problems if the publishing application was to use them.
An alternative would be to pass an email address into this class to lookup the user...

floehopper: I think it would be more consistent with the rest of the code to use [`User.gds_editor`](https://github.com/alphagov/manuals-publisher/blob/afe9b17a53035b75cf3d9bb44e3028ee5c554518/app/models/user.rb#L24) here unless there's a reason the lookup should be restricted in scope somehow. FWIW the longer term plan is to use something like a [Pundit scope](https://github.com/elabs/pundit/blob/72c6c1a6060e020d2f8cad999374b6221a27ec4a/README.md#scopes) as mentioned in #937.

steventux: Thanks @floehopper I'd not spotted that, updated.

chrisroos:
Hey @steventux. Are you waiting on anything from us (@floehopper, @chrislo and I) for this PR?

steventux:
@chrisroos no I've amended based on @floehopper's comments. @kevindew are the amendments ok with you?

chrisroos:
Good stuff @steventux. I just wanted to make sure that we weren't holding you up.

lib/section_slug_synchroniser.rb
floehopper: @steventux: Cool. I think your update means `SectionSlugSynchroniser#manual_records` is no longer used.

Issue #858, opened 2017-03-13
Convert all scripts into rake tasks


floehopper:
From [this comment][1] by @h-lame:
> A good thing to do would be to turn these into rake tasks as then they can be run through jenkins and we'd get history of how often they're used. (See: https://github.com/alphagov/govuk-puppet/blob/f1c27d74dd0f361159bb5531473343730c692ea2/modules/govuk_jenkins/templates/jobs/run_rake_task.yaml.erb)

[1]: https://github.com/alphagov/manuals-publisher/pull/832#issuecomment-285417474

PR #983, opened 2017-04-24
Convert scripts to rake tasks

chrislo:
I've converted the remaining scripts in `bin` to rake tasks. I've tried to do as straight-forward a conversion as I can, the main changes are around the way scripts and rake tasks handle arguments. 

I've left `bin/content_model_pdf_report` as a script as the side effect of this script is to generate a CSV file on the server, and I don't think running this from Jenkins would allow the jenkins user to get at the CSV file it produces. 

Fixes: #858


chrislo:
Thanks @floehopper! 

I've converted the `content_model_pdf_report` script to output to STDOUT and converted it to a rake task. 

Asking on Slack, there doesn't seem to be any standard way to namespace rake tasks. The `publisher` application [has them mostly in the global namespace](https://github.com/alphagov/publisher/blob/master/lib/tasks/bulk_discard_placeholder_draft_content.rake#L1), so I'll leave it as such here for now.

I've updated the README too.

chrislo:
Rebasing and force-pushing before merging.

PR #984, opened 2017-04-24
Collapse ChangeNoteValidator into Section

floehopper:
The "Rails Way" is to use [`ActiveModel::Validations`](http://api.rubyonrails.org/v3.2.22.3/classes/ActiveModel/Validations.html). Prior to this PR, there was a mixture of the latter and the custom validation logic in `ChangeNoteValidator`. This led to non-standard code and unnecessary levels of indirection, making the code hard to understand. This PR collapses the logic in `ChangeNoteValidator` into `Section` using the standard approach.

I've also done a little simplification of the logic and renaming of variables missed in the `SpecialistDocument` -> `Section` rename.

I did also consider further simplifying `Section#change_note_required?` by inlining `#never_published?` into it. However, it feels as if "never published"  is communicating something slightly different from "not published", even though the implementation appears to be the same. So I've decided to leave this for now.


app/models/section.rb
```diff
@@ -159,14 +160,18 @@ def eql?(other)
...
-protected
-
-  attr_reader :slug_generator, :edition_factory
-
   def never_published?
     !published?
   end
 
+  def change_note_required?
+    !(never_published? || minor_update?)
```
chrisroos: I wonder whether this would be a bit clearer written as `published? && !minor_update?`?

chrisroos: Or even `published? && major_update?` although it's not obvious that this alone would warrant the addition of a `major_update?` method.

floehopper: @chrisroos: As I alluded to in the PR description I did consider this, but decided that `never_published?` seems to be communicating something a bit different to `!published?` and that method is used elsewhere. I guess I could introduce `ever_published?`. I'll do that now.

floehopper: Hmm. Change of plan. There's already a `Section#has_ever_been_published?` method which seems to be doing something different again. I think this is all a bit of a tangle and so I'm going to leave it for now.

floehopper:

PR #985, opened 2017-04-24
Inline SectionFactory into Section

floehopper:
This continues the process of removing unnecessary levels of indirection in the code.

The new `Section.build` method replaces `SectionFactory#call`.


floehopper:
@chrisroos: Thanks.

Merging.

PR #986, opened 2017-04-24
Collapse ManualWithPublishTasks into Manual

floehopper:
This is a continuation of the work to reduce the unnecessary number of levels of indirection in the code and to move towards the "Rails Way".

The `ManualWithPublishTasks` class was only used to provide the association between a `Manual`
and its `ManualPublishTasks`. I'm hoping that this will eventually end up as an ActiveRecord/Mongoid association, but this is a good first step and removes an unnecessary and confusing level of indirection.


floehopper:
@chrislo: Thanks.

PR #987, opened 2017-04-24
Remove optional edition_factory arg from Section#initialize

floehopper:
The default value of this argument was only ever overridden in specs. We can use partial mocking to avoid needing to inject the edition factory in the specs and thus simplify the code.


floehopper:

PR #988, opened 2017-04-24
Remove unused DateValidator

floehopper:
This hasn't been used since the [big commit][1] where the bulk of the specialist publisher code was removed.

Note that this was not obvious, because the validation was being triggered by the `date` option in lines like this:

```ruby
    validates :tribunal_decision_decision_date, presence: true, date: true
```

There are no such lines left in the code now.

[1]: https://github.com/alphagov/manuals-publisher/commit/3162f0e72fd53bfd1cf87f5661bb198ff785940e


floehopper:

PR #989, opened 2017-04-24
Inline SectionRepository into Section

chrisroos:
This is a continuation of the work to reduce the unnecessary number of levels of indirection in the code and to move towards the "Rails Way".


spec/models/section_spec.rb
```diff
@@ -142,6 +142,24 @@ def key_classes_for(hash)
     end
   end
 
+  describe '#save' do
+    it 'saves the last two editions' do
```
floehopper: I notice that the spec name has changed from "saves the the two most recent editions" to "saves the last two editions". Was this intentional? The new spec name seems more accurate in a way, although it does make me wonder whether these editions should be explicitly ordered the same way as when the section is built. At the very least I think this change merits a mention in the commit note.

app/repositories/section_repository.rb
```diff
@@ -9,10 +9,6 @@ def fetch(section_id)
     end
   end
 
-  def initialize(manual:)
-    @manual = manual
-  end
-
```
chrisroos: I've updated the commit subject line to "Move manual param from SectionRepository#initialize to #fetch".

app/repositories/section_repository.rb
```diff
@@ -1,19 +1,10 @@
...
-  def fetchable_fetch(key, not_found_value = NO_DEFAULT_GIVEN)
-    if not_found_value != NO_DEFAULT_GIVEN && block_given?
-      raise ArgumentError.new("Cannot provide both a default arg and block to #fetch")
-    end
-
+  def fetchable_fetch(key)
     result = public_send(:[], key)
 
     if result.nil?
-      return yield(key) if block_given?
```
chrisroos: This is removed in the commit with subject "Simplify SectionRepository#fetchable_fetch".

app/repositories/section_repository.rb
```diff
@@ -10,15 +10,9 @@ def fetch(section_id)
...
-      result = Section.build(manual: manual, id: section_id, editions: editions)
-    end
-
-    if result.nil?
       raise KeyError.new("key not found #{section_id}").extend(NotFoundError)
     else
-      result
+      Section.build(manual: manual, id: section_id, editions: editions)
     end
   end
```
chrisroos: I've corrected the typo.

spec/repositories/section_repository_spec.rb
```diff
@@ -70,23 +70,23 @@ def build_published_edition(version: 1)
...
 
     context "when there are no editions" do
       before do
         allow(editions_proxy).to receive(:to_a).and_return([])
       end
 
-      it "returns nil" do
-        expect(section_repository[section_id]).to be_nil
+      it "raises a key error" do
+        expect { section_repository.fetch(section_id) }.to raise_error(KeyError)
```
chrisroos: I've added the following to the commit note: 

> I'm happy to accept this change in behaviour because we were never calling `#[]` directly: we only ever used `#fetch` which invoked `#[]`.

spec/models/section_spec.rb
chrisroos: I rewrote the spec based on the current behaviour rather than the behaviour that was being implied by the previous spec. I would expect the spec the change again if/when we need to update the implementation.

I've added the following to the commit note to try to make this clearer:

> I've chosen to rewrite the tests rather than try to port them from section_repository_spec as I didn't think the existing ones were particularly good. The previous tests appear to imply some behaviour that doesn't actually exist in the implementation. The new tests describe the behaviour as it stands. The difference between the behaviour being described in the old tests and the actual behaviour might be indicative of a problem with the code but I'm happy to leave that for now.

spec/models/section_spec.rb
```diff
@@ -142,6 +142,24 @@ def key_classes_for(hash)
...
+      editions = [
+        edition_1 = double(:edition_1),
+        edition_2 = double(:edition_2),
+        edition_3 = double(:edition_3)
+      ]
+      section = Section.new(slug_generator, 'section-id', editions)
+
+      expect(edition_1).to_not receive(:save!)
+      expect(edition_2).to receive(:save!)
+      expect(edition_3).to receive(:save!)
```
chrisroos: I rewrote the spec based on the current behaviour rather than the behaviour that was being implied by the previous spec. I would expect the spec the change again if/when we need to update the implementation.

I've added the following to the commit note to try to make this clearer:

> I've chosen to rewrite the tests rather than try to port them from section_repository_spec as I didn't think the existing ones were particularly good. The previous tests appear to imply some behaviour that doesn't actually exist in the implementation. The new tests describe the behaviour as it stands. The difference between the behaviour being described in the old tests and the actual behaviour might be indicative of a problem with the code but I'm happy to leave that for now.

chrisroos:
Thanks for reviewing, @floehopper. I've addressed your feedback, rebased on master and force pushed in preparation for merging.

spec/models/section_spec.rb
floehopper: OK. I've added #1016 to capture this possible problem, because I'm concerned that the code my only be working by accident and we probably won't have any other trigger to investigate it.

PR #1019, opened 2017-04-25
[Do not merge] Mark copied sections as cloned...

steventux:
https://trello.com/c/WKtK3EIj/920-investigate-underlying-bug-with-setting-publication-date-of-a-manual

Do not merge because we're testing alternatives to this approach.

When a manual is updated in order to amend the public updated at field,
all the manual sections are copied with no amendments, this makes it
difficult for the `PublicationLogger` to discern which section change notes
are copies of the previous edition and so we may end up with duplicate
notes appearing as inaccuracies in the manual change history.
A recent issue with the Highway Code meant these change notes appeared
under the wrong date and implied the guidance was added at a different
time to legislative changes.
This commit persists a boolean  field to indicate a cloned
`SectionEdition`, the `PublicationLogger` will ignore these editions and
no `PublicationLog` record will be created, this will mean duplicate change
notes do not get sent to the content store and rendered in the manual history.


chrisroos:
Hi @steventux. Do we definitely need to duplicate these `SectionEdition`s when changing the public updated at timestamp? I'm wondering whether we could solve this problem by avoiding the duplicate `SectionEdition`s in the first place.

steventux:
@chrisroos I had wondered the same, it doesn't seem optimal to clone sections like this and export them again as drafts. To be honest I'd assumed there was [_a good reason for doing this_](https://github.com/alphagov/manuals-publisher/commit/9c7951b0e2074ad41e3114bee70102a14d959d4d) given @h-lame's commit message.

steventux:
@chrisroos so having investigated the approach of only updating the manual on a different branch in a dev VM and on integration the main side-effect I can see is that `section` content items will end up with different `first_published_at` dates to the manual they belong to. This synchronising of dates is a bit buried but is present here https://github.com/alphagov/manuals-publisher/blob/master/app/exporters/section_publishing_api_exporter.rb#L55
I'm not fully sure what the implications of this are for the frontend (I can't find a rendered `first_published_at` date for a section) but it feels like the data will become polluted by this.

steventux:
Closing in favour of https://github.com/alphagov/manuals-publisher/pull/1040

PR #1023, opened 2017-04-26
Remove section renderer

### LINKED ISSUE #1021: Remove SectionRenderer
This appears to be used in two places, and might be sensibly inlined in the Section::PreviewService

Arises from #995

chrislo:
Fixes #1021 

This PR removes the SectionRenderer class, putting the responsibility instead inside the Section::PreviewService. I think this is an improvement as it moves more of the functionality inside the service objects, but the fact that this service is called "Preview" and is used in the `SectionPublishingAPIExporter` makes me think it is either doing too much, or needs a different name.

### DIFF
--- a/app/controllers/sections_controller.rb
+++ b/app/controllers/sections_controller.rb
@@ -71,7 +71,6 @@ class SectionsController < ApplicationController
 
   def preview
     service = Section::PreviewService.new(
-      section_renderer: SectionRenderer.new,
       context: self,
     )
     section = service.call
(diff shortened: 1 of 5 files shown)


chrislo:
@floehopper yes, I started thinking something along those lines too as I made the changes. I'll see what that looks like in a branch.

floehopper:
@chrislo: It think it might make sense to close this PR while you investigate the alternative approach?

chrislo:
I'll have a think and reopen another one later.

PR #1024, opened 2017-04-26
Collapse VersionedManualRepository into Manual

### LINKED ISSUE #1018: Remove use of repositories

floehopper:
This is a continuation of the process to simplify the app, reducing the number of levels of indirection, and making the app more like a standard Rails app.

This PR moves the remaining behaviour from `VersionedManualRepository` into `Manual` making as much use of existing behaviour as possible, thereby reducing some non-obvious duplication. I'm pretty convinced some more simplification is possible, but this will do for now.

This is the last repository class and therefore it also closes #1018.


floehopper:
@chrislo: Thanks for reviewing and testing! 😄 

Rebasing against master and rebasing in preparation for merging.

PR #1025, opened 2017-04-26
Display new section slug in ManualRelocator

chrisroos:
I'm fairly confident that this message should print the old
(`section.slug`) and new (`new_section_slug`) slugs instead of the
printing the old slug twice.

Have run the script I can see that the section slug is indeed updated to
`new_section_slug`.

This appears to have been incorrect since the `ManualRelocator` class
was added in 43b669600d3ddc30fd9fccb7bdbe1e4b4ecf9887.

### DIFF
--- a/lib/manual_relocator.rb
+++ b/lib/manual_relocator.rb
@@ -132,9 +132,8 @@ private
     new_section_ids.each do |section_id|
       sections = all_editions_of_section(section_id)
       sections.each do |section|
-        reslug_msg = "Reslugging section '#{section.slug}'"
         new_section_slug = section.slug.gsub(from_slug, to_slug)
-        puts "#{reslug_msg} as '#{section.slug}'"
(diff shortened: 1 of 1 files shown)


lib/manual_relocator.rb
```diff
@@ -134,7 +134,7 @@ def reslug
       sections.each do |section|
         reslug_msg = "Reslugging section '#{section.slug}'"
         new_section_slug = section.slug.gsub(from_slug, to_slug)
-        puts "#{reslug_msg} as '#{section.slug}'"
+        puts "#{reslug_msg} as '#{new_section_slug}'"
```
floehopper: Minor: I might be tempted to inline `reslug_msg`. If you did that as a refactoring before your change, it would make it more obvious that there is a problem.

chrisroos: I've added a commit titled "Inline reslug_msg in ManualRelocator#reslug" to do as you've suggested.

chrisroos:

PR #1026, opened 2017-04-26
Remove reference to app/models/builders directory from README

floehopper:
This directory was removed in [this commit][1].

[1]: e99a62a5f3f1c449945bae652d77f2433a631a9a


floehopper:

Merging.

PR #1029, opened 2017-04-26
Log PublishManualWorker errors to Rails log

chrisroos:
Logging exceptions to the Rails log helped me debug a problem I was seeing when trying to publish manuals in development. Prior to this change I wasn't getting any information when `Manual::PublishService.new#call` failed. In my particular case I needed to set `PLEK_SERVICE_WHITEHALL_ADMIN_URI=https://www.gov.uk` when running `sidekiq` so that the app could communicate with the Organisations API.

### DIFF
--- a/app/workers/publish_manual_worker.rb
+++ b/app/workers/publish_manual_worker.rb
@@ -50,6 +50,7 @@ private
   end
 
   def log_error(error)
+    Rails.logger.error "#{self.class} error: #{error}"
     Airbrake.notify(error)
   end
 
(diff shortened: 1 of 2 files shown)


chrisroos:
> However, I do slightly wonder whether it's "safe" (e.g. in atomicity terms) for multiple processes to write to a single Rails log file.

I don't know the answer but I'm happy to proceed given that it feels like a relatively low risk and it's a simple change to undo should we need to.

Issue #881, opened 2017-03-16
Add a rake task to allow manuals to be renamed


chrislo:
Arrises from #878

In order to make `ManualRelocator` easier to use and to remind other developers that the code in the class should not be removed we should create a simple rake task that instantiates and calls the class with the command line arguments.

floehopper:
There's also some potentially useful info about this in #851.

chrisroos:
PR #1031 adds a Rake task to allow manuals to be relocated.

chrisroos:
PR #1031 has been merged so I'm closing this issue.

PR #1031, opened 2017-04-26
Add Rake task for ManualRelocator

chrisroos:
This addresses issue #881. We planned to remove the `ManualRelocator` class in PR #878 but closed it without merging as @h-lame explained that this class is still used. We're hoping that making it a Rake task will make it easier to use in future and prevent someone else from trying to remove it.

### DIFF
--- a/lib/tasks/relocate_manual.rake
+++ b/lib/tasks/relocate_manual.rake
@@ -1,0 +1,10 @@
+require "manual_relocator"
+require "logger"
+
+desc <<-EndDesc
+Relocate manual from <from-slug> to <to-slug>
+Both <from-slug> and <to-slug> need to be published manuals
+EndDesc
+task :relocate_manual, [:from_slug, :to_slug] => :environment do |_, args|
+  ManualRelocator.move(args[:from_slug], args[:to_slug])
+end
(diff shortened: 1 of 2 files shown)


lib/tasks/relocate_manual.rake
```diff
@@ -0,0 +1,10 @@
+require "manual_relocator"
+require "logger"
+
+desc <<-EndDesc
+Relocate manual from <from-slug> to <to-slug>
+Both <from-slug> and <to-slug> need to be published manuals
+EndDesc
```
floehopper: Note that only the first line of multi-line descriptions for Rake tasks appear in the output from `rake -T`. In this case the first line makes sense on its own, so I think this is OK. Note also that all lines of the description appear in the output from `rake -D`.

chrisroos:
>  I think it might be worthwhile including a note in the new documentation to say that the existence of this rake task should not be taken to imply that it is necessary. It might also be worth adding something to this effect in the commit notes.

I've added a note to the README that reads:

> NOTE. The behaviour of this script is a little confusing (essentially overwriting one published manual with another) so it's not entirely obvious that it's still required.

I've also tried to make it clearer in the commit notes that I don't understand when one might use this script.

chrisroos:

PR #1033, opened 2017-04-27
Remove unnecessary require statement from sections.rake

floehopper:
This was made redundant in [this recent commit][1].

[1]: https://github.com/alphagov/manuals-publisher/commit/d1ae6f246a9cddb0103ec89a63d2eab64d70128b


chrisroos:
Good spot, @floehopper.

PR #1034, opened 2017-04-27
Introduce SearchIndexAdapter

floehopper:
This new `SearchIndexAdapter`class  is intended to provide the **only** interface that the application code should use to interact with the Rummager API. The intention is that the method signatures, return values and exceptions should all be in terms of the application's own domain model and **not** the classes or concepts specific to the Rummager API, i.e. it should abstract the application away from knowing anything about the underlying external API. The choice of the word "Adapter" in the class name is taken from the [Ports and Adapters Pattern][1].

I've chosen to use "SearchIndex" and not "Rummager" in the class name, because the application shouldn't care what search-indexing service we use. I know that the Rummager application itself abstracts us from the underlying technology (Elastic Search?), but it's still not obvious to me that there's a benefit in the application code "knowing" that it is talking to the Rummager application.

Another motivation for this change is to replace two classes each of which has a single method (`#call`) with a single class which has multiple methods with useful names. I think this is easier to understand and it has allowed me to remove some duplication in the process.

A significant chunk of this PR is taken up with changing the `SectionReslugger` to use the new adapter rather than talking directly to the Rummager API. I'm not completely happy with the final result, but I think it is still a step forward.

Thinking about the bigger picture, another reason for this PR is to demonstrate the approach I'm hoping to take with other external services, notably the Publishing API. I'm hoping that approach will allow us to simplify the plethora of classes currently in the `app/exporters` directory.

Note that I considered inlining the code in `RummagerIndexer` into `SearchIndexAdapter`, but in the end I decided against it for now.

[1]: http://alistair.cockburn.us/Hexagonal+architecture


chrisroos:
Is there a word missing from subject commit, "Assert SectionReslugger old slug content from search index"?

chrisroos:
I can see that there aren't any tests for the new `SearchIndexAdapter`. Presumably because there weren't previously any tests for `RummagerManualWithSectionsWithdrawer` or `RummagerManualWithSectionsExporter`. What do you think about adding some? Or do you think we've got good enough coverage elsewhere?

chrisroos:
@floehopper and I have just chatted about adding tests for `SearchIndexAdapter` and have agreed that adding tests to assert that we're calling the Rummager API correctly is a good idea.

I'm going to mark the PR as approved because I don't think it'll be the end of the world if the tests aren't added in this branch.

floehopper:

I've added a spec for `SearchIndexAdapter`, fixed the commit note subject line, and I'm now rebasing against master in preparation for merging.

floehopper:
Amending last commit and force-pushing to fix Rubcop offenses.

PR #1035, opened 2017-04-27
Rename SectionEdition#section_id field

chrisroos:
We're planning to introduce a Mongo-backed `Section` class in the near future. This will have its own internal Mongoid ID field as well as the UUID currently stored on the `SectionEdition` records. Renaming `SectionEdition#section_id` to `SectionEdition#section_uuid` should hopefully help us avoid confusion when we introduce the `Section` class.

I plan to rename `Section#id`, `ManualRecord::Edition#section_ids` and `ManualRecord::Edition#removed_section_ids` in future PRs.

### DIFF
--- a/db/migrate/20170427130832_rename_section_edition_section_id_to_section_uuid.rb
+++ b/db/migrate/20170427130832_rename_section_edition_section_id_to_section_uuid.rb
@@ -1,0 +1,13 @@
+class RenameSectionEditionSectionIdToSectionUuid < Mongoid::Migration
+  def self.up
+    SectionEdition.all.each do |section_edition|
+      section_edition.rename(:section_id, :section_uuid)
+    end
+  end
+
+  def self.down
+    SectionEdition.all.each do |section_edition|
+      section_edition.rename(:section_uuid, :section_id)
+    end
+  end
+end
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -37,7 +37,7 @@ class Section
 
   def self.find(manual, section_id, published: false)
     editions = SectionEdition
-      .where(section_id: section_id)
+      .all_for_section(section_id)
       .order_by([:version_number, :desc])
       .to_a
       .drop_while { |e| published && !e.published? }
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -35,9 +35,9 @@ class Section
     ]
   end
 
-  def self.find(manual, section_id, published: false)
+  def self.find(manual, section_uuid, published: false)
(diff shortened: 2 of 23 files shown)


db/migrate/20170427130832_rename_section_edition_section_id_to_section_uuid.rb
```diff
@@ -0,0 +1,13 @@
...
+      section_edition.rename(:section_id, :section_uuid)
+    end
+  end
+
+  def self.down
+    SectionEdition.all.each do |section_edition|
+      section_edition.rename(:section_uuid, :section_id)
+    end
+  end
+end
```
floehopper: Minor: Are you confident that this will be quick enough running against production data? Did you consider using a bulk update command like the one in [RenameManualRecordEditionsRemovedDocumentIdsToRemovedSectionIds](https://github.com/alphagov/manuals-publisher/blob/master/db/migrate/20170329103706_rename_manual_record_editions_removed_document_ids_to_removed_section_ids.rb)?

chrisroos:
> Minor: In the PR description you talk about plans to make other changes in future commits. I think it would be worth saying that these commits will be in subsequent PRs, assuming that's what you mean.

Done.

db/migrate/20170427130832_rename_section_edition_section_id_to_section_uuid.rb
chrisroos: I ran it locally using recent data from production and it ran in a matter of seconds. I'm hoping that's going to be representative of how it'll run in production too!

I considered using the bulk update command but decided against it when I saw that this version seems to run quickly enough.

chrisroos:
I'm going to investigate whether renaming the field will require us to recreate the Mongo indexes.

chrisroos:
Thanks for the review, @floehopper. I've rebased on master in order to get it merged.

chrisroos:
As far as I can see we **will** need to regenerate the Mongo indexes after deploying this branch. 

I imported a recent dump of production data and ran the migration in this branch. I used the MongoDB Compass tool to view the indexes defined on the manual_section_editions collection and can confirm that running the migration has no effect on them.

I suggest we run `rake db:mongoid:remove_indexes` and `rake db:mongoid:created_indexes` after deploying this branch. That'll ensure we remove the old `document_id_1` and `section_id_1` indexes and create the new `section_uuid_1` index.

Mentioning @floehopper / @chrislo / @steventux in the hope that it'll help us to remember to do this after the next deploy.

PR #1036, opened 2017-04-27
Upgrade non-locked Gems and their dependencies

chrisroos:
I've updated each of the Gems listed in the Gemfile (using `bundle update <gem-name`) that isn't locked to a specific version.

The tests (`bundle exec rake`) pass after each commit. The app appears to work fine in development after these upgrades.

My hope is that this'll make it slightly easier to upgrade other Gems, including Rails, in future.

I've not upgraded govuk-lint or gds-api-adapters in this branch because they both require more work. I'll create issues to capture the fact that they still need to be upgraded.


chrisroos:
I've created issue #1037 to remind us to upgrade govuk-lint and issue #1038 to remind us to upgrade gds-api-adapters.

chrisroos:
> Minor: I think it would be clearer if the title of this PR was something more like: "Upgrade non-locked gems and their dependencies".

I've updated the title of the PR. Ideally I'd rename the branch but that'd involve opening a new PR so I'm going to do that.

chrisroos:
> Minor: I think it would be clearer if the subject line of commits used the word "upgrade" vs "update", although I know the change is made by running `bundle update`.

Done.

chrisroos:
> Minor: I think it would be clearer if subject lines of commits made it clear whether the version of the mentioned gem has changed and/or the versions of its dependencies:
> 
> * "Upgrade raindrops" (only gem version has changed; not versions of dependencies)
> * "Upgrade gds-sso dependencies" (only dependency versions have changed)
> * "Upgrade test-unit and its dependencies" (both gem version and versions of dependencies have changed)

Done.

chrisroos:
> I'm not 100% sure that our test suite alone is good enough to make these changes with confidence. Out of interest did you look at any of the change notes for the primary/production gems? If you did, I think it would be worth noting that.

I didn't look at any change notes. I've added a note to each commit to say that I'm relying on the tests to give me confidence (despite the fact that I share your concern that our tests alone aren't good enough).

chrisroos:
Thanks for reviewing, @floehopper. I've addressed your feedback, rebased on master and force pushed in preparation for merging.

PR #1040, opened 2017-04-28
Explicitly draft new sections with a nil change_note

steventux:
https://trello.com/c/WKtK3EIj/920-investigate-underlying-bug-with-setting-publication-date-of-a-manual

2nd attempt, see https://github.com/alphagov/manuals-publisher/pull/1019 for initial go at this.

When updating the `first_published_at` date the change note passed to the
section should be nil, this will then be identified and skipped by the PublicationLogger
and avoid duplicate change notes appearing in the frontend as [documented here](https://github.com/alphagov/manuals-publisher/issues/970)
Previously the section was copied along with the message in the change note
and these could appear several times in the updates page on the frontend.

### DIFF
--- a/app/services/manual/update_original_publication_date_service.rb
+++ b/app/services/manual/update_original_publication_date_service.rb
@@ -39,8 +39,8 @@ private
 
   def update_sections
     manual.sections.each do |section|
-      # a no-op update will force a new draft if we need it
-      section.update({})
+      # a nil change note will omit this update from publication logs
+      section.update(change_note: nil)
     end
   end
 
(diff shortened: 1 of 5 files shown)


app/services/manual/update_original_publication_date_service.rb
```diff
@@ -40,7 +40,7 @@ def manual
   def update_sections
     manual.sections.each do |section|
       # a no-op update will force a new draft if we need it
```
kevindew: Does this comment need updating now it's no longer an empty hash?

spec/services/manual/update_original_publication_date_service_spec.rb
```diff
@@ -41,11 +41,11 @@
       )
   end
 
-  it "forces all the manuals sections to require an export with an empty update message" do
+  it "forces all the manuals sections to require an export with an nil change note" do
```
kevindew: s/an/a/ ?

steventux:
Thanks @kevindew updated as per the comments :+1:

Issue #996, opened 2017-04-25
Extract API adapters from exporters


chrislo:
The idea is that there will be a single adapter class for each of the external APIs that the application needs to interact with. These classes should be the **only** way that application code interacts with these external services. The method signatures, method return values and exceptions raised by methods on these classes should all be in terms of the application's own domain model and **not** the classes or concepts specific to the external API, i.e. they should abstract the application away from knowing anything about the underlying external API. The choice of the word "Adapter" is taken from the [Ports and Adapters Pattern][1].

[1]: http://alistair.cockburn.us/Hexagonal+architecture

floehopper:
The following were motivated by this issue:

* [Introduce SearchIndexAdapter](https://github.com/alphagov/manuals-publisher/pull/1034)
* [Collapse formatters and RummagerIndexer into SearchIndexAdapter](https://github.com/alphagov/manuals-publisher/pull/1046)

floehopper:
The following PRs were also motivated by this issue:

* [Introduce OrganisationsAdapter & inline OrganisationFetcher into it](https://github.com/alphagov/manuals-publisher/pulls/1053)
* [Extract version_type methods for Manual and Section](https://github.com/alphagov/manuals-publisher/pulls/1077)
* [Simplify PublishingApiUpdateTypes](https://github.com/alphagov/manuals-publisher/pulls/1078)
* [Use PublishingAdapter in CliManualDeleter instead of using Publishing API directly](https://github.com/alphagov/manuals-publisher/pulls/1110)

floehopper:
The work on this is largely completed, but there are still three places where the Publishing API is used directly, i.e. not via the `PublishingAdapter`:

* [`ManualRelocator`](https://github.com/alphagov/manuals-publisher/blob/4d6f8b02f034b2e39746b22dae658c408b640220/lib/manual_relocator.rb)
* [`MarkedSectionDeleter`](https://github.com/alphagov/manuals-publisher/blob/4d6f8b02f034b2e39746b22dae658c408b640220/lib/marked_section_deleter.rb)
* [`DuplicateDraftDeleter`](https://github.com/alphagov/manuals-publisher/blob/4d6f8b02f034b2e39746b22dae658c408b640220/lib/duplicate_draft_deleter.rb)

The need for some of these might be reduced/removed by the following PRs:

* [Avoid persisting updated sections when saving to Publishing API fails](https://github.com/alphagov/manuals-publisher/pull/1141)
[...]

PR #1046, opened 2017-05-03
Collapse formatters and RummagerIndexer into SearchIndexAdapter

floehopper:
This PR is part of #996 and collapses the behaviour from the following classes into `SearchIndexAdapter`:

* `RummagerIndexer`
* `AbstractIndexableFormatter`
* `ManualIndexableFormatter`
* `SectionIndexableFormatter`

I think the resulting code makes it easier to see the mapping from a `Manual` and its `Sections` into calls to the Rummager API, mainly because the code is all *in one place* and not scattered amongst an inheritance hierarchy.

The `SearchIndexAdapter` spec which I added in [this commit](https://github.com/alphagov/manuals-publisher/commit/8117e7e77440fd246a672458ca7eaaf644e60cbf) gives sufficient coverage to give me confidence that these changes haven't broken anything.

I think this validates [the idea of extracting API adapters from exporters](https://github.com/alphagov/manuals-publisher/issues/996).


floehopper:

Merging.

Issue #995, opened 2017-04-25
Make Renderers more Rails-y


chrislo:
e.g. use a `Presenter` suffix on the class name, move to a `app/presenters` directory, etc.

floehopper:
As per [this comment](https://github.com/alphagov/manuals-publisher/pull/1023#pullrequestreview-34811844), @chrislo & I have agreed that rather than removing the renderers, this issue should be about converting them into more standard Rails presenters. Based on this, I'm going to change the title of this issue and move it out of the [Simplify milestone](https://github.com/alphagov/manuals-publisher/milestone/1) into the [Standardize milestone](https://github.com/alphagov/manuals-publisher/milestone/7).

chrislo:
With the above PRs merged, we have now removed all of the "renderer" classes and replaced them with two new presenters `ManualPresenter` and `SectionPresenter`.

PR #1050, opened 2017-05-03
Convert renderers to presenters

chrislo:
This PR moves the `SectionRenderer` and `ManualRenderer` classes into the `app/presenters` directory, renaming them as `SectionPresenter` and `ManaualPresenter`. This is more consistent with how presenter objects are used in, for example, [publisher](https://github.com/alphagov/publisher/tree/master/app/presenters) and [specialist publisher](https://github.com/alphagov/specialist-publisher/tree/master/app/presenters). 

I've started to refactor `ManualPresenter` too removing delegation to make it clear what the interface of the presenter is. There's quite a bit more to do on `SectionPresenter`, but I don't think this PR leaves things in a too inconsistent state and can be landed while I work on `SectionPresenter`. 

This is part of #995.


app/presenters/manual_presenter.rb
```diff
@@ -16,7 +16,7 @@ def summary
   end
 
   def body
-    GovspeakToHTMLRenderer.create.call(manual).attributes.fetch(:body)
+    GovspeakHtmlConverter.new.call(manual.body)
```
floehopper: I think this change makes the `require "govspeak_to_html_renderer"` statement at the top of the file redundant.

app/presenters/manual_presenter.rb
```diff
@@ -1,7 +1,13 @@
 require "govspeak_to_html_renderer"
 
 class ManualPresenter
-  def call(manual)
+  attr_reader :manual
```
floehopper: Minor/subjective: Does this reader method need to be public or do we even need it at all?

chrislo:
Force pushing before merging.

chrislo:
Thanks @floehopper! I think it's a good idea to not have presented objects returned from the services. I'll create an issue to investigate that.

chrislo:
I raised #1054 to capture the above.

Issue #1017, opened 2017-04-25
Persist Sections in the database


chrisroos:
We think it's going to make the code simpler if we store `Section`s in Mongo and make use of associations between `Section`s and `SectionEdition`s.

chrisroos:
I've started working toward this but have realised that there are probably going to be a few preparatory changes required before I can actually make this change.

chrisroos:
PR #1035 (renaming `SectionEdition#section_id` to `#section_uuid`) should help us persist Sections in Mongo.

chrisroos:
From @chrislo's comment on PR #1052:

> As well as uniqueness, "UUID" implies a specific length and text format. When we introduce the persisted Section model, I think we might want to add a validation on the UUID field to check that it is not only unique, but also has those other properties to prevent someone accidentally persisting a Section with a "UUID" that isn't.

floehopper:
Although we've done some preparatory work towards this, it became obvious that it was going to be a lot of effort and it wasn't obvious that the effort was justified. As a result of this we've decided to focus on [Encapsulate SectionEditions within Section](https://github.com/alphagov/manuals-publisher/issues/1165). After that we can decide whether this is still a useful step.

It's also worth noting that we're wondering whether moving the app to be "fully-migrated" in the Publishing API might be a way to side-step this issue.

PR #1052, opened 2017-05-03
Rename Section#id to Section#uuid

chrisroos:
We're planning to introduce a Mongo-backed `Section` class in the near future. This will have its own internal Mongoid ID field as well as the UUID currently stored on the `SectionEdition` records. Renaming `Section#id` to `Section#uuid` should hopefully help us avoid confusion when we introduce the `Section` class.

I plan to rename `ManualRecord::Edition#section_ids` and `ManualRecord::Edition#removed_section_ids` in future PRs.

This follows on from PR #1035 which renamed `SectionEdition#section_id` to `#section_uuid`.

### DIFF
--- a/spec/models/section_spec.rb
+++ b/spec/models/section_spec.rb
@@ -121,7 +121,7 @@ describe Section do
       end
 
       it 'builds a section using the section id' do
-        expect(Section).to receive(:build).with(including(id: 'section-id'))
+        expect(Section).to receive(:build).with(including(uuid: 'section-id'))
         Section.find(manual, 'section-id')
       end
 
--- a/spec/models/section_spec.rb
+++ b/spec/models/section_spec.rb
@@ -606,9 +606,9 @@ describe Section do
       )
     end
 
-    it "returns a has including the section's id" do
+    it "returns a has including the section's uuid" do
       expect(section.attributes).to include(
-        id: section_uuid,
+        uuid: section_uuid,
       )
     end
   end
--- a/spec/models/section_spec.rb
+++ b/spec/models/section_spec.rb
@@ -165,14 +165,14 @@ describe Section do
   describe "#eql?" do
     let(:editions) { [draft_edition_v1] }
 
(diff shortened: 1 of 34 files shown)


spec/models/section_spec.rb
```diff
@@ -165,14 +165,14 @@ def key_classes_for(hash)
   describe "#eql?" do
     let(:editions) { [draft_edition_v1] }
 
-    it "is considered the same as another section instance if they have the same id" do
+    it "is considered the same as another section instance if they have the same uuid" do
```
chrislo: Now we are using UUIDs the description of this spec reads slightly odd - as they are unique by definition, I'd never expect two instances to have the same UUID.

chrislo:
This looks good to me. A couple of comments: 

- is it worth creating an issue for the improvement suggested in the commit message in cb91133 after this lands? 
- As well as uniqueness UUID implies a [specific length and text format](https://en.wikipedia.org/wiki/Universally_unique_identifier). When we introduce the persisted Section model, I think we might want to add a validation on the UUID field to check that it is not only unique, but also has those other properties. 

👍

chrisroos:
> is it worth creating an issue for the improvement suggested in the commit message in cb91133 after this lands?

I've created issue #1060 to capture this improvement.

chrisroos:
> As well as uniqueness, "UUID" implies a specific length and text format. When we introduce the persisted Section model...

I've added this comment to PR #1017.

spec/models/section_spec.rb
chrisroos: I think this is asserting that a `Section` behaves in an ActiveRecord-like way, i.e. that two in-memory instances of objects representing the same record in a database will be "equal". Can you think of better wording to make this clearer?

chrisroos:
Thanks for the review, @chrislo! I've rebase on master and force pushed in preparation for merging.

chrisroos:
@chrislo: Can you approve this again when you have a moment, please? Rebasing and pushing has dismissed your previous review.

PR #1053, opened 2017-05-03
Introduce OrganisationsAdapter & inline OrganisationFetcher into it

### LINKED ISSUE #1049: Introduce OrganisationsAdapter
I'm investigating this as part of #996.

floehopper:
The new `OrganisationsAdapter` class is intended to provide the **only** interface that the application code should use to interact with the Organisations API. The intention is that the method signatures, return values and exceptions should all be in terms of the application's own domain model and **not** the classes or concepts specific to the Organisations API, i.e. it should abstract the application away from knowing anything about the underlying external API. To this end, I've introduced a new `Organisation` model. I've also introduced an "organisation" factory to DRY up the exporter specs. As before the choice of the word "Adapter" in the class name is taken from the [Ports and Adapters Pattern][1].

This is a continuation of the work described in #996. It's really all about establishing a consistent level of abstraction.

Closes #1049.

[1]: http://alistair.cockburn.us/Hexagonal+architecture

### DIFF
--- a/app/adapters/organisations_adapter.rb
+++ b/app/adapters/organisations_adapter.rb
@@ -1,0 +1,11 @@
+class OrganisationsAdapter
+  def find(slug)
+    response = OrganisationFetcher.fetch(slug)
+    Organisation.new(
+      title: response["title"],
+      web_url: response["web_url"],
+      abbreviation: response["details"]["abbreviation"],
+      content_id: response["details"]["content_id"]
+    )
+  end
+end
--- a/app/adapters/organisations_adapter.rb
+++ b/app/adapters/organisations_adapter.rb
@@ -1,11 +1,19 @@
 class OrganisationsAdapter
+  def initialize
+    @api = Services.organisations
+    @cache = {}
+  end
+
   def find(slug)
-    response = OrganisationFetcher.fetch(slug)
-    Organisation.new(
-      title: response["title"],
-      web_url: response["web_url"],
-      abbreviation: response["details"]["abbreviation"],
-      content_id: response["details"]["content_id"]
-    )
+    @cache.fetch(slug) do
+      response = @api.organisation(slug)
+      organisation = Organisation.new(
+        title: response["title"],
+        web_url: response["web_url"],
+        abbreviation: response["details"]["abbreviation"],
+        content_id: response["details"]["content_id"]
+      )
+      @cache[slug] = organisation
+    end
   end
 end
--- a/app/exporters/publishing_api_draft_section_exporter.rb
+++ b/app/exporters/publishing_api_draft_section_exporter.rb
@@ -1,6 +1,6 @@
 class PublishingApiDraftSectionExporter
   def call(section, manual)
-    organisation = OrganisationFetcher.fetch(manual.organisation_slug)
+    organisation = OrganisationsAdapter.new.find(manual.organisation_slug)
 
     SectionPublishingAPILinksExporter.new(organisation, manual, section).call
     SectionPublishingAPIExporter.new(organisation, manual, section).call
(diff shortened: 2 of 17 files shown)


app/adapters/organisations_adapter.rb
```diff
@@ -0,0 +1,11 @@
+class OrganisationsAdapter
+  def find(slug)
+    response = OrganisationFetcher.fetch(slug)
+    Organisation.new(
+      title: response["title"],
```
chrislo: Minor: I see these lines have been copied from `app/exporters/manual_publishing_api_exporter.rb` but as there's no interpolation happening I think I prefer single quotes (`'`) to double (`"`).

app/exporters/publishing_api_draft_section_exporter.rb
```diff
@@ -1,6 +1,6 @@
 class PublishingApiDraftSectionExporter
   def call(section, manual)
-    organisation = OrganisationFetcher.fetch(manual.organisation_slug)
+    organisation = OrganisationsAdapter.new.find(manual.organisation_slug)
```
chrislo: Would it be any clearer to have a class method `find` that instantiated an instance of `OrganisationsAdaptor`?

app/adapters/organisations_adapter.rb
floehopper: I believe this is an issue across the codebase. So far I've tended to lean towards keeping the double-quotes to make it easier to see more significant changes. In the same way, I've been using double-quotes across the board, e.g. in spec names.

If we are going to address this issue, I think I'd prefer to do it across the board and ideally enable some kind of Rubocop rule to keep the code consistent. So I hope it's OK, but I'm not going to make the change you suggest in this PR.

app/exporters/publishing_api_draft_section_exporter.rb
floehopper: Possibly, but I think I'd prefer to introduce some kind of "global registry" where we can obtain *instances* of these adapters, c.f. the `Services` module. However, I think I'll do that as a separate PR or possibly as part of another PR where having it would make life easier.

floehopper:

Merging.

app/adapters/organisations_adapter.rb
```diff
@@ -0,0 +1,19 @@
+class OrganisationsAdapter
```
chrisroos: @floehopper: It looks like we need to add `require 'services'` to this file:

```
$ rails c
>> OrganisationsAdapter.new
NameError: uninitialized constant OrganisationsAdapter::Services
	from /Users/chrisroos/Code/alphagov/manuals-publisher/app/adapters/organisations_adapter.rb:5:in `initialize'
	from (irb):2:in `new'
	from (irb):2
	from /Users/chrisroos/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/railties-3.2.22.3/lib/rails/commands/console.rb:47:in `start'
	from /Users/chrisroos/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/railties-3.2.22.3/lib/rails/commands/console.rb:8:in `start'
	from /Users/chrisroos/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/railties-3.2.22.3/lib/rails/commands.rb:41:in `<top (required)>'
	from script/rails:4:in `require'
	from script/rails:4:in `<main>'
>> require 'services'
=> true
>> OrganisationsAdapter.new
=> #<OrganisationsAdapter:0x007ff219895720 @api=#<GdsApi::Organisations:0x007ff2198951a8 @options={:logger=>#<NullLogger::Logger:0x007ff219895018>, :endpoint_url=>"http://whitehall-admin.dev.gov.uk"}, @endpoint="http://whitehall-admin.dev.gov.uk">, @cache={}>
```

floehopper: @chrisroos: Thanks. Fixed in #1056.

PR #1055, opened 2017-05-04
Standardise section presenter

chrislo:
This PR makes the SectionPresenter behave in a similar way to the ManualPresenter introduced in #1050. I've also taken the opportunity to more explicitly define and simplify the interface to SectionPresenter which will hopefully make future refactoring easier.

Contributes to #995

### DIFF
--- a/app/controllers/sections_controller.rb
+++ b/app/controllers/sections_controller.rb
@@ -71,7 +71,6 @@ class SectionsController < ApplicationController
 
   def preview
     service = Section::PreviewService.new(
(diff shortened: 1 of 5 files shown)


chrislo:
Thanks @floehopper - I'll fix the inconsistencies between ManualPresenter and SectionPresenter in a separate PR. 

Adding some tests for SectionPresenter feels like a good next step. I was just looking at SectionPresenter#body and I'm fairly certain it can be simplified considerably.

PR #1056, opened 2017-05-04
Add missing require statement to adapter classes

floehopper:
The `Services` module lives in the `lib` directory and so is not auto-loaded.

Thanks to @chrisroos for [spotting this][1].

[1]: https://github.com/alphagov/manuals-publisher/pull/1053#pullrequestreview-36269656


floehopper:

PR #1059, opened 2017-05-04
Remove SectionHeaderExtractor and dependencies

chrislo:
The SectionHeaderExtractor's purpose was to add an attribute called 'headers' to the passed
in Section. It was only called from the SectionPresenter. SectionPresenter does not have a `headers` method in its public interface and as best as I can tell the SectionHeaderExtractor
and its dependency GovspeakHeaderExtractor are redundant. This PR removes them.


chrisroos:
Thanks for talking me through this change, @chrislo :-) It looks good to me.

PR #1064, opened 2017-05-05
Remove unnecessary examples from Publishing API exporter specs

floehopper:
These examples are testing scenarios which do not happen in production, so it's safe to remove them.

The fact that they are shared examples means they are executed multiple times in different contexts exacerbating the problem.


floehopper:
Rebasing and force-pushing in preparation for merging.

PR #1065, opened 2017-05-05
Rename ManualRecord::Edition#section_ids to #section_uuids

chrisroos:
We're planning to introduce a Mongo-backed `Section` class in the near future (see issue #1017). This will have its own internal Mongoid ID field as well as the UUID currently stored on the `SectionEdition` records. Renaming `ManualRecord::Edition#section_ids` to `ManualRecord::Edition#section_uuids` should hopefully help us avoid confusion when we introduce the `Section` class.

I plan to rename `ManualRecord::Edition#removed_section_ids` in a future PR.

### DIFF
--- a/lib/manual_relocator.rb
+++ b/lib/manual_relocator.rb
@@ -37,11 +37,11 @@ private
   end
 
   def old_section_ids
-    @old_section_ids ||= old_manual.editions.flat_map(&:section_ids).uniq
+    @old_section_ids ||= old_manual.editions.flat_map(&:section_uuids).uniq
   end
 
   def new_section_ids
-    @new_section_ids ||= new_manual.editions.flat_map(&:section_ids).uniq
+    @new_section_ids ||= new_manual.editions.flat_map(&:section_uuids).uniq
   end
 
   def validate_manuals
--- a/spec/lib/section_reslugger_spec.rb
+++ b/spec/lib/section_reslugger_spec.rb
@@ -31,7 +31,7 @@ describe SectionReslugger do
       body: 'section-edition-body'
     )
     manual_record.editions.create!(
-      section_ids: [
+      section_uuids: [
         'section-id'
       ]
     )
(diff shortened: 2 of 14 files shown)


lib/manual_relocator.rb
```diff
@@ -37,11 +37,11 @@ def fetch_manual(slug)
   end
 
   def old_section_ids
-    @old_section_ids ||= old_manual.editions.flat_map(&:section_ids).uniq
+    @old_section_ids ||= old_manual.editions.flat_map(&:section_uuids).uniq
   end
 
   def new_section_ids
-    @new_section_ids ||= new_manual.editions.flat_map(&:section_ids).uniq
+    @new_section_ids ||= new_manual.editions.flat_map(&:section_uuids).uniq
```
floehopper: Minor: Is it worth renaming `@old_section_ids` & `@new_section_ids` at the same time?

spec/lib/section_reslugger_spec.rb
```diff
@@ -31,7 +31,7 @@
       body: 'section-edition-body'
     )
     manual_record.editions.create!(
-      section_ids: [
+      section_uuids: [
         'section-id'
```
floehopper: Minor: Is it worth renaming "section-id" at the same time?

lib/manual_relocator.rb
chrisroos: I consciously didn't rename them in this commit. There are quite a few variables/methods in the code that still refer to `id` instead of `uuid` (when dealing with `Section`s) and I plan to open a separate PR to try to rename all those up together.

spec/lib/section_reslugger_spec.rb
chrisroos: I consciously didn't rename them in this commit. There are quite a few variables/methods in the code that still refer to `id` instead of `uuid` (when dealing with `Section`s) and I plan to open a separate PR to try to rename all those up together.

chrisroos:
Thanks for the review, @floehopper!

PR #1066, opened 2017-05-05
Rename ManualRecord::Edition#removed_section_ids to #removed_section_uuids

chrisroos:
We're planning to introduce a Mongo-backed `Section` class in the near future (see issue #1017). This will have its own internal Mongoid ID field as well as the UUID currently stored on the `SectionEdition` records. Renaming `ManualRecord::Edition#removed_section_ids` to `ManualRecord::Edition#removed_section_uuids` should hopefully help us avoid confusion when we introduce the `Section` class.

### DIFF
--- a/app/models/manual.rb
+++ b/app/models/manual.rb
@@ -96,7 +96,7 @@ class Manual
     removed_sections.each(&:save)
 
     edition.section_uuids = sections.map(&:uuid)
-    edition.removed_section_ids = removed_sections.map(&:uuid)
+    edition.removed_section_uuids = removed_sections.map(&:uuid)
 
     manual_record.save!
   end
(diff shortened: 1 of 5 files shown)


app/models/manual.rb
```diff
@@ -277,7 +277,7 @@ def add_sections_to_manual(manual, edition, published: false)
         Section.find(manual, section_uuid, published: published)
       }
 
-      removed_sections = Array(edition.removed_section_ids).map { |section_uuid|
+      removed_sections = Array(edition.removed_section_uuids).map { |section_uuid|
```
floehopper: I was slightly surprised that the `section_uuid ` block arg didn't need renaming. Presumably it was renamed in a previous PR?

chrisroos: Yeah, that was renamed in PR #1052. There are still quite a few variables/methods referring to id instead of uuid when dealing with `Section`s. I plan to fix those separately.

chrisroos:
Thanks for reviewing, @floehopper! I've rebased on master and force pushed. Can you approve it again, please?

PR #1068, opened 2017-05-05
Inline Section#build into Section#new

chrisroos:
I _think_ that having a single way of instantiating `Section`s is going to make things easier when we come to store Sections in Mongo.

### DIFF
--- a/spec/models/section_spec.rb
+++ b/spec/models/section_spec.rb
@@ -111,8 +111,6 @@ describe Section do
   end
 
   describe '.find' do
-    let(:manual) { double(:manual, slug: 'manual-slug') }
-
     context 'when there are associated section editions' do
       let(:section_edition) { FactoryGirl.build(:section_edition) }
       let(:editions_proxy) { double(:editions_proxy, to_a: [section_edition]).as_null_object }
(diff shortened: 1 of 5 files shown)


spec/models/section_spec.rb
```diff
@@ -104,9 +106,18 @@ def key_classes_for(hash)
...
-    let(:manual) { double(:manual, slug: 'manual-slug') }
+  before do
+    allow(SlugGenerator).to receive(:new).with(prefix: manual_slug).and_return(slug_generator)
+  end
+
+  describe '#initialize' do
+    it 'instantiates a slug generator using the manual slug' do
+      expect(SlugGenerator).to receive(:new).with(prefix: manual_slug)
+      subject
+    end
```
floehopper: Minor: I'm not super keen on this example - specifically **expecting** the call to `SlugGenerator.new`. I think it's probably enough that we're stubbing the method in order to inject a mock object and that other examples depend on that injection, i.e. they would fail if a `SlugGenerator` was not instantiated. Having said that, maybe no other examples do depend on it the injection of the mock.

chrisroos: I've removed this test. I think you're right that it's probably good enough that we're stubbing the method to inject a mock.

chrisroos:
Thanks for the review, @floehopper. I've addressed your feedback, rebased on master and force pushed. Can you re-review/approve, please?

PR #1069, opened 2017-05-05
Refactor section presenter

chrislo:
This PR refactors the `SectionPresenter`, simplifying the `body` method and inlining some of the implementation. As a result I've also been able to delete two additional renderers `FootnotesSectionHeadingRenderer` and `GovspeakToHTMLRenderer`. 

I'm quite happy with where I've ended up, but made some slightly odd refactorings along the way. I'd be interested to know if you think this history needs cleaning up.


spec/presenters/section_presenter_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+describe SectionPresenter do
+  let(:section) { double(:section, body: '', attributes: {}, attachments: []) }
+  subject { described_class.new(section) }
+
+  [:slug, :title, :summary, :valid?, :errors].each do |method|
+    it "delegates #{method} to the section" do
+      expect(section).to receive(method)
+      subject.send(method)
+    end
+  end
```
floehopper: Minor/subjective: I'm not sure examples testing just delegation add much value.

spec/presenters/section_presenter_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'spec_helper'
+
+describe SectionPresenter do
+  let(:section) { double(:section, body: '', attributes: {}, attachments: []) }
```
floehopper: Minor: Unless there's a strong reason not to I'd be tempted to use a real `Section` instance instead of a mock.

spec/presenters/section_presenter_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    it "delegates #{method} to the section" do
+      expect(section).to receive(method)
+      subject.send(method)
+    end
+  end
+
+  describe '#body' do
+    context 'a section with a footnotes div' do
+      before do
+        allow(section).to receive(:body).and_return('<div class="footnotes">')
```
floehopper: Minor: Maybe this should include the closing `div` tag like the output...? I was a bit confused why it had been added. Alternatively if it really is being added and that's important behaviour, perhaps there should be a separate example.

spec/presenters/section_presenter_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+
+      it 'adds a title before the existing footnotes' do
+        expect(subject.body).to eq("<h2 id=\"footnotes\">Footnotes</h2><div class=\"footnotes\">\n</div>\n")
```
floehopper: Minor: I'd prefer to use `%{<h2 id="footnotes">Footnotes</h2><div class="footnotes">\n</div>\n}` here to avoid having to escape the double-quotes which I think makes the text less readable.

spec/presenters/section_presenter_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    end
+
+    context 'a section with Govspeak in the body' do
+      before do
+        govspeak = %{
+$I
+This is information
+$I}
```
floehopper: Minor: I think it might be better to use Rails `String#strip_heredoc` to avoid having to have the unindented string literal.

app/presenters/section_presenter.rb
```diff
@@ -14,14 +14,12 @@ def initialize(section)
...
+    original_section = @section
 
-    pipeline.reduce(@section) { |current_section, next_renderer|
-      next_renderer.call(current_section)
-    }.attributes.fetch(:body)
+    processed_section_1 = MarkdownAttachmentProcessor.method(:new).call(original_section)
+    processed_section_2 = GovspeakToHTMLRenderer.create.call(processed_section_1)
+    processed_section_3 = FootnotesSectionHeadingRenderer.create.call(processed_section_2)
+
+    processed_section_3.attributes.fetch(:body)
```
floehopper: Yes! Thrice this! 👍

app/presenters/section_presenter.rb
```diff
@@ -29,8 +29,9 @@ def render_govspeak(body)
...
   def render_footnotes_heading(body)
-    FootnotesSectionHeadingRenderer.create.call(
-      OpenStruct.new(body: body)
-    ).body
+    footnote_open_tag = '<div class="footnotes">'
+    heading_tag = '<h2 id="footnotes">Footnotes</h2>'
+
+    body.gsub(footnote_open_tag, "#{heading_tag}\\&")
   end
 end
```
floehopper: Minor: I'd be tempted to squash this commit with the subsequent one that removes the `FootnotesSectionHeadingRenderer` and its spec.

spec/markdown_attachment_processor_spec.rb
```diff
@@ -83,24 +83,4 @@
...
-      }
-    }
-
-    it "merges the rendered body into the attributes" do
-      expect(renderer.attributes).to eq(
-        body: processed_body,
-        other_stuff: "other stuff",
-      )
-    end
-  end
```
floehopper: Minor: You could consider squashing this into the commit where it stopped being used.

spec/presenters/section_presenter_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe SectionPresenter do
-  let(:section) { double(:section, body: '', attributes: {}, attachments: []) }
+  let(:section) { double(:section, body: '') }
```
floehopper: Minor: Ideally squash this into the commit where they became unused.

chrislo:
I've addressed the feedback, thanks @floehopper. I've forced pushed before merging.

PR #1077, opened 2017-05-08
Extract version_type methods for Manual and Section

floehopper:
Extracting `Manual#version_type` and `Section#version_type` from `ManualPublishingAPIExporter` and `SectionPublishingAPIExporter` respectively makes the logic more explicit/intention-revealing and the combinatorial explosion of examples in the specs for these classes is somewhat reduced. These changes should help with the larger goal of [extracting API adapters from exporters][1].

In particular, this PR introduces the idea of a distinct version type for a new document as opposed to a major or minor version. This means we can make the logic which sets the Publishing API update type to "major" for a new version, even if it is marked as a minor update, more explicit.

Ideally it probably should be *possible* to have new/first version which is also marked as a minor update, i.e. I think some of the `Section` or `SectionEdition` state (flags, etc) could be combined to avoid non-allowed states existing in the first place.


floehopper:

floehopper:

PR #1081, opened 2017-05-09
Rename methods and instance variables calling section uuid

chrislo:
In #1017 we want to store `Section`s in Mongo instead of having an in-memory representation. To make this refactoring easier we are renaming `Section#id` to `Section#uuid`. This PR completes the rename by using `uuid` consistently in method and variable names that previously referred to `id`.

The only places where I haven't renamed are:
- In old migrations
- In the parameters used in the `section` routes (which in turn are passed in the `params` hash to the Service objects). 

@floehopper - what do you think about this last omission? I think I could do something with `routes.rb` to rename this parameter, but I'm not sure it's worth the effort given that we may rename back to `id` from `uuid` once the conversion is complete.

### DIFF
--- a/spec/exporters/publishing_api_publisher_spec.rb
+++ b/spec/exporters/publishing_api_publisher_spec.rb
@@ -4,8 +4,8 @@ require "publishing_api_publisher"
 
 describe PublishingAPIPublisher do
   let(:publishing_api) { double(:publishing_api, publish: nil) }
-  let(:section_id) { "12345678-9abc-def0-1234-56789abcdef0" }
-  let(:section) { double(:section, id: section_id) }
+  let(:section_uuid) { "12345678-9abc-def0-1234-56789abcdef0" }
+  let(:section) { double(:section, id: section_uuid) }
 
   it "raises an argument error if update_type is supplied, but not a valid choice" do
     expect {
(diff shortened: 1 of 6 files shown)


app/exporters/section_publishing_api_exporter.rb
```diff
@@ -13,17 +13,13 @@ def initialize(organisation, manual, section, update_type: nil)
...
   end
 
 private
 
   attr_reader :organisation, :manual, :section
 
-  def content_id
-    section.uuid
-  end
-
```
floehopper: Minor: This seems a bit orthogonal to the purpose of the PR. Also I'm doing some significant refactoring in this area and so it might make life easier from a merging point of view if this commit was left out.

spec/exporters/publishing_api_publisher_spec.rb
```diff
@@ -66,7 +66,7 @@
       it "asks the publishing api to publish the section with the specific update_type" do
         subject.call
 
-        expect(publishing_api).to have_received(:publish).with(section_id, "republish")
+        expect(publishing_api).to have_received(:publish).with(section_uuid, "republish")
       end
     end
   end
```
floehopper: Minor: I think it would make things clearer if you distinguish this commit from the other 2 by appending "in `PublishingAPIPublisher`" to the subject line. Ditto for the other three commits.

chrislo:
Force pushing before merging

chrislo:
I removed the commit you suggested and improved the commit message on the other three. Could you re-approve if you're happy @floehopper?

PR #1084, opened 2017-05-09
Introduce PublishingAdapter and inline some exporters into it

floehopper:

* Introduced `Adapters` module to act as a global registry of adapters, c.f. `Services`.
* Introduced `PublishingAdapter` with `#save` and `#save_section` methods incorporating logic from the following exporters which have been simplified, de-duped and/or in-lined:
  * `PublishingApiDraftManualWithSectionsExporter`
  * `PublishingApiDraftManualExporter `
  * `PublishingApiDraftSectionExporter `
  * `ManualPublishingAPILinksExporter`
  * `SectionPublishingAPILinksExporter `
* The only classes left in `app/exporters` with an `Exporter` suffix are:
  * `ManualPublishingAPIExporter`
  * `SectionPublishingAPIExporter`
* I plan to deal with these in a separate pull request, possibly in-lining them into `PublishingAdapter` too.
* The other classes in  `app/exporters` deal with things like publishing, withdrawing & redirecting which will be extracted into new methods on `PublishingAdapter` in other pull requests.

While there's clearly more work to do on this front, I think this is still a step forward in its own right. I'm sorry it's quite a big PR, but I wanted to get confidence I was heading in the right direction. Hopefully it won't be _too bad_ to review!

Note that I've introduced a fair bit of duplication by writing the `PublishingAdapter` spec as more of an integration test, because it overlaps substantially with the specs for `ManualPublishingAPIExporter` and `SectionPublishingAPIExporter`, the two remaining exporters. I hope that we'll be able to remove these specs in the not too distant future.


floehopper:
@chrislo: Thanks for the review. I have addressed your feedback, rebased against master and force-pushing in preparation for merging.

floehopper:
@chrislo:

> - squash d98db into da50b9

Addressed in 3fc019dd3762c68e6967a9063d3eb6cd04063f13.

> - reword commit message for e26f098 as it's not really a straight "move" of the spec.

Addressed in 1a72a32ee175efc30029a60e663855f22e158613.

> - the method signature of `PublishingAdapter#save` got a little complicated. It might help to add some helper methods on the class, e.g. `#save_with_links`.

I've decided not to do this for now. I have some hope that we may be able to do away with some of these options and so I'm less worried about the number of options.

PR #1085, opened 2017-05-10
Inline PublishingAPIRedirecter into PublishingAdapter#redirect_section

floehopper:
This PR is based on top of #1084 and continues the work to [extract API adapters from exporters][1].

It would probably be sensible to get #1084 merged before reviewing this, so the large number of changes in the first "squash" commit do not show up here.


floehopper:

spec/adapters/publishing_adapter_spec.rb
```diff
@@ -537,5 +537,14 @@
 
       subject.redirect_section(section, to: "/new-location")
     end
+
+    it "redirects section via Publishing API with valid attributes" do
+      expect(publishing_api).to receive(:put_content).with(
+        redirect_content_id,
+        be_valid_against_schema("redirect")
```
chrislo: This reads odd to me compared to the version it was moved from. Does `be_valid_against_schema` return the JSON (which happens to be truthy and therefore works as an expectation in the original spec)? I think it would be useful to split those two concerns into two separate expectations.

floehopper: Having chatted with Chris about this, I've got a couple of ideas on how to improve matters, but I'm going to merge this as stands and then look at making an improvement to the three places where this type of matcher is being used as an argument matcher in a separate PR.

PR #1086, opened 2017-05-10
Remove redundant require statement from Services module

floehopper:
This should have been removed as part of [this commit][1].

[1]: https://github.com/alphagov/manuals-publisher/commit/98683b06bb3ea4aadb26d1156fbbc10178ca1b35


floehopper:

PR #1095, opened 2017-05-15
Inline Publishing Api publishers into PublishingAdapter#publish

floehopper:

I've also done some refactoring to make `PublishingAdapter#save` and `PublishingAdapter#publish` as similar as possible to each other.


floehopper:
Rebased against master, resolved conflicts, and force pushed. @chrislo: Please could you re-approve?

PR #1097, opened 2017-05-15
Avoid unnecessary calls to Time#iso8601

floehopper:
There's no need to convert instances of `Time` into an ISO8601 string when we're sending a Hash to `GdsApi::PublishingApiV2`, because `gds-api-adapters` [does this in `GdsApi::JsonClient#do_json_request`][1] when `Object#to_json` gets called on the `params` Hash passed in.

```
> time = Time.now
=> 2017-05-15 18:48:14 +0100
> JSON.parse({test: time}.to_json)['test'] == time.iso8601
=> true
```

Note that `JSON.parse` does not reverse this conversion, so I've left the relevant test assertions in `ManualHelpers` unchanged. I've also left `DateTimeHelper#nice_time_format`, because that is unrelated to exporting to the Publishing API.

[1]: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/json_client.rb#L137

### DIFF
--- a/app/exporters/manual_publishing_api_exporter.rb
+++ b/app/exporters/manual_publishing_api_exporter.rb
@@ -62,8 +62,8 @@ private
   def optional_exportable_attributes
     attrs = {}
     if manual.originally_published_at.present?
-      attrs[:first_published_at] = manual.originally_published_at.iso8601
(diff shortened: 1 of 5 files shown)


floehopper:

floehopper:
@chrisroos: Please can you re-approve?

PR #1099, opened 2017-05-16
Fix section publishing when relocating a manual

suzannehamilton:
When moving a manual from one URL to another, ensure that its sections are properly created as drafts with the correct slug and then published.

Previously, the section publishing step was failing because the section drafts had not been created. This was because the `PublishingAdapter` only creates drafts for sections if the `republish` flag is added or if the section has never been exported. This has been fixed by adding the `republish` flag.

The tests did not catch the problem because the sections in the test data had no export date, so they *were* being created as drafts and republished. This has been fixed by adding export dates to the sections in the test.

(Side-note: I've made this change so I can move a manual that's been renamed to fix a 2ndline ZenDesk ticket. I'm not completely certain about whether adding `republish: true` flag is the correct thing to do, so feedback on that would be good.)

### DIFF
--- a/lib/manual_relocator.rb
+++ b/lib/manual_relocator.rb
@@ -195,11 +195,8 @@ private
   end
 
   def send_draft(manual)
-    puts "Sending a draft of manual #{manual.id} (version: #{manual.version_number})"
-    manual.sections.each do |section|
-      puts "Sending a draft of manual section #{section.uuid} (version: #{section.version_number})"
-    end
-    Adapters.publishing.save(manual, include_links: false)
+    puts "Sending a draft of manual #{manual.id} (version: #{manual.version_number}) and its sections"
+    Adapters.publishing.save(manual, include_links: false, republish: true)
   end
 
   def send_gone(section_uuid, slug)
--- a/lib/manual_relocator.rb
+++ b/lib/manual_relocator.rb
@@ -160,8 +160,8 @@ private
 
     # Clean up manual sections belonging to the temporary manual path
     new_section_uuids.each do |section_uuid|
-      puts "Redirecting #{section_uuid} to '/#{to_slug}'"
       most_recent_edition = most_recent_edition_of_section(section_uuid)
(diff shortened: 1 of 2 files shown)


floehopper:
@suzannehamilton:

This makes sense to me in the context of the refactoring @chrisroos, @chrislo & I have been doing. Can I just double-check that I've understood correctly...?

1. Previously the spec wasn't properly representative of the production data, because (in the scenario we're interested in) the sections would've already been exported (i.e. saved to the content store via the Publishing API).

2. However, when you fix the data by adding an `exported_at` to the sections, the spec started failing, because the sections were no longer exported.

3. You then fixed the spec by adding the `republish: true` option to the call to `PublishingAdapter#save` which forces sections to be exported, even if they've been exported before.

Does that all sound correct?

Note that I haven't added this as a review, because I didn't want to step on anyone's toes, but let me know if you'd like me to review it.

floehopper:
It might be worth asking @steventux or @h-lame to have a look, because I think they know more of the history of the `ManualRelocator`.

suzannehamilton:
Yes, that's all correct. 👍

floehopper:
@h-lame: Thanks. That makes sense to me. I think I introduced the problem in [this commit](https://github.com/alphagov/manuals-publisher/commit/40f27b9fff76cd40154938dc84fe3c1c3f2e7875), because I relied too heavily on the test coverage from the `ManualRelocator` spec.

@suzannehamilton: I'm guessing [these exceptions](https://errbit.integration.publishing.service.gov.uk/apps/57dfff6a65786363d4270100/problems/59198e326578630e8b510000) might have been you testing something to do with this PR. If so, please could you add a comment in Errbit and ideally mark them as resolved? We're trying to keep on top of the exceptions for this app in case we accidentally break something while refactoring.

suzannehamilton:
@floehopper Yep, those look like errors from the rake task I ran. I've added a comment.

PR #1100, opened 2017-05-16
Inline SectionPublishingAPIExporter into PublishingAdapter

floehopper:

Note that this results in `PublishingAdapter#save_section_content` becoming a very big method. However, I think it's a lot clearer what's going on and I'm hopeful that once I've also inlined `ManualPublishingAPIExporter`, I'll be able to extract some useful abstractions which will make the code terser.


floehopper:

floehopper:
Splitting a commit into two, rebasing against master and force-pushing.

chrisroos:
This looks good to me, @floehopper.

The only very, very minor thing I spotted was in a commit message. The commit with subject, "Use attribute reader methods vs fetching from attributes Hash" contains "and a `Section` never has a `content_id`" where I think it should be "and an `Attachment` never has a `content_id`".

floehopper:
@chrisroos: Thanks.

Rebased against master, fixed commit message, and force-pushed in preparation for merging.

Issue #1051, opened 2017-05-03
Upgrade to Rails 5.0


chrislo:
This was mentioned by @boffbowsh during our meeting today.

See https://docs.publishing.service.gov.uk/manual/upgrade-rails.html

floehopper:
A number of other GOV.UK apps have recently been upgraded to Rails v5. It might be worth looking at the relevant PRs to avoid reinventing the wheel, e.g. https://github.com/alphagov/hmrc-manuals-api/pull/174.

floehopper:
Also see https://github.com/alphagov/router-api/pull/106.

PR #1102, opened 2017-05-16
Upgrade to rails 4

chrisroos:
This branch upgrades the app to Rails 4.2.8 and Mongoid 4.

We used `rake rails:upgrade` to generate the updated Rails config files. We've committed each of the updates separately so that we can explain any important changes.

This is a step toward PR #1051.


chrisroos:
Thanks for reviewing, @floehopper. I'm going to merge this now.

chrisroos:
> re: StateMachine Gem - I think https://github.com/state-machines/state_machines is a drop-in replacement with a compatible API. I think I've used it on other projects when doing upgrades.

Thanks @h-lame. I've captured this in issue #1103.

PR #1104, opened 2017-05-17
Fix mongoid.yml for Mongoid 4

chrisroos:
The automatic deploy to integration failed after merging PR #1102 (upgrade to Rails 4/Mongoid 4). This was caused by some invalid options in the production config of mongoid.yml.

I've deployed this branch to integration and can confirm that it both deploys successfully and that the app works as expected.

### DIFF
--- a/config/mongoid.yml
+++ b/config/mongoid.yml
@@ -110,7 +110,7 @@ production:
       options:
         logger: false
         read: :primary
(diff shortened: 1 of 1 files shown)


chrisroos:
Thanks @floehopper and @chrislo!

I'll get it merged. And apologies for breaking things in the first place!

PR #1105, opened 2017-05-17
Inline ManualPublishingAPIExporter into PublishingAdapter

floehopper:
This continues the work to [extract API adapters from exporters][1]. It is very similar to #1100 which did the same thing for `SectionPublishingAPIExporter`.

Note that this results in `PublishingAdapter#save_manual_content ` becoming a very big method. However, I think it's a lot clearer what's going on and I'm hopeful that we'll able to extract some useful abstractions which will make the code terser in a future PR.


floehopper:

PR #1107, opened 2017-05-17
Improvements to delete_draft_manual Rake task

floehopper:
This fixes a problem with the Rake task, corrects its feature description, and makes some improvements to the Rake task and the underlying `CliManualDeleter` class.


floehopper:
@chrislo: Thanks.

Merging.

PR #1108, opened 2017-05-17
Rename block args in PublicationLogger#call

chrisroos:


chrisroos:
Thanks @floehopper. I've rebased and force-pushed in preparation for merging.

PR #1109, opened 2017-05-17
Fix JS toggle for hiding/showing the change note field

chrisroos:
I presume this was broken when we renamed document to section.

I've updated the selectors and checked that the toggling works as
expected in the browser. I've added a couple of view specs in the hope
that they'll help us catch these sort of problems earlier in future.


chrisroos:
Thanks @floehopper. I've rebased on master and force pushed.

PR #1110, opened 2017-05-17
Use PublishingAdapter in CliManualDeleter instead of using Publishing API directly

floehopper:
This is closely related to the work in #996.

This means the code can operate at a higher level of abstraction, passing instances of `Manual` rather than just some magic IDs. In doing this, I've extracted `PublishingAdapter#discard` which calls the already existing `PublishingAdapter#discard_section`.

In doing this I've also improved the overall level of abstraction used in this class, by using instances of `Manual` as much as possible rather than instances of `ManualRecord`. In doing so, I've extracted `Manual#destroy` which destroys the underlying `ManualRecord` and all associated `SectionEditions`.

### DIFF
--- a/app/adapters/publishing_adapter.rb
+++ b/app/adapters/publishing_adapter.rb
@@ -34,6 +34,13 @@ class PublishingAdapter
     end
   end
 
+  def discard(manual)
+    manual.sections.each do |section|
+      discard_section(section)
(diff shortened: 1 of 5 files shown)


floehopper:
@chrislo: Thanks!

Merging.

PR #1113, opened 2017-05-18
Exclude arbitrary attrs when copying SectionEditions

chrisroos:
This should hopefully avoid the following errors when creating new
drafts of SectionEditions that contain old data:

> Mongoid::Errors::UnknownAttribute: Problem: Attempted to set a value
> for '<old-field-name>' which is not allowed on the model SectionEdition.
> Summary: Without including Mongoid::Attributes::Dynamic in your model
> and the attribute does not already exist in the attributes hash,
> attempting to call SectionEdition#old-field-name= for it is not
> allowed. This is also triggered by passing the attribute to any method
> that accepts an attributes hash, and is raised instead of getting a
> NoMethodError. Resolution: You can include
> Mongoid::Attributes::Dynamic if you expect to be writing values for
> undefined fields often.

We've been seeing this error since the upgrade to Mongoid 4 where the default
behaviour of allowing dynamic fields has changed. In Mongoid 3 the
behaviour was controlled by `Mongoid.allow_dynamic_fields` which was set
to true by default. In Mongoid 4 it's controlled at the model level by
including the `Mongoid::Attributes::Dynamic` module.

I think the test I've added could be better but I think it's good enough
for now. I have a separate branch where I'm refactoring Section and I
hope to improve the tests in there.

Note the call to `#with_indifferent_access` in the setup for the test.
This better mirrors the behaviour of `SectionEdition#attributes`.

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -244,8 +244,7 @@ protected
 
   def previous_edition_attributes
     latest_edition.attributes
-      .slice(:_id, :section_uuid, :version_number, :title, :slug, :summary, :body, :state, :change_note, :minor_update, :public_updated_at, :exported_at, :created_at, :updated_at)
-      .except(*no_copy_attributes)
+      .slice(:section_uuid, :version_number, :title, :slug, :summary, :body, :state, :change_note, :minor_update, :public_updated_at)
       .symbolize_keys
   end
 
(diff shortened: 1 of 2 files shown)


chrisroos:
> [...] and maybe add migrations to remove any irrelevant data from mongodb (even if that's in separate PRs).

I've created issue #1115 to capture the idea of removing the stale data.

chrisroos:
> And the public_updated_at field is in the included field list, but not in mongodb.

I'll investigate removing this in my separate refactoring branch.

chrisroos:
Thanks for reviewing, @floehopper.

I've updated the first commit and message and added a second commit in the hope that it's easier to understand what's going on. Can you take another look, please?

chrisroos:
Thanks @floehopper! I'll merge it now.

PR #1117, opened 2017-05-18
Refactor Section

chrisroos:
I think/hope these changes make it slightly easier to see the behaviour of the Section class.

I still think there's some confusion around the role of `editions` and `latest_edition` in the class but I plan to address that separately.

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -76,9 +76,9 @@ class Section
         .symbolize_keys
 
       attributes = previous_edition_attributes
-        .merge(state: "draft")
         .merge(params)
         .merge(
+          state: 'draft',
           version_number: latest_edition.version_number + 1,
           slug: slug,
           attachments: attachments,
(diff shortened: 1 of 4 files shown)


app/models/section.rb
```diff
@@ -71,9 +71,7 @@ def to_param
   end
 
   def update(params)
-    allowed_update_params = self.class.edition_attributes
-      .-([:updated_at, :slug, :version_number])
-      .map(&:to_s)
+    allowed_update_params = [:title, :summary, :body, :change_note, :minor_update].map(&:_to_s)
```
chrislo: Should this be calling `_to_s` or `to_s` as in the code before this commit?

chrisroos: Good spot! That's definitely a typo. I'll fix the commit.

chrisroos:
Thanks for reviewing @chrislo. I've addressed your feedback, rebased on master and force pushed in preparation for merging.

Issue #1106, opened 2017-05-17
Make Service initializers more explicit


chrislo:
We currently pass the entire controller `context` into the initializer of each `Service`. This makes it a bit unclear what information the service needs (and also causes us to create some dummy "context" objects in places where services are called outside of controllers). Passing the data in as named arguments should help with this.

PR #1118, opened 2017-05-18
Parameterize services

chrislo:
Before this PR, all of the Service objects took a `context` parameter in the initializer. This made it difficult to see what data each of the services required in order to work, and also meant that when a Service was called from outside of a Controller we had to create a `context` object that behaved like a Controller instance. 

I've expanded the `context` argument to named arguments in all cases. There are some Services which deal with the parameters generated from a Rails form submission - I've left these as hash values as I think this is conventional (and less brittle in case objects change). 

There are a few instances where making this change reveals that some redundant information is being passed into the Service. I've left these for now, but hopefully this change will make it easier to refactor/remove later. 

Fixes: #1106

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -129,7 +129,7 @@ class ManualsController < ApplicationController
     service = Manual::PreviewService.new(
       manual_id: params[:id],
       attributes: update_manual_params,
-      context: self,
+      user: current_user,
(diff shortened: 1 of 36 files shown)


chrislo:
Thanks @floehopper! I'll make the changes you suggest in a separate PR, they all sound sensible to me. 

Force pushing before merging.

PR #1119, opened 2017-05-18
Use Manual vs ManualRecord in AttachmentReporting

floehopper:
`ManualRecord` should be an implementation detail of `Manual`. Ideally classes other than `Manual` shouldn't access `ManualRecord` directly.

I've also removed a comment which has been rendered obsolete by these changes.

### DIFF
--- a/lib/attachment_reporting.rb
+++ b/lib/attachment_reporting.rb
@@ -8,25 +8,19 @@ class AttachmentReporting
   end
 
   def create_organisation_attachment_count_hash
-    manual_records = ManualRecord.all
-    unique_owning_organisation_slugs = manual_records.map(&:organisation_slug).uniq
+    manuals = Manual.all(User.gds_editor).to_a
+    unique_owning_organisation_slugs = manuals.map(&:organisation_slug).uniq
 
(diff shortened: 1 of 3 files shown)


lib/attachment_reporting.rb
```diff
@@ -8,25 +8,19 @@ def initialize(first_period_start_date, last_time_period_days, attachment_file_e
...
-      # Rather than examine each manual edition and its set of section editions and attachments in turn,
-      # we instead get all unique section ids associated with this manual, then walk through
-      # the editions of these sections in version order to find unique PDF attachments and their
-      # publication times.
-      all_unique_section_uuids_for_manual(manual).each do |section_uuid|
-        section_editions = SectionEdition.all_for_section(section_uuid).order_by([:version_number, :asc])
-
-        section_editions.each do |section_edition|
+      manual.sections.each do |section|
+        section.editions.sort_by(&:version_number).each do |section_edition|
```
chrisroos: I don't know whether it's important but `section.editions` will only contain the latest **two** editions as found by `Section.find`. This is different to the previous behaviour that used `SectionEdition.all_for_section` to return all editions for a section.

floehopper:

PR #1122, opened 2017-05-22
Improve consistency of controllers/services

chrislo:
This PR addresses most of the feedback to PR #1118. Specifically:

> Using symbols as keys for controller parameters is more idiomatic Rails; we're still using strings in some places.

Addressed in 7f45f72e

> I think it would make the division between controller and service clearer if any relevant constructor arguments were suffixed with `_attributes` vs `_params`.

Addressed in 4ac9889, 0a24416, 5e584e7

> If a constructor argument is a `Hash` of the attributes associated with the model that is the main subject of the service, I probably wouldn't include the model name in the argument name, e.g. in an `Attachment` service I'd use `attributes` vs `attachment_attributes` as the constructor argument name.

Addressed in 4ac9889

> It would be nice if all the service constructor arguments had a consistent ordering, e.g. `user` always comes first or last; and `manual_id`, `section_uuid` & `attachment_id` are always in that order.

Addressed in 5025380

> You mention in the PR description that you've kept some form hashes intact and passed them in without splitting them into separate explicit arguments. I think this is a good idea. However, it looks as if in some places you've split them out, e.g. `file` & `title` in `Attachment::CreateService#initialize`. I'm not sure I'm so keen on that.

Addressed in 0a24416, 5e584e7

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -41,8 +41,8 @@ class ManualsController < ApplicationController
 
   def create
     service = Manual::CreateService.new(
-      attributes: create_manual_params,
       user: current_user,
+      attributes: create_manual_params
     )
     manual = service.call
     manual = manual_form(manual)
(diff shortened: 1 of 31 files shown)


chrislo:
I'm going to fix the conflicts, rebase and force push.

PR #1123, opened 2017-05-22
Extract Manual.find_by_slug! and use Manual vs ManualRecord in a bunch of rake tasks

floehopper:
This started off as an attempt to reduce duplication where a `Manual` was being looked up by its `slug` via a `ManualRecord` query. However, it has also led me into using `Manual` vs `ManualRecord` in a bunch of Rake tasks and extracting `Manual#set` & `Manual#editions`.

The higher level goal of preferring `Manual` over `ManualRecord` is that the latter should be an implementation detail of the former and ideally shouldn't be accessed from outside the former.

### DIFF
--- a/spec/models/manual_spec.rb
+++ b/spec/models/manual_spec.rb
@@ -1074,4 +1074,13 @@ describe Manual do
       expect(manual.editions).to eq(manual_record.editions)
     end
   end
+
+  describe "#set" do
+    let!(:manual_record) { FactoryGirl.create(:manual_record, manual_id: manual.id) }
+
+    it "sets attributes on underlying manual record" do
+      manual.set(slug: "new-slug")
+      expect(manual_record.reload.slug).to eq("new-slug")
+    end
+  end
 end
--- a/spec/models/manual_spec.rb
+++ b/spec/models/manual_spec.rb
@@ -1066,4 +1066,12 @@ describe Manual do
       end
     end
   end
+
+  describe "#editions" do
+    let!(:manual_record) { FactoryGirl.create(:manual_record, manual_id: manual.id) }
+
+    it "returns editions from underlying manual record" do
+      expect(manual.editions).to eq(manual_record.editions)
+    end
+  end
 end
--- a/spec/models/manual_spec.rb
+++ b/spec/models/manual_spec.rb
@@ -1016,4 +1016,54 @@ describe Manual do
       end
     end
   end
+
+  describe ".find_by_slug!" do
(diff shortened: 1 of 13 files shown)


spec/models/manual_spec.rb
```diff
@@ -20,15 +20,15 @@
...
-  let(:originally_published_at) { double(:originally_published_at) }
-  let(:use_originally_published_at_for_public_timestamp) { double(:use_originally_published_at_for_public_timestamp) }
-  let(:title) { double(:title) }
-  let(:summary) { double(:summary) }
-  let(:body) { double(:body) }
-  let(:organisation_slug) { double(:organisation_slug) }
-  let(:state) { double(:state) }
-  let(:slug) { double(:slug) }
+  let(:updated_at) { Time.parse("2001-01-01") }
+  let(:originally_published_at) { DateTime.parse("2002-02-02") }
```
chrisroos: Is there a reason that `updated_at` is a `Time` and `originally_published_at` is a `DateTime`?

chrisroos:
I notice a number of commits don't contain any explanation (e.g. "Extract Manual.find_by_slug!" and "Introduce Manual#editions"). I can see that an explanation might not be necessary while looking at the commits in the context of the branch but I can imagine that it might be useful when coming back to these commits in future. What do you think?

chrisroos:
The commit titled "Use a real user vs OpenStruct in SectionReslugger spec" changes the SectionReslugger rather than its spec. It might be worth updating the commit note.

spec/models/manual_spec.rb
floehopper: [`ManualRecord::Edition#updated_at` is defined by `Mongoid::Timestamps`](https://github.com/alphagov/manuals-publisher/blob/56aae0810219bbc119c28698a8b4663fc050672b/app/models/manual_record.rb#L71) whereas [`originally_published_at` is has a custom definition as `DateTime`](https://github.com/alphagov/manuals-publisher/blob/56aae0810219bbc119c28698a8b4663fc050672b/app/models/manual_record.rb#L80). My guess is the developer either didn't realise they were different or had a specific reason to make them different.

I consciously chose not to address this problem in this PR, but I have have just captured it in #1126.

floehopper:

floehopper:
@chrisroos: Thanks for the feedback. I've fixed the one incorrect commit note and elaborated on my motivations in a bunch of others. I've force-pushed my changes and plan to merge this if/when the CI build passes.

PR #1125, opened 2017-05-22
Inline Manual.build into Manual#initialize

floehopper:
This makes `Manual` more like a standard `ActiveModel`-type class which should make it easier to convert into a `Mongoid::Document`.

This change was made easier by the extraction of a factory for `Manual`. The factory is a bit non-standard, because `Manual` does not implement attribute writer methods which is what `FactoryGirl` uses by default to set the attributes.

There is also a small change in the behaviour of `Manual#new`. Previously if no ID was supplied, a `KeyError` would be raised; whereas now an ID is generated. As I explain in more detail in the commit note, this is closer to standard `Mongoid::Document` behaviour an so I'm happy to make the change.

I'm sure there is further simplification we could do to `Manual#initialize`, but I think this is a sufficient step forward to merit a PR.

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -54,7 +54,7 @@ FactoryGirl.define do
     slug "manual-slug"
 
     initialize_with do
-      Manual.build(attributes)
+      Manual.new(attributes)
     end
   end
 
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -50,6 +50,14 @@ FactoryGirl.define do
     sequence(:change_note) { |n| "Change note #{n}" }
   end
 
+  factory :manual do
+    slug "manual-slug"
+
+    initialize_with do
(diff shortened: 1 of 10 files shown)


floehopper:

spec/factories.rb
```diff
@@ -50,6 +50,14 @@
     sequence(:change_note) { |n| "Change note #{n}" }
   end
 
+  factory :manual do
+    slug "manual-slug"
+
+    initialize_with do
```
chrisroos: Can this be removed now that it's using `.new` instead of `.build`?

floehopper: No. As I mentioned in the 2nd commit note:

> The factory is a bit non-standard, because `Manual` does not implement
attribute writer methods which is what `FactoryGirl` uses by default to
set the attributes.

PR #1127, opened 2017-05-23
Upgrade to Mongoid 5

chrisroos:
I updated the Mongoid version constraint in the Gemfile and ran `bundle
update mongoid --conservative` to upgrade the minimum set of
dependencies.

I regenerated the mongoid.yml config file by using `rails g
mongoid:config`. I accepted all the changes and then replaced the
`database` and `hosts` options with the `uri` option that we were using
previously.

Note the removal of the `refresh_interval` option from our production
session config. This option was specific to Moped (Mongoid 5 uses the
Mongo Ruby driver instead of Moped) and I've been unable to find an
equivalent option in the new driver. I've looked at a sample of
mongoid.yml config files within the alphagov organisation[1] and can't
see any that are setting an equivalent option so I'm happy to simply
remove this.

Note that Mongoid 5 drops support for MongoDB 2.2 and now only supports
2.4 and higher[2]. This shouldn't be a problem as we're using 2.4.9 on
our servers[3].

I've deployed this branch to integration and everything appears to work
as expected.

[1]:
https://github.com/search?q=org%3Aalphagov+filename%3Amongoid.yml&type=Code&utf8=%E2%9C%93
[2]:
https://docs.mongodb.com/ruby-driver/master/tutorials/5.2.0/mongoid-upgrade/
[3]:
https://github.com/alphagov/govuk-puppet/blob/f3614e33bcf037b218e0b9e816f0994786b41efb/hieradata/common.yaml#L1256


chrisroos:
Thanks @chrislo.

> I looked at the search results for the max_pool_size option and it seems that a couple of other apps use the default value of 1 too.

Thanks for checking. I've updated the commit note to explain that this is the default option when generating the new config file.

I've rebased on master and will merge once the tests pass.

PR #1130, opened 2017-05-23
Avoid persisting new sections when saving to Publishing API fails

chrisroos:
This fixes a specific problem when creating multiple sections in a
manual with the same title. The sections would be saved to the local
database but would fail to save to Publishing API because the slug of
the section would already be in use.

section with a duplicate title. A further improvement will be to either
avoid trying to save the section in the first place (i.e. because we've
determined that it's a duplicate) or to handle the validation errors we
get back from the Publication API.

It's possible that this change will allow us to remove the following
classes that detect different sections sharing the same slug (i.e. the
same title):

* DuplicateDocumentFinder
* DuplicateDraftDeleter

### DIFF
--- a/spec/services/section/create_service_spec.rb
+++ b/spec/services/section/create_service_spec.rb
@@ -1,0 +1,96 @@
+require "spec_helper"
+
+RSpec.describe Section::CreateService do
+  let(:user) { User.gds_editor }
+  let(:manual) { Manual.new(title: 'manual-title') }
+  let(:section_attributes) { double(:section_attributes) }
+  let(:publishing_api_adapter) { double(:publishing_api) }
+  let(:new_section) {
+    Section.new(manual: manual, uuid: 'uuid', editions: [])
+  }
+
+  subject do
+    described_class.new(
+      user: user,
+      manual_id: manual.id,
+      attributes: section_attributes
+    )
+  end
+
+  before do
+    allow(Manual)
+      .to receive(:find).with(manual.id, user)
+      .and_return(manual)
+    allow(manual)
+      .to receive(:build_section).with(section_attributes)
+      .and_return(new_section)
+    allow(Adapters).to receive(:publishing)
+      .and_return(publishing_api_adapter)
+    allow(publishing_api_adapter).to receive(:save)
+    allow(publishing_api_adapter).to receive(:save_section)
+  end
+
+  context 'when the new section is valid' do
+    before do
+      allow(new_section).to receive(:valid?).and_return(true)
+    end
+
+    it 'marks the manual as draft' do
+      expect(manual).to receive(:draft)
+
+      subject.call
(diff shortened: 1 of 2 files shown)


spec/services/section/create_service_spec.rb
```diff
@@ -0,0 +1,96 @@
+require "spec_helper"
+
+RSpec.describe Section::CreateService do
+  let(:user) { User.gds_editor }
+  let(:manual) { Manual.new(title: 'manual-title') }
+  let(:section_attributes) { double(:section_attributes) }
```
floehopper: Minor: It feels a bit weird to me to use a mock object here.

spec/services/section/create_service_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    )
+  end
+
+  before do
+    allow(Manual)
+      .to receive(:find).with(manual.id, user)
+      .and_return(manual)
+    allow(manual)
+      .to receive(:build_section).with(section_attributes)
+      .and_return(new_section)
```
floehopper: Minor: This feels like a lot of stubbing. How hard would it be to rely on the real `Manual#build_section` method here?

chrisroos: I tried to use the real `Manual#build_section` method initially but ran into problems when trying to test the call to `Adapters.publishing.save_section` with the section that's returned from `build_section`. I'm going to leave this as it is at the moment and see whether I come up with anything better while working on the rest of the services.

spec/services/section/create_service_spec.rb
chrisroos: I agree that it's not ideal. I ended up going with a double so that I could be confident that the `attributes` passed to `Section::CreateService#initialize` was the same as the attributes passed to `Manual#build_section`.
I'm going to leave this as it is at the moment and see whether I come up with anything better while working on the rest of the services.

chrisroos:
> I think the `MarkedSectionDeleter` class might also be a candidate for removal.

Thanks. I've added this to the commit note.

chrisroos:
> This might be out-of-scope for this change, but is there a potential issue if the manual is successfully exported to the Publishing API, but the section is not? With the changes in this PR, we won't update anything in the local database, but the state of the manual will have changed in the Publishing API, i.e. the database and the Publishing API will be out of sync. Do we need to consider somehow rolling back the manual export step in this scenario?

I think the problem you've identified is real but I don't think we're any worse off than we were before: the database and Publishing API would have been out of sync prior to this change because the section would've been created locally but not in the Publishing API.

Rolling back might well be sensible but I'm not planning to do anything about it in the near future.

chrisroos:
Thanks for the review, @floehopper. I've addressed your comments, rebased on master and force pushed in preparation for merging.

PR #1131, opened 2017-05-23
Refactor Section::CreateService

chrisroos:


chrisroos:
Thanks @floehopper. Rebased on master and force pushed in preparation for merging.

PR #1133, opened 2017-05-23
Refactor Section::NewService

chrisroos:
I don't think this single line method was adding anything.


chrisroos:

PR #1134, opened 2017-05-23
Refactor Section::PreviewService

chrisroos:

### DIFF
--- a/app/services/section/preview_service.rb
+++ b/app/services/section/preview_service.rb
@@ -10,7 +10,7 @@ class Section::PreviewService
     section = if section_uuid
                 existing_section
               else
-                ephemeral_section
+                manual.build_section(attributes)
               end
     section.update(attributes)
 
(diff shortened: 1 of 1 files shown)


app/services/section/preview_service.rb
```diff
@@ -7,6 +7,14 @@ def initialize(user:, manual_id:, section_uuid:, attributes:)
...
 
   def call
+    manual = Manual.find(manual_id, user)
+    section = if section_uuid
+                manual.sections.find { |sec|
+                  sec.uuid == section_uuid
+                }
+              else
+                manual.build_section(attributes)
+              end
```
floehopper: Minor: I know Rubocop required it, but the hanging indent makes my eyes bleed. I think you could fix that by extracting a `Manual#find_section` method and converting back to a ternary expression; or even extract the whole thing into `Manual#find_or_build_section`.

chrisroos:

app/services/section/preview_service.rb
chrisroos: I'm going to leave this as it is for now. I'm hoping to introduce the `Manual#find_section` method soon at which point this'll probably/hopefully change.

PR #1135, opened 2017-05-23
Refactor Section::RemoveService

chrisroos:

### DIFF
--- a/app/services/section/remove_service.rb
+++ b/app/services/section/remove_service.rb
@@ -18,7 +18,7 @@ class Section::RemoveService
       manual.draft
 
       manual.remove_section(section_uuid)
-      persist
+      manual.save(user)
       export_draft_manual_to_publishing_api
       discard_section_via_publishing_api
     end
@@ -29,10 +29,6 @@ class Section::RemoveService
 private
 
   attr_reader :user, :manual_id, :section_uuid, :attributes
-
-  def persist
-    manual.save(user)
-  end
 
   def section
     @section ||= manual.sections.find { |s| s.uuid == section_uuid }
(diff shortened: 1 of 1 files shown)


app/services/section/remove_service.rb
```diff
@@ -9,18 +9,29 @@ def initialize(user:, manual_id:, section_uuid:, attributes:)
   end
 
   def call
+    begin
+      manual = Manual.find(manual_id, user)
+    rescue KeyError
+      raise ManualNotFoundError.new(manual_id)
```
floehopper: Possibly out-of-scope: I don't think `Manual.find` raises a `KeyError` any more; I think it raises a `Manual::NotFoundError` instead. Thus we could probably ditch this `rescue` block and rescue the latter exception in `ManualWithdrawer#execute`.

app/services/section/remove_service.rb
```diff
@@ -9,18 +9,29 @@ def initialize(user:, manual_id:, section_uuid:, attributes:)
...
+    end
+
+    section = manual.sections.find { |s| s.uuid == section_uuid }
```
floehopper: Minor: See my [earlier comment](https://github.com/alphagov/manuals-publisher/pull/1134#discussion_r118053067) about extracting a `Manual#find_section` method.

app/services/section/remove_service.rb
chrisroos: Ah, that's useful to know! I'm going to leave this for now as I can imagine addressing it as I continue to work through these services ensuring that the save to the publishing api happens before the save to our database.

app/services/section/remove_service.rb
chrisroos: I'm planning to introduce a `Manual#find_section` method outside of this PR.

chrisroos:

PR #1138, opened 2017-05-23
Refactor Section::UpdateService

chrisroos:

### DIFF
--- a/app/services/section/update_service.rb
+++ b/app/services/section/update_service.rb
@@ -25,5 +25,5 @@ class Section::UpdateService
 
 private
 
-  attr_reader :user, :manual_id, :section_uuid, :attributes, :listeners
+  attr_reader :user, :manual_id, :section_uuid, :attributes
 end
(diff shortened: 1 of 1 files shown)


app/services/section/update_service.rb
```diff
@@ -9,35 +9,21 @@ def initialize(user:, manual_id:, section_uuid:, attributes:)
   end
 
   def call
+    manual = Manual.find(manual_id, user)
+    section = manual.sections.find { |s| s.uuid == section_uuid }
```
floehopper: Minor: See my [earlier comment](https://github.com/alphagov/manuals-publisher/pull/1134#discussion_r118053067) about extracting a `Manual#find_section` method.

chrisroos:

app/services/section/update_service.rb
chrisroos: I plan to introduce a `Manual#find_section` method in a separate PR.

PR #1139, opened 2017-05-23
Add spec for Section::NewService

chrisroos:
This should make it easier to make changes to this service in future.

### DIFF
--- a/spec/services/section/new_service_spec.rb
+++ b/spec/services/section/new_service_spec.rb
@@ -1,0 +1,16 @@
+require "spec_helper"
+
+RSpec.describe Section::NewService do
+  it 'returns the manual and a new section' do
+    user = FactoryGirl.build(:user, organisation_slug: 'org-slug')
+    manual = FactoryGirl.build(:manual, organisation_slug: user.organisation_slug)
+    manual.save(user)
+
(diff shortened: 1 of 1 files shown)


spec/services/section/new_service_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "spec_helper"
+
+RSpec.describe Section::NewService do
+  it 'returns the manual and a new section' do
+    user = double(:user)
+    section = double(:section)
+    manual = double(:manual, build_section: section)
```
floehopper: Minor: Can we not use instances of real classes for `user`, `section` & `manual` here? Also how hard would it be to rely on the real implementation of `Manual#build_section`...?

chrisroos: I've added a second commit to this branch ("Use real objects in the test") that uses real objects instead of stubs. Can you take a look and let me know whether you think it's any better please, @floehopper?

chrisroos: Having thought about it some more I've decided that I prefer the updated tests that avoid using `double`s. I've updated this branch so that it contains a single commit introducing the more realistic tests.

PR #1140, opened 2017-05-23
Add spec for Section::ListService

chrisroos:
This should make it easier to make changes to this service in future.

### DIFF
--- a/spec/services/section/list_service_spec.rb
+++ b/spec/services/section/list_service_spec.rb
@@ -1,0 +1,18 @@
+require "spec_helper"
+
+RSpec.describe Section::ListService do
+  it 'returns the manual and its sections' do
+    user = FactoryGirl.build(:user, organisation_slug: 'org-slug')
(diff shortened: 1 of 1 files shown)


spec/services/section/list_service_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "spec_helper"
+
+RSpec.describe Section::ListService do
+  it 'returns the manual and its sections' do
+    user = double(:user)
+    sections = double(:sections)
+    manual = double(:manual, sections: sections)
```
floehopper: Minor: See [my earlier comment](https://github.com/alphagov/manuals-publisher/pull/1139#discussion_r118056507) about using too many mock objects.

chrisroos:
Thanks for the review, @floehopper. I've updated the tests so that they use realistic objects instead of `double`s (in the same way that I change the tests in PR #1139). I've rebased on master and force pushed in preparation for merging.

PR #1141, opened 2017-05-23
Avoid persisting updated sections when saving to Publishing API fails

chrisroos:
Avoid persisting updated section when saving to Publishing API fails

This fixes a specific problem when updating a section and giving it the
same title as another section in the same manual. The changed section
would be saved to the local database but would fail to save to
Publishing API because the slug of the section would already be in use.

section with a duplicate title. A further improvement will be to either
avoid trying to save the section in the first place (i.e. because we've
determined that the title is already in use) or to handle the validation
errors we get back from the Publication API.

This should get us closer to being able to remove the following classes
that detect different sections sharing the same slug (i.e. the same
title):

* DuplicateDocumentFinder
* DuplicateDraftDeleter

### DIFF
--- a/app/services/section/update_service.rb
+++ b/app/services/section/update_service.rb
@@ -15,9 +15,9 @@ class Section::UpdateService
 
     if section.valid?
       manual.draft
-      manual.save(user)
       Adapters.publishing.save(manual, include_sections: false)
       Adapters.publishing.save_section(section, manual)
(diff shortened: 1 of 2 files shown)


chrisroos:
Thanks for the review @floehopper. I think my responses to your feedback on PR #1130 apply to this branch too so I'm not repeating them here.

PR #1145, opened 2017-05-24
Upgrade to Rails 5.0

chrislo:
This PR upgrades the application to the latest patch release in the 5.0 Rails series (5.0.3). 

We tried to keep the set of gem changes required to upgrade the rails gem as small as possible by first making some gem upgrades that could be done against rails 4, and by using `--conservative` when running bundler in aa9f3c9. 

Note that moving to Rails 5 means we have also upgraded Mongoid to 6.1.0. This version of Mongoid still works with the version of mongodb we are using in production (2.4) and having tested in locally and on the Integration environment, I'm reasonably confident that no changes are required to the data or the mongo configuration. 

I've kept the "migration options to ease your Rails 5.0 upgrade" options in `config/initializers/new_framework_defaults.rb` unchanged for now. We should consider toggling them at some point (probably before we upgrade to Rails 5.1). This is captured in [this new issue](https://github.com/alphagov/manuals-publisher/issues/1147).

The rest of the changes are in order to suppress deprecation warnings raised when running the full test suite with`rake`. We've used the [hmrc-manuals-api](https://github.com/alphagov/hmrc-manuals-api/pull/174/) rails 5 upgrade PR for inspiration on how to fix some of these.

Fixes: #1051


config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
```
chrisroos: Do we need this file?

config/environments/development.rb
```diff
@@ -24,15 +39,13 @@
   # number of complex assets.
   config.assets.debug = true
 
-  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
-  # yet still be able to expire them through the digest params.
-  config.assets.digest = true
```
chrisroos: Is this no longer a valid option in Rails 5? If so I think it'd be good to explain that in the commit note.

config/environments/development.rb
```diff
@@ -24,15 +39,13 @@
...
-
-  # Adds additional error checking when serving assets at runtime.
-  # Checks for improperly declared sprockets dependencies.
-  # Raises helpful error messages.
-  config.assets.raise_runtime_errors = true
```
chrisroos: Is this no longer a valid option in Rails 5? If so I think it'd be good to explain that in the commit note.

config/environments/production.rb
```diff
@@ -49,16 +47,15 @@
   config.log_level = :debug
 
   # Prepend all log lines with the following tags.
-  # config.log_tags = [ :subdomain, :uuid ]
-
-  # Use a different logger for distributed setups.
-  # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
+  config.log_tags = [ :request_id ]
```
chrisroos: It'd be good to explain the addition of this config option in the commit note.

config/environments/test.rb
```diff
@@ -25,15 +27,13 @@
...
   config.action_controller.allow_forgery_protection = false
+  config.action_mailer.perform_caching = false
 
   # Tell Action Mailer not to deliver emails to the real world.
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
 
-  # Randomize the order test cases are executed.
-  config.active_support.test_order = :random
```
chrisroos: Is this no longer a valid option in Rails 5? If so I think it'd be good to explain that in the commit note.

chrislo:
Thanks @chrisroos - I've expanded on the explanation in the commit message that updates the config files. 

Rebasing and force pushing before merging.

PR #1148, opened 2017-05-24
Avoid persisting new manuals when saving to Publishing API fails

chrisroos:
This fixes a specific problem when creating a manual and giving it the
same title as another manual. The new manual would be saved to the
local database but would fail to save to Publishing API because the slug
of the manual would already be in use.

manual with a duplicate title. A further improvement will be to either
avoid trying to save the manual in the first place (i.e. because we've
determined that the title is already in use) or to handle the validation
errors we get back from the Publication API.

### DIFF
--- a/spec/services/manual/create_service_spec.rb
+++ b/spec/services/manual/create_service_spec.rb
@@ -1,0 +1,52 @@
+require "spec_helper"
+
+RSpec.describe Manual::CreateService do
+  let(:user) { double(:user) }
+  let(:manual) { double(:manual, valid?: nil, save: nil) }
+  let(:publishing_api_adapter) { double(:publishing_api_adapter, save: nil) }
+
+  subject do
+    described_class.new(
+      user: user,
+      attributes: {}
+    )
+  end
+
+  before do
+    allow(Manual).to receive(:new).and_return(manual)
+    allow(Adapters)
+      .to receive(:publishing).and_return(publishing_api_adapter)
+  end
+
+  context 'when the manual is valid' do
+    before do
+      allow(manual).to receive(:valid?).and_return(true)
+    end
+
+    it 'saves the manual' do
+      expect(manual).to receive(:save)
+      subject.call
+    end
+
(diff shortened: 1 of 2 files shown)


spec/services/manual/create_service_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      subject.call
+    end
+  end
+
+  context 'when the manual is not valid' do
+    before do
+      allow(manual).to receive(:valid?).and_return(false)
+    end
+
+    it 'saves the manual' do
```
floehopper: I think the example name should be the inverse, i.e. `it 'does not save the manual' do`.

spec/services/manual/create_service_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      expect(manual).to_not receive(:save)
+      subject.call
+    end
+
+    it 'saves the manual to the Publishing API' do
```
floehopper: I think the example name should be the inverse, i.e. `it 'does not save the manual to the Publishing API' do`.

spec/services/manual/create_service_spec.rb
```diff
@@ -34,6 +34,25 @@
     end
   end
 
+  context 'when the manual is valid but saving it to the publishing api fails' do
```
floehopper: Minor: I think we've been capitalising "Publishing API" in most places. It would be nice to be consistent.

chrisroos: Good spot. I've fixed this.

spec/services/manual/create_service_spec.rb
chrisroos: Good spot. I've fixed this.

spec/services/manual/create_service_spec.rb
chrisroos: Good spot. I've fixed this.

chrisroos:

PR #1150, opened 2017-05-24
Extract Manual#find_section

chrisroos:
We remove a certain amount of duplication by extracting this method and
using it instead of manually iterating over the `Manual#sections`.


chrisroos:
Thanks @chrislo. I've rebased on master and force pushed in preparation for merging.

PR #1151, opened 2017-05-25
Gemfile-related changes

floehopper:
* Avoid bundler warning
* Remove explicit dependencies on multi_json & rake gems
* Remove redundant faraday & test-unit gems
* Upgrade generic_form_builder & plek gems to latest version

There is still more work that can be done here, but at least this is a start.


floehopper:
Rebased and force-pushed in preparation for merging.

PR #1155, opened 2017-05-25
Avoid exposing editions from Section

chrisroos:
The `SectionEdition`s are an implementation detail of `Section` and shouldn't be accessed from outside that class. This branch removes uses of `Section#editions` and `Section#latest_edition` and makes these methods private.

### DIFF
--- a/spec/models/section_spec.rb
+++ b/spec/models/section_spec.rb
@@ -110,6 +110,21 @@ describe Section do
     allow(SlugGenerator).to receive(:new).with(prefix: manual_slug).and_return(slug_generator)
   end
 
+  describe '#update_slug!' do
+    it 'updates the slug of the section' do
+      allow(SlugGenerator).to receive(:new).and_call_original
+
+      manual = Manual.new(title: 'manual-title')
+      section = manual.build_section(title: 'section-title')
+      manual.save(User.gds_editor)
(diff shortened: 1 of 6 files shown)


spec/models/section_spec.rb
```diff
@@ -110,6 +110,21 @@ def key_classes_for(hash)
     allow(SlugGenerator).to receive(:new).with(prefix: manual_slug).and_return(slug_generator)
   end
 
+  describe '#update_slug!' do
+    it 'updates the slug of the section' do
+      allow(SlugGenerator).to receive(:new).and_call_original
```
floehopper: I don't understand why you need this. If you really do need it, can you explain why in the commit note?

chrisroos: Good spot. I've updated the commit note to explain why I'm using it:

> Note the use of `and_call_original` to restore the original behaviour of
> `SlugGenerator.new`. Without this, the stubbing of `SlugGenerator.new`
> in the main `before` block was complaining because it was expecting to
> be initialized with the slug that's set on the `manual` double for the
> rest of the tests.

chrisroos:
Thanks for reviewing, @floehopper. I've rebased on master and force pushed in preparation for merging.

PR #1156, opened 2017-05-25
Construct Sections with previous and latest editions

chrisroos:
Outside of the tests we'd only ever construct a `Section` with a maximum of the 2 latest editions (in `Section.find`). Explicitly passing the previous and latest editions doesn't change the behaviour but does aim to make that behaviour more explicit.

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -36,12 +36,17 @@ class Section
   def initialize(manual:, uuid:, previous_edition: nil, latest_edition: nil)
     @slug_generator = SlugGenerator.new(prefix: manual.slug)
     @uuid = uuid
-    @editions = [previous_edition, latest_edition].compact.uniq
-    if @editions.empty?
-      edition = SectionEdition.new(state: "draft", version_number: 1, section_uuid: uuid)
-      @editions.push(edition)
+
+    @previous_edition = previous_edition
+    @latest_edition = latest_edition
+
+    if @previous_edition == @latest_edition
+      @previous_edition = nil
     end
-    @latest_edition = @editions.last
+
+    if @latest_edition.nil?
+      @latest_edition = SectionEdition.new(state: "draft", version_number: 1, section_uuid: uuid)
+    end
   end
 
   def update_slug!(full_new_section_slug)
@@ -53,7 +58,8 @@ class Section
     # think it's safer to save latest two as both are exposed to the and have
     # potential to change. This extra write may save a potential future
     # headache.
-    editions.last(2).each(&:save!)
+    previous_edition && previous_edition.save!
+    latest_edition.save!
   end
 
   def minor_update?
@@ -86,20 +92,20 @@ class Section
           slug: slug,
           attachments: attachments,
         )
+      @previous_edition = latest_edition
       @latest_edition = SectionEdition.new(attributes)
-
-      editions.push(@latest_edition)
     end
 
     nil
   end
 
   def published?
-    editions.any?(&:published?)
+    latest_edition.published? ||
+      (previous_edition && previous_edition.published?)
   end
 
   def has_ever_been_published?
-    return false if @editions.size == 1 && needs_exporting?
+    return false if previous_edition.nil? && needs_exporting?
     published?
   end
 
@@ -152,7 +158,8 @@ class Section
   end
 
   def persisted?
-    editions.any?(&:persisted?)
+    latest_edition.persisted? ||
+      (previous_edition && previous_edition.persisted?)
   end
 
   def eql?(other)
@@ -181,10 +188,14 @@ class Section
 
 private
 
-  attr_reader :slug_generator, :editions, :latest_edition
+  attr_reader :slug_generator, :latest_edition, :previous_edition
 
   def published_edition
-    most_recent_non_draft = editions.reject(&:draft?).last
+    most_recent_non_draft = if !latest_edition.draft?
+                              latest_edition
+                            elsif previous_edition && !previous_edition.draft?
+                              previous_edition
+                            end
 
     if most_recent_non_draft && most_recent_non_draft.published?
       most_recent_non_draft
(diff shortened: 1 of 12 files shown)


spec/models/section_spec.rb
```diff
@@ -162,7 +163,7 @@ def key_classes_for(hash)
       end
 
       it 'builds a section using the editions' do
```
floehopper: Minor: Maybe reword the example name...?

app/models/section.rb
```diff
@@ -25,18 +25,18 @@ def self.find(manual, section_uuid, published: false)
...
   def_delegators :latest_edition, :title, :slug, :summary, :body, :updated_at, :version_number, :change_note, :minor_update, :exported_at
 
   attr_reader :uuid
 
-  def initialize(manual:, uuid:, editions:)
+  def initialize(manual:, uuid:, previous_edition: nil, latest_edition: nil)
     @slug_generator = SlugGenerator.new(prefix: manual.slug)
     @uuid = uuid
-    @editions = editions
+    @editions = [previous_edition, latest_edition].compact.uniq
```
floehopper: Can you explain why you needed to add the call to `Array#uniq`? Also are there any examples in the specs which check this behaviour?

app/models/section.rb
```diff
@@ -181,10 +188,14 @@ def all_editions
...
   def published_edition
-    most_recent_non_draft = editions.reject(&:draft?).last
+    most_recent_non_draft = if !latest_edition.draft?
+                              latest_edition
+                            elsif previous_edition && !previous_edition.draft?
+                              previous_edition
+                            end
 
     if most_recent_non_draft && most_recent_non_draft.published?
       most_recent_non_draft
```
floehopper: Observation: I haven't looked at it in detail, but I suspect there are some similarities between the logic in this method and [`Manual#current_published_version`](https://github.com/alphagov/manuals-publisher/blob/1947ca0d0cc264bde50bc542f95dd3a07de2cee5/app/models/manual.rb#L310-L328).

app/models/section.rb
```diff
@@ -181,10 +188,14 @@ def all_editions
...
-  attr_reader :slug_generator, :editions, :latest_edition
+  attr_reader :slug_generator, :latest_edition, :previous_edition
 
   def published_edition
-    most_recent_non_draft = editions.reject(&:draft?).last
+    most_recent_non_draft = if !latest_edition.draft?
+                              latest_edition
+                            elsif previous_edition && !previous_edition.draft?
+                              previous_edition
+                            end
```
floehopper: Minor: You mention in the commit note that you had to make some changes due to a Rubocop violation, but you had to change this method anyway, didn't you? Unless I've misunderstood, I might be tempted to remove that sentence from the commit note.

app/models/section.rb
```diff
@@ -152,7 +158,8 @@ def mark_as_exported!(exported_at = Time.zone.now)
   end
 
   def persisted?
-    editions.any?(&:persisted?)
+    previous_edition && previous_edition.persisted? ||
+      latest_edition.persisted?
```
floehopper: Minor: I might swap this round, i.e. check `latest_edition.persisted?` first.

floehopper: Minor: I would wrap the first two conditions in parentheses rather than relying on operator precedence.

app/models/section.rb
```diff
@@ -36,12 +36,17 @@ def self.find(manual, section_uuid, published: false)
...
-    if @editions.empty?
-      edition = SectionEdition.new(state: "draft", version_number: 1, section_uuid: uuid)
-      @editions.push(edition)
+
+    @previous_edition = previous_edition
+    @latest_edition = latest_edition
+
+    if @previous_edition == @latest_edition
+      @previous_edition = nil
+    end
```
floehopper: I think this is related to my previous question about the use of `Array#uniq`, but why is this `if` block necessary?

app/models/section.rb
```diff
@@ -86,20 +92,20 @@ def update(attributes)
...
-      editions.push(@latest_edition)
     end
 
     nil
   end
 
   def published?
-    editions.any?(&:published?)
+    previous_edition && previous_edition.published? ||
+      latest_edition.published?
```
floehopper: Minor: My two comments on `Section#persisted?` below apply to this method too.

spec/models/section_spec.rb
chrisroos: I've reworded this example and added another to demonstrate the `Section` being built with both a previous and latest edition.

app/models/section.rb
chrisroos: I've updated the commit note to explain the call to `Array#uniq?`:

> Note the call to `compact` for the array of editions in
> `Section#initialize`. This avoids nil elements in the array when
> `previous_edition` and/or `latest_edition` is nil. The call to `uniq`
> avoids a theoretical problem of the same edition being passed as both
> `previous_edition` and `latest_edition`. I've not worried about testing
> this behaviour as I'm expecting this to change again the near future.

app/models/section.rb
chrisroos: I've updated the commit note to explain this code:

> Note that I'm continuing to avoid storing a `previous_edition` and
> `latest_edition` in `Section#initialize` when they're both the same
> object. I haven't tested this behaviour as I'm hoping/expecting it to go
> away in the near future.

app/models/section.rb
chrisroos: I've updated this so that I check `latest_edition.published?` first and I've wrapped the `previous_edition` conditions in parentheses.

app/models/section.rb
chrisroos: I've updated this so that I check `latest_edition.persisted?` first and I've wrapped the `previous_edition` conditions in parentheses.

app/models/section.rb
chrisroos: Good spot. I had it in mind that I was changing existing code when I addressed the Rubocop feedback but I was actually only changing the code that I'd just added. I've removed my comment about Rubocop from the commit note.

PR #1157, opened 2017-05-29
Refactor attachment services

chrisroos:
I think inlining these methods makes it easier to see the behaviour of the various `#call` methods.

### DIFF
--- a/app/services/attachment/new_service.rb
+++ b/app/services/attachment/new_service.rb
@@ -6,18 +6,15 @@ class Attachment::NewService
   end
 
   def call
+    section = manual.find_section(section_uuid)
     attachment = Attachment.new({})
-    
+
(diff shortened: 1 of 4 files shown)


app/services/attachment/new_service.rb
```diff
@@ -14,7 +14,7 @@ def call
   attr_reader :user, :manual_id, :section_uuid
 
   def attachment
-    Attachment.new(initial_params)
+    Attachment.new({})
```
floehopper: Minor: I think this could be further simplified to `Attachment.new`, because `Attachment` is a `Mongoid::Document` and so an empty Hash of attributes is the default.

chrisroos:
Thanks @floehopper and @chrislo!

PR #1158, opened 2017-05-29
Refactor manual services

chrisroos:
I think inlining these methods makes it easier to see the behaviour of the various `#call` methods.

### DIFF
--- a/app/services/manual/preview_service.rb
+++ b/app/services/manual/preview_service.rb
@@ -16,10 +16,6 @@ private
   attr_reader :user, :manual_id, :attributes
 
   def manual
-    manual_id ? existing_manual : Manual.new(attributes)
-  end
-
-  def existing_manual
-    @existing_manual ||= Manual.find(manual_id, user)
+    manual_id ? Manual.find(manual_id, user) : Manual.new(attributes)
   end
 end
--- a/app/services/manual/preview_service.rb
+++ b/app/services/manual/preview_service.rb
@@ -16,11 +16,7 @@ private
   attr_reader :user, :manual_id, :attributes
 
   def manual
-    manual_id ? existing_manual : ephemeral_manual
-  end
-
-  def ephemeral_manual
-    Manual.new(attributes)
+    manual_id ? existing_manual : Manual.new(attributes)
   end
 
   def existing_manual
--- a/app/services/manual/preview_service.rb
+++ b/app/services/manual/preview_service.rb
@@ -6,6 +6,11 @@ class Manual::PreviewService
   end
 
   def call
+    manual = if manual_id
+               Manual.find(manual_id, user)
+             else
+               Manual.new(attributes)
+             end
     manual.update(attributes)
 
     ManualPresenter.new(manual)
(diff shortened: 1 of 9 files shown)


app/services/manual/publish_service.rb
```diff
@@ -10,7 +10,7 @@ def initialize(user:, manual_id:, version_number:)
   def call
     if version_number == manual.version_number
       manual.publish
-      log_publication
+      PublicationLogger.new.call(manual)
```
floehopper: Minor: I might be tempted to take this a step further and build the `PublicationLogger` in the service constructor.

app/services/manual/preview_service.rb
```diff
@@ -6,6 +6,11 @@ def initialize(user:, manual_id:, attributes:)
   end
 
   def call
+    manual = if manual_id
+               Manual.find(manual_id, user)
+             else
+               Manual.new(attributes)
+             end
```
floehopper: Minor: I find these hanging indents hard to read. Is there a reason this can't remain a ternary expression?

app/services/manual/queue_publish_service.rb
```diff
@@ -12,6 +12,12 @@ def call
         manual_id: manual.id,
         version_number: manual.version_number,
       )
+
+      govuk_header_params = {
+        request_id: GdsApi::GovukHeaders.headers[:govuk_request_id],
+        authenticated_user: GdsApi::GovukHeaders.headers[:x_govuk_authenticated_user],
+      }
+
```
floehopper: Out-of-scope: I think it would be better to extract these header values within the controller action and pass them into the service as constructor args.

app/services/manual/withdraw_service.rb
```diff
@@ -40,8 +40,6 @@ def remove_from_search_index
 
   def manual
     @manual ||= Manual.find(manual_id, user)
-  rescue KeyError => error
-    raise ManualNotFoundError.new(error)
```
floehopper: I think this should be dealt with separately, because `ManualWithdrawer#execute` is rescuing `Manual::WithdrawService::ManualNotFoundError`. I think the latter should probably be rescuing `Manual::NotFoundError` instead and that change would probably be better done at the same time as this.

chrisroos:
Thanks @floehopper. I'm going to address your feedback and get this merged.

app/services/manual/preview_service.rb
chrisroos: I don't think there's any reason it can't be a ternary expression. I'm going to leave it as it is but please feel free to change it as you see fit.

app/services/manual/publish_service.rb
chrisroos: I don't feel strongly about this so I'm going to leave it as it is.

app/services/manual/withdraw_service.rb
chrisroos: I've removed the commit that removed this `rescue` statement but I'm not planning to do anything more about it. I'm happy to leave it until we revisit this code.

chrisroos:

chrisroos:

chrisroos:
> As per my comment, I'd prefer if the "Remove rescue statement from Manual#WithdrawServiece#manual" commit was removed from this PR and a new one create which also deals with the rescuing of `Manual::WithdrawService::ManualNotFoundError` in `ManualWithdrawer#execute`.

I've removed the commit. I'm not planning to create a new PR with the changes suggested. I'm happy that we can make them if/when we revisit this code.

PR #1159, opened 2017-05-29
Improve Section#version_type

chrisroos:
The `Section#version_type` method is used to determine the `update_type` that's sent to the Publishing API. This should always be `major` for the first edition of a section and determined by `Section#minor_update?` otherwise. I think this change makes that intention clearer.

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -172,15 +172,19 @@ class Section
     published? && !minor_update?
   end
 
+  def first_edition?
+    version_number == 1
+  end
+
   def version_type
-    if has_ever_been_published?
+    if first_edition?
+      :new
+    else
       if minor_update?
         :minor
       else
         :major
       end
-    else
-      :new
     end
   end
 
--- a/spec/models/section_spec.rb
+++ b/spec/models/section_spec.rb
@@ -850,10 +850,33 @@ describe Section do
     end
   end
 
+  describe '#first_edition?' do
+    let(:manual) { Manual.new(title: 'manual-title') }
+    let(:section) { manual.build_section(title: 'section-title') }
+
+    before do
+      allow(SlugGenerator).to receive(:new).and_call_original
+    end
+
+    it 'returns true when the version_number is 1' do
+      expect(section.version_number).to eq(1)
+      expect(section).to be_first_edition
+    end
+
+    it 'returns false when the version_number is greater than 1' do
+      section.publish!
+      section.save
+      section.update(title: 'new-section-title')
+
+      expect(section.version_number).to eq(2)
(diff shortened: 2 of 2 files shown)


app/adapters/publishing_adapter.rb
```diff
@@ -200,7 +200,7 @@ def save_section_content(section, manual, republish: false)
     organisation = organisation_for(manual)
 
     update_type = case version_type(republish) || section.version_type
-                  when :new, :major
+                  when :major
                     GdsApiConstants::PublishingApiV2::MAJOR_UPDATE_TYPE
                   when :minor
                     GdsApiConstants::PublishingApiV2::MINOR_UPDATE_TYPE
```
floehopper: While in principle I think this is a sensible change, at the moment there is [some duplication][1] between `PublishingAdapter#save_section_content` and `#save_manual_content`. I had it on my to-do list to remove this duplication, but hadn't got round to DRY-ing it up. I'm slightly wary of making this change which causes the duplicate code to diverge. I'd rather we made the same change to both (if that's possible).

[1]: https://github.com/alphagov/manuals-publisher/blob/8d9878843a8bb6b80405007588d652d8f6b5c05b/app/adapters/publishing_adapter.rb#L98-L107

chrisroos:
> This PR is in effect moving the logic that says a new edition is always regarded as a major update as far as the Publishing API is concerned from `PublishingAdapter` into `Section`. I guess the question is whether that logic should be part of the domain or whether it's something very specific to the Publishing API. I'm not sure what the answer is to that, but I'm leaning towards the latter.

I think the problem is that we need to have some knowledge of this in order to know whether the user can mark an edition of a section as minor or major - the first edition is _always_ major and all subsequent editions can be toggled. I have another branch that builds on this and uses `Section#first_edition` to decide whether the minor/major toggle can be shown in the interface.

My overriding motivation is to get rid of the `Section#has_ever_been_published?` method so what do you think about me just merging the first commit of this branch ("Improve Section#version_type") and leaving the other changes for another time?

chrisroos:
> Perhaps a middle-ground would be to remove `Section#version_type` and just call `Section#first_edition?` and `Section#minor_update?` from `PublishingAdapter`.

I think this is probably a good idea although I'm not sure I'll get to it soon.

floehopper:
@chrisroos:

> what do you think about me just merging the first commit of this branch ("Improve Section#version_type") and leaving the other changes for another time?

I'm happy with that approach 👍

chrisroos:
Thanks @floehopper. I've retained the first commit and removed the others. I've rebased and force-pushed in preparation for merging.

PR #1160, opened 2017-05-29
Simplify Section#publish!

chrisroos:
Inlining `Section#published_edition` made it clear that the `Section#publish!` method was more complicated than it needed to be.


app/models/section.rb
```diff
@@ -123,19 +123,10 @@ def attachments
...
-
-      if most_recent_non_draft && most_recent_non_draft.published?
-        published_edition = most_recent_non_draft
+      if previous_edition && previous_edition.published?
+        previous_edition.archive
       end
 
-      published_edition.archive if published_edition
-
       latest_edition.publish
```
floehopper: This all makes sense to me, but it's pretty hard to be sure that the new implementation is correct. I think it might help to refer to [the examples for this method][1] from the commit note and either confirm that they provide sufficient coverage or add more coverage.

[1]: https://github.com/alphagov/manuals-publisher/blob/4d6f8b02f034b2e39746b22dae658c408b640220/spec/models/section_spec.rb#L528-L562

chrisroos:
Thanks @floehopper. I'm _fairly_ confident I haven't changed the behaviour of this method so I'm happy to merge it without checking any further.

PR #1169, opened 2017-05-31
Remove Section#has_ever_been_published?

chrisroos:
This follows on from PR #1159 and replaces the final use of `Section#has_ever_been_published?` before then removing the method.

### DIFF
--- a/spec/view_adapters/section_view_adapter_spec.rb
+++ b/spec/view_adapters/section_view_adapter_spec.rb
@@ -1,0 +1,48 @@
+require "spec_helper"
+
+RSpec.describe SectionViewAdapter do
+  let(:manual) { Manual.new(title: 'manual-title') }
+  let(:section) { manual.build_section(title: 'section-title') }
+
+  subject do
+    described_class.new(manual, section)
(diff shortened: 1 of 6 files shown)


spec/view_adapters/section_view_adapter_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it 'returns true if section is not first edition and is published' do
+      section.publish!
+      section.save
+      section.update(title: 'new-section-title')
+
+      expect(subject).to be_draft
+      expect(subject).to_not be_first_edition
+      expect(subject.accepts_minor_updates?).to be(true)
+    end
+  end
```
floehopper: Minor: Is it worth including the final logical combination, i.e. not first edition and published?

chrisroos: I've updated the first commit to include this test.

chrisroos:
Thanks for the review, @floehopper. I've modified the first commit based on your feedback, rebased on master and force pushed in preparation for merging.

PR #1170, opened 2017-06-02
Avoid unnecessary variable assignment in Section methods

chrisroos:


chrisroos:

PR #1178, opened 2017-06-23
Remove govuk_content_models reference

alecgibson:
This app no longer depends on `govuk_content_models`, so remove
a comment referring to it.

### Trello

https://trello.com/c/9Ho9O9Yo/219-epic-retire-govukcontentmodels


floehopper:
I've added [a new issue](https://github.com/alphagov/manuals-publisher/issues/1179) to suggest removing the version constraint on the govspeak gem now that it is no longer needed.

PR #1190, opened 2017-11-14
[Do not merge] Only unpublish sections which have been previously published

steventux:
https://trello.com/c/OmstjgzY/30-investigate-manuals-publisher-bug-which-prevented-publish

There's a problem when a draft section which hasn't been published and
is then removed, the publishing API will discard the draft section but
the section is also added to `Manual#removed_sections` which is processed
on publishing the manual, attempting to unpublish an non-existent section
from the publishing API and resulting in a `GdsApi::HTTPNotFound` error.

So only attempt to unpublish previously published sections, drafts are
already discarded at the removal workflow step.

### DIFF
--- a/spec/adapters/publishing_adapter_spec.rb
+++ b/spec/adapters/publishing_adapter_spec.rb
@@ -561,6 +561,7 @@ describe PublishingAdapter do
       manual.removed_sections = [removed_section]
 
       allow(removed_section).to receive(:withdrawn?).and_return(false)
+      allow(removed_section).to receive(:published?).and_return(true)
 
       allow(publishing_api).to receive(:publish).with(anything, anything)
       allow(publishing_api).to receive(:unpublish).with(anything, anything)
--- a/app/adapters/publishing_adapter.rb
+++ b/app/adapters/publishing_adapter.rb
@@ -279,7 +279,12 @@ private
 
   def unpublish_section(section, manual, republish:)
     if !section.withdrawn? || republish
-      Services.publishing_api.unpublish(section.uuid, type: "redirect", alternative_path: "/#{manual.slug}", discard_drafts: true)
+      if section.published?
+        Services.publishing_api.unpublish(section.uuid,
+                                          type: "redirect",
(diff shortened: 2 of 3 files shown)


spec/adapters/publishing_adapter_spec.rb
```diff
@@ -560,7 +560,7 @@
       manual.sections = [section]
       manual.removed_sections = [removed_section]
 
-      allow(removed_section).to receive(:withdrawn?).and_return(false)
+      allow(removed_section).to receive(:published?).and_return(true)
```
floehopper: It looks as if the `if !section.withdrawn?` condition still exists in the implementation. How come you've *replaced* the stubbing of `Section#withdrawn?` with the stubbing of `Section#published?` rather than stubbing combinations of both (introducing new examples where necessary)...?

steventux:
@floehopper thanks for the suggestion, having a contextual removal button may well be simpler. I'm concerned this check for a previously published item might have some edge cases which we won't find until some magic workflow is performed.

steventux:
Having considered contextual buttons to remove draft sections and withdraw previously published sections, I wonder about the state of the editions in the Publishing API if we were to move the unpublishing of a section to an earlier workflow step (ie clicking a "withdraw" button).
I think we'd need to communicate a change in workflow so that it was clear than removing a previously published section was an action which immediately affected the live site. The current workflow is that a section gets unpublished on publishing the manual.

floehopper:
@steventux: If the users would find it acceptable, I certainly think it would make the publishing app simpler. It would also make it easier to remove the local database in the future, i.e. to make the app "fully migrated" from a Publishing Platform point of view.

h-lame:
I reckon it'd be totally weird to have discarding an published section be something that happens directly to the live site whereas everything else happens only when you press publish on the manual.

floehopper:
@h-lame: Good point 👍

steventux:
A problem spotted on integration needs a fix before merging this.

PR #1195, opened 2017-11-27
Add Dockerfile and seeds for using within end-to-end tests

RyanMacG:
This allows for the app to be included in the end-to-end testing project.


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ruby:2.2.4
+RUN apt-get update -qq && apt-get upgrade -y
+RUN apt-get install -y build-essential nodejs && apt-get clean
+
```
kevindew: You seem to have a rogue extra line here

PR #1219, opened 2017-12-12
Add link check report to db

tsteer:
Add link check report and link check report links to db. This stores status of links when link checker is run

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -1,0 +1,17 @@
+class Link
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  embedded_in :link_check_report
+
+  field :uri, type: String
+  field :status, type: String
+  field :checked, type: DateTime
+  field :check_warnings, type: Array
+  field :check_errors, type: Array
+  field :problem_summary, type: String
+  field :suggested_fix, type: String
+
+  validates :uri, presence: true
+  validates :status, presence: true
+end
--- a/app/models/link_check_report.rb
+++ b/app/models/link_check_report.rb
@@ -1,0 +1,19 @@
+class LinkCheckReport
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  embeds_many :links
+
+  accepts_nested_attributes_for :links
+
+  field :batch_id, type: Integer
+  field :status, type: String
+  field :manual_id, type: String
+  field :section_id, type: String
+  field :completed_at, type: DateTime
+
+  validates :batch_id, presence: true
+  validates :status, presence: true
+  validates :links, presence: true
+  validates :manual_id, presence: true
+end
--- a/spec/models/link_check_report_spec.rb
+++ b/spec/models/link_check_report_spec.rb
@@ -1,0 +1,50 @@
+require "spec_helper"
+
+describe LinkCheckReport, type: :model do
+  let(:attributes) do
+    {
+      links: [{ uri: "http://www.example.com", status: "error" }],
+      batch_id: 1,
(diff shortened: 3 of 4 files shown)


app/models/link_check_report.rb
```diff
@@ -0,0 +1,18 @@
...
+  include Mongoid::Timestamps
+
+  embeds_many :links
+
+  accepts_nested_attributes_for :links
+
+  field :batch_id, type: Integer
+  field :status, type: String
+  field :link_reportable_type, type: String
+  field :link_reportable_id, type: Integer
```
lukemorton: I wonder if we need the same polymorphic relationship with link check reports in manuals publisher.

Are we going to be checking links for more than manuals? Perhaps @thomasleese would know.

tsteer: @lukemorton Seb thinks that there are a few different types, so think link_reportable_type is needed

app/models/link.rb
```diff
@@ -0,0 +1,19 @@
+class Link
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  embedded_in :link_check_report
+
+  field :uri, type: String
```
thomasleese: Do we need to add some `null: false` on these?

lukemorton: Will that then mean the key does not exist in mongo? If that's preferred then okay we can add to any fields we expect to be null.

lukemorton: I think though that :uri will never be null as we have validations below to ensure that.

thomasleese: Hmm, no, I thought it would add some validation to make sure the fields aren't ever `nil`, but it looks like you can't actually do that in Mongoid anyway.

app/models/link.rb
```diff
@@ -0,0 +1,19 @@
...
+  field :status, type: String
+  field :checked, type: DateTime
+  field :check_warnings, type: String
```
thomasleese: Should these not be arrays of strings?

lukemorton:
Yes good catch. I think with Mongoid you can only specify the type as
`Array` rather than an array of strings.

Thinking about it, feels like this should be an embeds_many and we
introduce another model for errors and warnings.

app/models/link.rb
thomasleese: I would just stick with `type: Array` to be honest, no need to specify the element type exactly.

tsteer:
@thomasleese Could you have a look at this please?

.rubocop.yml
```diff
@@ -1,6 +1,7 @@
 AllCops:
   TargetRubyVersion: 2.2
-
+Metrics/BlockLength:
```
thomasleese: I don't quite understand why we're making this change?

h-lame: We had a discussion about this in #1224 but ultimately decided we don't need it.  This commit needs to be dropped form this PR (and any others it's been introduced in).

lukemorton: Dropped 👍

app/models/link_check_report.rb
thomasleese: Here https://github.com/alphagov/manuals-publisher/pull/1225/files#diff-4c62e07f8da55a84ff899aa3c1293d9eR26 we see to be referring to some a `manuals_id` and `section_id` field, do we need to change that here too?

thomasleese: As mentioned on that PR, if we're going to only have `manual_id` and `section_id`, we could define a method:

```
def is_for_section?
  section_id.present?
end
```

PR #1222, opened 2017-12-13
Implement link checking endpoint for manuals

tsteer:
- Add link checker api to services

### DIFF
--- a/app/services/link_check_report/create_service.rb
+++ b/app/services/link_check_report/create_service.rb
@@ -1,0 +1,63 @@
+require "services"
+require "govspeak/link_extractor"
+
+class LinkCheckReport::CreateService
+  class InvalidReport < RuntimeError
+    def initialize(original_error)
+      @message = original_error.message
+    end
+  end
+
+  def initialize(user:, link_reportable_type:, link_reportable_id:)
+    @user = user
+    @link_reportable_type = link_reportable_type
+    @link_reportable_id = link_reportable_id
+  end
+
+  def call
+    link_report = call_link_checker_api
+
+    report = LinkCheckReport.new(
+      batch_id: link_report.fetch(:batch_id),
+      completed_at: link_report.fetch(:completed_at),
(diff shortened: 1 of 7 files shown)


app/services/link_check_report/create_service.rb
```diff
@@ -0,0 +1,58 @@
...
+      Manual.find(link_reportable_id, user)
+    end
+  end
+
+  def uris
+    Govspeak::LinkExtractor.new(link_reportable.body).links
+  end
+
+  def call_link_checker_api
+    Services.link_checker_api.create_batch(uris)
```
SebAshton: Needs the `webhook_uri` and a `webmock_secret` too

lukemorton: We'll cover this in a subsequent PR once there is a webhook_uri to point to.

app/services/link_check_report/create_service.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def call
+    link_report = call_link_checker_api
+
+    report = LinkCheckReport.new(
+      batch_id: link_report.fetch(:batch_id),
+      completed_at: link_report.fetch(:completed_at),
+      status: link_report.fetch(:status),
+      link_reportable_type: link_reportable_type,
```
thomasleese: Should this be using `manual_id` and `section_id`?

spec/services/link_check_report/create_service_spec.rb
```diff
@@ -0,0 +1,124 @@
+require "spec_helper"
+
+RSpec.describe LinkCheckReport::CreateService do
+  let(:user) { FactoryGirl.create(:user) }
+
+  let(:link_check_report) { double(:link_check_report, valid?: nil, save!: nil) }
```
thomasleese: Would be good to use a Factory Bot class here to make sure the fields definitely exist on the model.

PR #1224, opened 2017-12-14
Ignore BlockLength as jenkins doesn’t care

tsteer:
We have removed the lint check for block length from rubocop as the configuration on jenkins indicates that when we explicitly ignore it for a single block that in fact it's not necessary at all:

```
spec/models/link_check_report_spec.rb:3:1: W: Lint/UnneededDisable: Unnecessary disabling of Metrics/BlockLength.
# rubocop:disable Metrics/BlockLength
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/models/link_spec.rb:3:1: W: Lint/UnneededDisable: Unnecessary disabling of Metrics/BlockLength.
# rubocop:disable Metrics/BlockLength
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,7 @@
 AllCops:
   TargetRubyVersion: 2.2
-
+Metrics/BlockLength:
+  Enabled: false
 Lint/HandleExceptions:
   Exclude:
     - 'spec/**/*'


floehopper:
I may well be missing something, but isn't this cop [already disabled in specs by `govuk-lint`][1]?

[1]: https://github.com/alphagov/govuk-lint/blob/4caf53ffa6be259097ac7ecda03facca9efc2118/configs/rubocop/other-excludes.yml#L5-L9

h-lame:
I think you're right @floehopper  - when I wrote my comment @tsteer hadn't written the extra info above so I didn't clock that it was about some specs.  I don't think we need to make this change at all, we already ignore block length for specs so as long as we don't include `# rubocop:disable Metrics/BlockLength` magic comments in the specs then lint won't warn that it's an unneeded disabling.

Does that make sense to you @tsteer?  Would it be ok to close this without merging?

tsteer:
@h-lame Hmmmm, I was getting BlockLength metric errors previously. Removing this line now and rubocop is working fine again.

Nothing to see here... move along.

PR #1225, opened 2017-12-14
Add link check reports for sections

tsteer:


h-lame:
We should probably remove the `Ignore BlockLength as jenkins doesn’t care` commit from here based on the discussion on #1224

app/services/link_check_report/create_service.rb
```diff
@@ -33,11 +35,14 @@ def call
 
 private
 
-  attr_reader :user, :link_reportable_type, :link_reportable_id
+  attr_reader :user, :link_reportable_type, :manual_id, :section_id
 
   def link_reportable
     if link_reportable_type == "manual"
```
thomasleese: Do we need this, could we just have?

```
def is_for_section?
  section_id.present?
end
```

thomasleese:
Seems like this could be merged with #1222?

PR #1226, opened 2017-12-14
Add callback for Link Checker API

SebAshton:
This PR adds:

- [x] LinkCheckReports::UpdateService service class to update link check reports
- [x] Controller action to be used by the batch endpoint callback that will update a link check report

This closed the loop for link check report creation, by updating created link reports with the results from the API

### DIFF
--- a/app/services/link_check_report/create_service.rb
+++ b/app/services/link_check_report/create_service.rb
@@ -2,6 +2,10 @@ require "services"
 require "govspeak/link_extractor"
 
 class LinkCheckReport::CreateService
+  include Rails.application.routes.url_helpers
+
+  CALLBACK_HOST = Plek.find('manuals-publisher')
+
   class InvalidReport < RuntimeError
     def initialize(original_error)
       @message = original_error.message
@@ -63,7 +67,13 @@ private
   end
 
   def call_link_checker_api
-    Services.link_checker_api.create_batch(uris)
+    callback = link_checker_api_callback_url(host: CALLBACK_HOST)
+
+    Services.link_checker_api.create_batch(
+      uris,
+      webhook_uri: callback,
+      webhook_secret_token: Rails.application.secrets.link_checker_api_secret_token
+    )
   end
 
   def map_link_attrs(link)
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -12,11 +12,14 @@
 
 development:
   secret_key_base: 575d1ff900b8efcf99c89980371365dc659ac3b3b8d6ee034ff27147a312a39c994eb2c189f06cf875bbcc645c7916080b95173b8dfe2b7063a34637a4a46541
+  link_checker_api_secret_token: KWceiCFx6GjzvKQh28EiRiUE
 
 test:
   secret_key_base: 46dbb3ca836e2c15ed6e4cb4b1d30c9b3e99ec83d472be4973c43fa5d6eaea84e49baeb5167bb77c82dde54edadda7d1abfa0b73c4d2d1f547542e804156612d
+  link_checker_api_secret_token: KWceiCFx6GjzvKQh28EiRiUE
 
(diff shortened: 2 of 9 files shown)


app/services/link_check_report/create_service.rb
```diff
@@ -51,7 +51,11 @@ def uris
   end
 
   def call_link_checker_api
-    Services.link_checker_api.create_batch(uris)
+    Services.link_checker_api.create_batch(
+      uris,
+      webhook_uri: "#{Plek.find('manuals-publisher')}/link-checker-api-callback",
```
thomasleese: Could we use the Rails route methods here to avoid hard coding the string?

config/secrets.yml
```diff
@@ -12,11 +12,14 @@
...
 
 test:
   secret_key_base: 46dbb3ca836e2c15ed6e4cb4b1d30c9b3e99ec83d472be4973c43fa5d6eaea84e49baeb5167bb77c82dde54edadda7d1abfa0b73c4d2d1f547542e804156612d
+  link_checker_api_secret_token: KWceiCFx6GjzvKQh28EiRiUE
 
 # Do not keep production secrets in the repository,
 # instead read values from the environment.
 production:
   secret_key_base: <%= ENV["SECRET_TOKEN"] %>
+  link_checker_api_secret_token: <%= ENV["LINK_CHECKER_API_SECRET_TOKEN"] %>
```
thomasleese: Might be good to use `ENV.fetch` here so there would be a clear error if it wasn't set.

SebAshton: Adding a `fetch` blows up the test suite 😞

thomasleese: That's weird... :/ Shouldn't it be using the `test` values :S

SebAshton: When the `yml` is parsed, I expect it tries to populate all the key/values first then takes then takes the specific env.

thomasleese: Oh ugh, I see. I suppose that makes sense.

spec/services/link_check_report/update_service_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    }]
+  end
+
+  subject do
+    described_class.new(report: link_check_report, payload: payload)
+  end
+
+  before do
+    allow(link_check_report).to receive(:update!)
+    allow(Link).to receive(:find).with(link_check_report: link_check_report, uri: "http://www.example.com").and_return(link_1)
```
thomasleese: Is it better to use Factory Bot here to have these available?

SebAshton: Updated to use factory @thomasleese :)

app/services/link_check_report/create_service.rb
```diff
@@ -53,7 +55,13 @@ def uris
   end
 
   def call_link_checker_api
-    Services.link_checker_api.create_batch(uris)
+    callback = link_checker_api_callback_url(host: Plek.find('manuals-publisher'))
```
thomasleese: I would possible extract this out into a memoised method.

spec/services/link_check_report/update_service_spec.rb
thomasleese: Could we use `create` here so we get real objects and avoid having to stub out the database query methods?

SebAshton: Happy to, but I tried to keep these tests symmetrical to the other [update tests](https://github.com/alphagov/manuals-publisher/blob/master/spec/services/section/update_service_spec.rb) within the repo.

thomasleese: The history of this app means it's written in a way that's not very conventional Rails, I wouldn't worry about matching the other tests.

SebAshton:
👍  confirmed working on integration.

PR #1228, opened 2017-12-15
Implement link checking endpoint for manuals and sections

tsteer:
- Add link checker api to Services
- Add link check reports for sections

### DIFF
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -1,0 +1,19 @@
+class LinkCheckReportsController < ApplicationController
+  def create
+    service = LinkCheckReport::CreateService.new(
+      user: current_user,
+      link_reportable_type: link_reportable_params[:type],
+      link_reportable_id: link_reportable_params[:id]
+    )
+
+    service.call
+
+    head :created
+  end
+
+private
+
+  def link_reportable_params
+    params.require(:link_reportable).permit(:type, :id)
+  end
+end
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -3,7 +3,8 @@ class LinkCheckReportsController < ApplicationController
     service = LinkCheckReport::CreateService.new(
       user: current_user,
       link_reportable_type: link_reportable_params[:type],
-      link_reportable_id: link_reportable_params[:id]
+      manual_id: link_reportable_params[:manual_id],
+      section_id: link_reportable_params[:section_id]
     )
 
     service.call
@@ -14,6 +15,6 @@ class LinkCheckReportsController < ApplicationController
 private
 
   def link_reportable_params
-    params.require(:link_reportable).permit(:type, :id)
+    params.require(:link_reportable).permit(:type, :manual_id, :section_id)
   end
 end
--- a/app/services/link_check_report/create_service.rb
+++ b/app/services/link_check_report/create_service.rb
@@ -1,0 +1,63 @@
+require "services"
+require "govspeak/link_extractor"
+
+class LinkCheckReport::CreateService
+  class InvalidReport < RuntimeError
+    def initialize(original_error)
+      @message = original_error.message
+    end
+  end
+
+  def initialize(user:, link_reportable_type:, link_reportable_id:)
+    @user = user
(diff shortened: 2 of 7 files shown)


app/controllers/link_check_reports_controller.rb
```diff
@@ -0,0 +1,20 @@
+class LinkCheckReportsController < ApplicationController
+  def create
+    service = LinkCheckReport::CreateService.new(
+      user: current_user,
+      link_reportable_type: link_reportable_params[:type],
```
thomasleese: Do we still need this?

app/services/link_check_report/create_service.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  attr_reader :user, :manual_id, :section_id
+
+  def is_for_section?
+    section_id.present?
+  end
+
+  def link_reportable
+    if is_for_section?
+      manual = Manual.find(manual_id, user)
```
thomasleese: This line could be moved out of the `if` as we do it in both branches.

thomasleese: In fact, would be tempted to move it out into a method:

```ruby
def manual
  @manual ||= Manual.find(manual_id, user)
end
```

spec/controllers/link_check_reports_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "spec_helper"
+
+describe LinkCheckReportsController, type: :controller do
+  describe "#create" do
+    let(:service) { double(:service, call: nil) }
```
thomasleese: Can we use the real service here?

spec/services/link_check_report/create_service_spec.rb
```diff
@@ -0,0 +1,217 @@
+require "spec_helper"
+
+RSpec.describe LinkCheckReport::CreateService do
+  let(:user) { FactoryGirl.create(:user) }
+
+  let(:link_check_report) { double(:link_check_report, valid?: nil, save!: nil) }
```
thomasleese: Could we use a real link check report here?

tsteer:
@thomasleese Hi Thomas, I've put the manuals and sections changes into one PR as you rightly pointed out manual_id and section_id should be used rather that link_reportable_type.

spec/controllers/link_check_reports_controller_spec.rb
lukemorton: We've mirrored the other controller specs. I'd rather leave this inline with how the rest of the application tests are structured. That said, I'd probably agree that using the real thing would be preferable. Perhaps an issue for another time?

spec/services/link_check_report/create_service_spec.rb
lukemorton: Ditto previos comment.

thomasleese: My concern here is that I can see some of the code was using fields like `link_reportable_type` which no longer exists, if this was a real object, the tests would have caught it.

spec/controllers/link_check_reports_controller_spec.rb
thomasleese: My concern here is that I can see some of the code was using fields like `link_reportable_type` which no longer exists, if this was a real object, the tests would have caught it.

thomasleese: Especially as the controllers feel like one level above from the model unit tests (and arguably they're more like integration tests) it seems like the controllers should be using the real models.

lukemorton: Totally agree though Tasmin has now addressed those comments.

Is there a place we can put these kinds of technical debt issues? It feels like we should update all the controller specs in this repo.

tsteer:
@thomasleese Updated this, does it look okay?

spec/controllers/link_check_reports_controller_spec.rb
thomasleese: Probably good enough as a GitHub issue, although since we're currently working in these controller tests I still reckon it's worth updating these ones in this PR.

lukemorton: @tsteer ^^

PR #1230, opened 2017-12-15
Add link check reports for manuals and sections

tsteer:
- Add link checker api to Services
- Add link check reports for sections

### DIFF
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -1,0 +1,18 @@
+class LinkCheckReportsController < ApplicationController
+  def create
+    service = LinkCheckReport::CreateService.new(
+      user: current_user,
+      manual_id: link_reportable_params[:manual_id]
+    )
+
+    service.call
+
+    head :created
+  end
+
+private
+
+  def link_reportable_params
+    params.require(:link_reportable).permit(:manual_id)
+  end
+end
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -2,7 +2,8 @@ class LinkCheckReportsController < ApplicationController
   def create
     service = LinkCheckReport::CreateService.new(
       user: current_user,
-      manual_id: link_reportable_params[:manual_id]
+      manual_id: link_reportable_params[:manual_id],
+      section_id: link_reportable_params[:section_id]
     )
 
     service.call
@@ -13,6 +14,6 @@ class LinkCheckReportsController < ApplicationController
 private
 
   def link_reportable_params
-    params.require(:link_reportable).permit(:manual_id)
+    params.require(:link_reportable).permit(:manual_id, :section_id)
   end
 end
--- a/app/services/link_check_report/create_service.rb
+++ b/app/services/link_check_report/create_service.rb
@@ -1,0 +1,63 @@
+require "services"
+require "govspeak/link_extractor"
+
+class LinkCheckReport::CreateService
+  class InvalidReport < RuntimeError
+    def initialize(original_error)
+      @message = original_error.message
+    end
+  end
+
+  def initialize(user:, manual_id:)
+    @user = user
+    @manual_id = manual_id
+  end
+
+  def call
+    link_report = call_link_checker_api
+
+    report = LinkCheckReport.new(
+      batch_id: link_report.fetch(:batch_id),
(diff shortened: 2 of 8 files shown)


app/services/link_check_report/create_service.rb
```diff
@@ -0,0 +1,77 @@
...
+
+private
+
+  attr_reader :user, :manual_id, :section_id
+
+  def manual
+    @manual ||= Manual.find(manual_id, user)
+  end
+
+  def section
```
thomasleese: We could add a `raise "Not a section" if section_id.nil?`.

spec/controllers/link_check_reports_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "spec_helper"
+require "services"
+
+describe LinkCheckReportsController, type: :controller do
+  describe "#create" do
+    let(:user) { FactoryGirl.create(:user) }
+    let(:manual) { double(:manual, id: 1, body: "[link](http://www.example.com)") }
```
thomasleese: Is there a factory we could use for this too?

spec/controllers/link_check_reports_controller_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "spec_helper"
+require "services"
+
+describe LinkCheckReportsController, type: :controller do
+  describe "#create" do
+    let(:user) { FactoryGirl.create(:user) }
```
thomasleese: Actually you can just use `create` here, no need for the `FactoryGirl`. (Will make it easier to upgrade to factory bot)

spec/services/link_check_report/create_service_spec.rb
```diff
@@ -0,0 +1,217 @@
+require "spec_helper"
+
+RSpec.describe LinkCheckReport::CreateService do
+  let(:user) { FactoryGirl.create(:user) }
+
+  let(:link_check_report) { double(:link_check_report, valid?: nil, save!: nil) }
```
thomasleese: Could we use a real object/factory here?

app/controllers/link_check_reports_controller.rb
```diff
@@ -0,0 +1,18 @@
+class LinkCheckReportsController < ApplicationController
+  def create
+    service = LinkCheckReport::CreateService.new(
```
thomasleese: Do we not need to support passing in a `section_id` here too?

spec/services/link_check_report/create_service_spec.rb
camillavk: I tried to use a factory here but because of the validation requirements it appears to be more effor than it's worth. I'm happy to get it working if it's required but I didn't feel like the time to get it working would be well spent

camillavk:
@thomasleese I think I've addressed all your comments on this PR

spec/services/link_check_report/create_service_spec.rb
thomasleese: Hmm, can we create a factory which satisfies the validation requirements?

app/services/link_check_report/create_service.rb
thomasleese: Hmm just thinking perhaps we should use `unless is_for_section?` here instead to avoid duplicating logic.

spec/controllers/link_check_reports_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+            suggested_fix: "example fix"
+          }
+        ]
+      }
+    end
+
+    before do
+      login_as_stub_user
+      allow(Services.link_checker_api).to receive(:create_batch).and_return(link_checker_api_response)
+      allow(Manual).to receive(:find).and_return(manual)
```
thomasleese: Could we use `FactoryGirl.create` so we have a real object and avoid the need to stub the `find`?

thomasleese: Here, and in a few other places.

spec/services/link_check_report/create_service_spec.rb
camillavk: Yup, I ended up having to do this anyway for my PR so I'll copy that code over into here then rebase when this gets merged in

camillavk:
@thomasleese I've addressed the other comments but kept the doubles and stubs of `find` methods as they were based on our conversation

PR #1233, opened 2017-12-19
Add UI for broken link check reports

camillavk:
- Include the Link Checker button on the Manuals show page
- Include suggested fix and summary of link problems with the broken links

This leads on from https://github.com/alphagov/manuals-publisher/pull/1230

screen shot 2017-12-19 at 09 13 08
screen shot 2017-12-19 at 09 41 48

### DIFF
--- a/app/models/manual.rb
+++ b/app/models/manual.rb
@@ -350,5 +350,9 @@ class Manual
     manual_record.set(attributes)
   end
 
+  def link_check_report
+    @link_check_report ||= LinkCheckReport.where(manual_id: id).last
+  end
+
   class RemovedSectionIdNotFoundError < StandardError; end
 end
--- a/app/views/admin/link_check_reports/_form.html.erb
+++ b/app/views/admin/link_check_reports/_form.html.erb
@@ -1,0 +1,3 @@
+<%= form_tag link_check_reports_path(link_reportable: { type: "manual", manual_id: reportable.id }), remote: true do %>
+  <%= submit_tag button_text, class: "btn btn-default add-top-margin remove-bottom-margin" %>
+<% end %>
--- a/app/assets/stylesheets/_broken-links-report.scss
+++ b/app/assets/stylesheets/_broken-links-report.scss
@@ -1,0 +1,41 @@
+.broken-links-report {
+  .issue-summary {
+    font-weight: normal;
+  }
+
+  .issue-list {
+    list-style: none;
+    padding-left: 0;
+
+    li {
+      margin-top: 5px;
+    }
+  }
+
+  .issue-status-description {
+    margin-top: 10px;
+  }
+
+  .issue-list + .issue-status-description {
+    margin: 15px 0;
+  }
+
+  .display-issue-details {
+    display: block;
+    font-weight: bold;
+    padding-top: 5px;
+  }
+
+  .display-issue-details::-webkit-details-marker {
+    display: none;
+  }
+
+  .display-issue-details::before {
+    content: '\25B6';
+    padding-right: 3px;
+  }
+
+  details[open] > .display-issue-details::before {
+    content: '\25BC';
+  }
+}
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -30,6 +30,7 @@
 @import "navbar";
 @import "header";
 @import "ordered_lists";
+@import "broken-links-report";
 
 p.no-content-message {
   @include core-19;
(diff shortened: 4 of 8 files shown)


app/models/manual.rb
```diff
@@ -16,7 +16,7 @@ class Manual
     :version_number,
     :updated_at,
     :originally_published_at,
-    :use_originally_published_at_for_public_timestamp,
+    :use_originally_published_at_for_public_timestamp
```
thomasleese: Would be good to leave this with a comma at the end.

app/models/manual.rb
```diff
@@ -350,5 +350,10 @@ def set(attributes = {})
     manual_record.set(attributes)
   end
 
+  def link_check_report
+    link_check_reports = LinkCheckReport.where(manual_id: id)
+    link_check_reports ? link_check_reports.first : nil
```
thomasleese: Can we memoize this?

app/assets/javascripts/_broken_links_report.js
```diff
@@ -0,0 +1,12 @@
+(function() {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  window.GOVUK.brokenLinksReport = {
+    init: function init() {
+      $('a.js-broken-links-refresh').each(function() {
```
nickcolley: Was this meant to be in this PR? It doesnt seem to do anything

app/helpers/javascript_helper.rb
```diff
@@ -0,0 +1,5 @@
+module JavascriptHelper
+  def initialise_script(constructor_or_singleton_name, params = {})
+    content_for :javascript_initialisers, raw("GOVUK.init(#{constructor_or_singleton_name}, #{params.to_json});\n");
```
nickcolley: This doesnt look right, there isnt a function to called init on here.

Doing eval of javascript like this is probably not a good idea either.

app/views/admin/link_check_reports/_form.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= form_tag link_check_reports_path(link_reportable: { type: "manual", manual_id: reportable.id }), remote: true do %>
```
nickcolley: Does this form make sense with only a button inside it?

app/views/admin/link_check_reports/create.js.erb
```diff
@@ -0,0 +1,2 @@
+$('.broken-links-report').replaceWith('<%=escape_javascript render("link_check_report", report: @report ) %>');
```
nickcolley: What is happening here?

app/models/manual.rb
tuzz: Won't `link_check_reports` always be non-`nil` because it's a scope of records?

Would this work `LinkCheckReport.find_by(manual_id: id)` instead of these two lines?

app/views/admin/link_check_reports/_form.html.erb
camillavk: I used a form for the button because it gets called three times inside https://github.com/alphagov/manuals-publisher/blob/832cf202d591e1f6d71527b1b77cc47d93a0adf5/app/views/admin/link_check_reports/_link_check_report.html.erb so I wanted to minimize code dulication

camillavk:
@nickcolley I've stripped out the javascript for now as you said it wasn't really necessary in this PR, it was pulled from the same functionality in Whitehall so some of it might return in a later PR but is not required just yet

SebAshton:
Hi @nickcolley do you have a moment to re-review this PR now the JS has been stripped out?

app/models/manual.rb
```diff
@@ -350,5 +350,10 @@ def set(attributes = {})
     manual_record.set(attributes)
   end
 
+  def link_check_report
+    return @link_check_report if defined? @link_check_report
```
thomasleese: You can replace this with `@link_check_report ||= ...`

SebAshton:
@nickcolley we could add an `aria-label` to the button to add extra information about the action it performs. Would that satisfy your concerns enough to unblock the PR?

nickcolley:
@SebAshton I think so, ideally test with a screenreader like VoiceOver to see if it makes sense. 👍

camillavk:
@nickcolley I've added the aria-label to the submit button:
screen shot 2018-01-04 at 11 50 03

screen shot 2018-01-04 at 11 50 13

camillavk:
@thomasleese @nickcolley I've removed the aria-label as per the discussion this afternoon

PR #1235, opened 2017-12-21
Add show endpoint for link check reports

SebAshton:
This PR adds the ability to get the status of a link check report based using its `id`. It provides both a JS and an HTML response for the show endpoint.

In addition to this, this PR also updates the `#create` specs to use Webmock and the gds api adapter helpers for the Link Checker API and it applies the same response options (JS and HTML) as `#show`

### DIFF
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -6,9 +6,14 @@ class LinkCheckReportsController < ApplicationController
       section_id: link_reportable_params[:section_id]
     )
 
-    service.call
+    @report = service.call
 
-    head :created
+    @reportable = find_reportable(link_reportable_params)
+
+    respond_to do |format|
+      format.js { render 'admin/link_check_reports/create' }
+      format.html { redirect_to_reportable }
+    end
   end
 
   def show
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -11,9 +11,43 @@ class LinkCheckReportsController < ApplicationController
     head :created
   end
 
+  def show
+    @report = LinkCheckReport::ShowService.new(
+      id: link_reportable_show_params[:id]
+    ).call
+
+    @reportable = find_reportable(section_id: @report.section_id, manual_id: @report.manual_id)
+
+    respond_to do |format|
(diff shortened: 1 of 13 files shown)


app/controllers/link_check_reports_controller.rb
```diff
@@ -6,14 +6,53 @@ def create
...
 
   def link_reportable_params
     params.require(:link_reportable).permit(:manual_id, :section_id)
   end
+
+  def link_reportable_show_params
+    params.permit(:id)
+  end
+
+  def find_reportable(reportable_params)
```
thomasleese: We already have something similar to this method in the `LinkCheckReport` class, are we able to reuse that at all?

SebAshton: I think we might something similar in the Whitehall version 🤔 ... The model here is [pretty barebones](https://github.com/alphagov/manuals-publisher/blob/master/app/models/link_check_report.rb).

thomasleese: Ah... I was thinking of the code here: https://github.com/alphagov/manuals-publisher/blob/1380109a7e1bcb5c49d49fb40a3c933e66119fb3/app/services/link_check_report/create_service.rb#L51

I wonder if we can extract them out into something else perhaps?

SebAshton: @thomasleese this change went in with e122429

SebAshton:
Depends on #1233 being merged in first

app/views/admin/link_check_reports/show.js.erb
```diff
@@ -0,0 +1,3 @@
+<% if @report.completed? %>
```
thomasleese: If the report hasn't been completed yet should we be showing a progress indicator? Or perhaps this does not deserve to be in this file?

thomasleese: If the report hasn't been completed yet should we be showing a progress indicator? Or perhaps this does not deserve to be in this file?

SebAshton: @thomasleese from my understanding the way the JS works/will work is if the poll to the show endpoint returns any content it executes the returned js. So, in this case, it will only replace the loading spinner with links/success if the report is complete, which is desired.

PR #1241, opened 2018-01-02
Use unicorn over WEBrick

adrianclay:
We wanted to use unicorn as it brings the environment closer to
what is used inside of production.

This also makes `startup.sh` boot the sidekiq workers.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
-web: bundle exec rails s -p 3205
+web: bundle exec unicorn -p ${PORT:-3205}
 worker: bundle exec sidekiq -C ./config/sidekiq.yml
--- a/startup.sh
+++ b/startup.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 bundle install
-bundle exec rails s -p 3205
+PORT=3205 bundle exec foreman start
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,4 +19,4 @@ ADD . $APP_HOME
 ARG COMPILE_ASSETS=false
 RUN if [ "$COMPILE_ASSETS" = "true" ] ; then bundle exec rails assets:precompile ; fi
 
-CMD bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p $PORT -b '0.0.0.0'"
+CMD bash -c "bundle exec foreman run web"


Procfile
```diff
@@ -1,2 +1,2 @@
-web: bundle exec rails s -p 3205
+web: bundle exec unicorn -p 3205
```
kevindew: Not such a concern on this app but will be on ones with draft stacks. Should we have this as: `${PORT:-3205}` so we can override this?

kevindew:
We'll also need to move foreman in the gemfile.

Recently I was caught in an issue with port numbers in Procfiles: https://github.com/alphagov/email-alert-api/commit/4357f3e9ef22d0687bb17f0281ce73a2864d47b0 I'm just looking into that a little to see whether this would be a factor in a conditional port number.

RyanMacG:
@kevindew I think the port issue you saw is down to using `foreman start` vs `foreman run`

kevindew:
@RyanMacG yup that's correct. I was wondering if that would cause problems with https://github.com/alphagov/govuk-puppet/blob/8e146d9942e82f3301fd51aa1df927ef4613e780/development-vm/Procfile where we use foreman start - but I think that's ok as the script already does `govuk_setenv` which sets up a $PORT env var.

kevindew:
The change looks good :+1:

I think with this change also affecting users running the app locally/dev vm we should make the commit message based more on justifying this as it gets a closer parity locally (as far as I'm aware there's not an advantage to running webrick locally and recent rails versions run puma in dev as default). It'd also be prudent to change [./startup.sh](https://github.com/alphagov/manuals-publisher/blob/master/startup.sh) to use `foreman start` since that would have parity with [bowl procfile](https://github.com/alphagov/govuk-puppet/blob/8e146d9942e82f3301fd51aa1df927ef4613e780/development-vm/Procfile#L69) and the current one will be missing workers.

startup.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 bundle install
-bundle exec rails s -p 3205
+bundle exec foreman start
```
kevindew: This won't run it with the correct port by default. We could prefix it with `PORT=3205` to solve that though.

PR #1252, opened 2018-01-09
Add discard draft manual functionality to ui

h-lame:
For: https://trello.com/c/PnoMy4LK/87-discard-manual

This adds a "Discard draft" button to manuals that have never been published.  Giving users access to functionality previously only available to devs using the console.

### DIFF
--- a/app/controllers/manuals_controller.rb
+++ b/app/controllers/manuals_controller.rb
@@ -125,6 +125,26 @@ class ManualsController < ApplicationController
     )
   end
 
+  def discard_draft
+    service = Manual::DiscardDraftService.new(
+      user: current_user,
+      manual_id: manual_id
+    )
+    result = service.call
+
+    if result.successful?
+      redirect_to(
+        manuals_path,
+        flash: { notice: "Discarded draft of #{result.manual_title}" }
+      )
+    else
+      redirect_to(
+        manual_path(manual_id),
+        flash: { notice: "Unable to discard draft of #{result.manual_title}" }
+      )
+    end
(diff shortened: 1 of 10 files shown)


app/controllers/manuals_controller.rb
```diff
@@ -125,6 +125,26 @@ def publish
...
+    result = service.call
+
+    if result.successful?
+      redirect_to(
+        manuals_path,
+        flash: { notice: "Discarded draft of #{result.manual_title}" }
+      )
+    else
+      redirect_to(
+        manual_path(manaul_id),
```
elenatanasoiu: manual_id*

elenatanasoiu: Can we get a test that checks for the unhappy path? We already have two scenarios:
a. a draft manual that has never been published is successfully discarded
b. a draft manual that has been published fails to be discarded

Perhaps a third scenario is necessary: 
c. a draft manual that has never been published that fails to be discarded

As an extra precaution that the redirect works, perhaps you could also check for the the flash message ("Unable to discard draft...") if the action has been unsuccessful? 

I'm pretty sure that's why the "manual_id" typo sneaked through.

h-lame: I don't have any controller specs, because there aren't many in this project, so yeah, there's no sad path test on this.  I'll see if it makes sense to add sad path feature spec or a controller spec.

h-lame: Added a controller spec in a rebase.

PR #1267, opened 2018-01-19
Add rake task to update the organisation of a manual

tomsabin:
Recently the Homes and Communities Agency organisation was closed and
are now known as Homes England. There is a manual that exists with the
old, closed, organisation and users that used to be able to access the
manual no longer can. This is because the manual still exists under the
Homes and Communities Agency and doesn't match the user scope.

The following rake task should be ran to update the organisation of the
manual so that users are able to view and edit the document again:

```
bundle exec rake update_manual_organisation[/guidance/capital-funding-guide,homes-england]
```

---

Related Zendesk ticket: https://govuk.zendesk.com/agent/tickets/2558744

### DIFF
--- a/lib/tasks/update_manual_organisation.rake
+++ b/lib/tasks/update_manual_organisation.rake
@@ -1,0 +1,32 @@
+require "logger"
+
+desc "Update organisation of a Manual"
+task :update_manual_organisation, %i(manual_base_path organisation_slug) => :environment do |_, args|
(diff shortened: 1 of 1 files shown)


tomsabin:
Deployed on integration and tested rake task (output visible at https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/93/console)

PR #1268, opened 2018-01-22
Send change notes to the publishing-api

thomasleese:
This PR means that change notes from sections will be sent to the Publishing API, and therefore picked up by the Email Alert Service.

[Trello Card](https://trello.com/c/5M9OTPir/517-get-change-notes-to-show-up-for-more-formats-email-updates)

### DIFF
--- a/app/adapters/publishing_adapter.rb
+++ b/app/adapters/publishing_adapter.rb
@@ -219,6 +219,7 @@ private
       update_type: update_type,
       publishing_app: GdsApiConstants::PublishingApiV2::PUBLISHING_APP,
       rendering_app: GdsApiConstants::PublishingApiV2::RENDERING_APP,
+      change_note: section.change_note,
       routes: [
         {
           path: "/#{section.slug}",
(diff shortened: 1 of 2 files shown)


app/adapters/publishing_adapter.rb
```diff
@@ -118,6 +120,7 @@ def save_manual_content(manual, republish: false)
       update_type: update_type,
       publishing_app: GdsApiConstants::PublishingApiV2::PUBLISHING_APP,
       rendering_app: GdsApiConstants::PublishingApiV2::RENDERING_APP,
+      change_note: latest_publication ? "#{latest_publication.title} - #{latest_publication.change_note}" : "",
```
kevindew: might we want this as nil or field not set (depending on which the Publishing API plays ball with) when there isn't a latest publication? Not sure if empty string leads to unintentional surprises or not. 

Also wonder if we risk having lots of items with same change note and I can't remember whether that would be a problem or not - I'm guessing specialist publisher already does that though

thomasleese: Yep, agreed, updated in 5df0153 to only set the field if the latest publication exists.

I don't think we have that risk as any major publication requires the section to have a change note.

PR #1314, opened 2018-03-26
Changes js to look for class instead of id

elenatanasoiu:
For: https://govuk.zendesk.com/agent/tickets/2617087


app/views/sections/_form.html.erb
```diff
@@ -86,12 +86,12 @@
 
   window.toggleDisplayWithCheckedInput({
     $input: $("#section_minor_update_1"),
-    $element: $("#change_note_container"),
+    $element: $(".change_note_container"),
```
fofr: Where classes are user for JS we should prefix with `.js-`

elenatanasoiu:
Thanks. Renamed to start with "js-".

PR #1323, opened 2018-04-04
Switch to using govuk_sidekiq

cbaines:


cbaines:
> I think most/all of the initialiser can be removed?

Good point, I think I forgot to check. I've now amended the first commit to remove that file.

PR #1342, opened 2018-06-01
Set primary publishing organisation

chao-xian:
This is set to be the same as the organisation.

As part of work to set `primary_publishing_organisation` on all content items so that we can slice data in the Content Performance Manager data-warehouse according to the organisation. 

Depends on https://github.com/alphagov/govuk-content-schemas/pull/764


chao-xian:
[(2) Set primary org in Manuals Publisher](https://trello.com/c/XZdljjFi/377-2-set-primary-org-in-manuals-publisher)

PR #1396, opened 2018-10-30
Add migration to fix broken manual

oscarwyatt:
Fixes manual marked as unpublished on the local database of manuals-publisher but publishing-api has successfully published the manual so it can't be updated as a minor change

PR #1394 was a previous, unsuccessful attempt at fixing the same problem

Ticket: https://govuk.zendesk.com/agent/tickets/3401170


db/migrate/20181030143621_another_force_update_broken_manual_published.rb
```diff
@@ -0,0 +1,22 @@
+class AnotherForceUpdateBrokenManualPublished < Mongoid::Migration
+  def self.up
+    user = User.find_by(email: "oscar.wyatt@digital.cabinet-office.gov.uk")
+    if user
+      manual = Manual.find("99005789-d82b-4045-932b-8ad9752a80bc", user)
```
leenagupte: This pains me, but I understand why you've had to do this.

PR #1567, opened 2020-01-31
Set org abbreviation to "" if nil

barrucadu:
Attempting to republish all the manuals fails part-way through with an
error from publishing-api about a nil abbreviation.  So if it's
missing, send an empty string instead.

The org in question is the Cabinet Office, which helpfully has the
acronym "" in its content item:

https://www.gov.uk/api/content/government/organisations/cabinet-office

### DIFF
--- a/app/adapters/publishing_adapter.rb
+++ b/app/adapters/publishing_adapter.rb
@@ -163,7 +163,7 @@ private
         organisations: [
           {
             title: organisation.title,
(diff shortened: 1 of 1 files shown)


sihugh:
I think there's a bunch of orgs with identical title/abbreviations too.  I think I remember one having an abbreviation that was longer than the title.

PR #1644, opened 2020-07-16
Upgrade to rails 6

cdccollins:
Main changes needed for this included removing the jasmine-rails gem, using libsass, and removing auto_load config.


spec/javascripts/support/jasmine.yml
```diff
@@ -1,50 +1,146 @@
...
+# by default, Port and AccessLog are passed.
+#
+# This is an advanced options, and left empty by default
+#
+# EXAMPLE
+#
+# rack_options:
+#   server: 'thin'
+
+# phantom_cli_options
```
Rhian, line 127: The commented PhantomJS references could possibly be removed?

Rhian:
I'd suggest two minor changes:

1. As per RailsDiff, config/puma.rb could be updated to replace:
`threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }
threads threads_count, threads_count`
with
`max_threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }
min_threads_count = ENV.fetch("RAILS_MIN_THREADS") { max_threads_count }
threads min_threads_count, max_threads_count`

2. There is a paragraph in the README about Cucumber and the bundler version which will be out of date once the updates in this PR are applied.

EGiataganas:
Hey @cdccollins this looks great, I've compared the pr changes with [railsdiff][1] and there are some changes that didn't happen with `bundle exec rails app:update` that could also be done:

1. Remove `bin/bundle`, `bin/update`
2. Update [system! "bin/rails db:setup"][2] with `system! "bin/rails db:prepare"`
3. In the `config/environments/test.rb` the `config.cache_classes = true` has changed to `false` and there is `config.action_view.cache_template_loading = true`
4. Remove `config/initializers/new_framework_defaults.rb`, config/initializers/new_framework_defaults_5_1.rb` as per the reason in alphagov/travel-advice-publisher@8b0745e. I also think we can remove , `config/initializers/rails_3_4_flash_compatibility.rb`
5. Update `app/views/layouts/application.html.erb`to include `csp_meta_tag` after `csrf_meta_tags`. Maybe it will be good to have a `csrf_meta_tags`, `csp_meta_tag`, `stylesheet_link_tag` and then `javascript_include_tag`
6. In the `config/spring.rb` the `%w[]` needs to be updated to `Spring.watch()`
```ruby
Spring.watch(
  ".ruby-version",
  ".rbenv-vars",
  "tmp/restart.txt",
  "tmp/caching-dev.txt"
)
```

[1]: http://railsdiff.org/5.2.4.3/6.0.3.2
[2]: https://github.com/alphagov/manuals-publisher/blob/master/bin/setup#L28

PR #1715, opened 2021-02-11
Limit the number of manuals on index page to 25

beccapearce:
Users with GDS editor permissions are able to see all manuals when
Manual.all is called. This causes the  index page to time out when
trying to load the page as a user with gds_editor permissions.

We have limited the number of manuals returned on the index page to
prevent this happening again.


bilbof (approved):
Great work 🎉

PR #1717, opened 2021-02-16
Improve query performance for Manual.all

bilbof:
## What

The intent of this change is to reduce the number of cursors used by the app for a single query.

We experienced [an issue impacting multiple apps](https://docs.google.com/document/d/1FWhRp4EV5PYss_hlEChiqFpP0Dv4PQlGq03K1lUh7FA/edit#) due to Manual Publisher making an N+1 query, consuming the maximum 450 cursors (per DocumentDB instance) causing other requests to fail.

## How

Rather than do an N+1 query for all manuals and their latest edition (hitting the DB 200+ times), we can use [eager loading](https://docs.mongodb.com/mongoid/current/tutorials/mongoid-queries/#eager-loading) to make 2 requests to the DB.

This will make the Manuals Publisher homepage (/manuals) load faster, and at prevent most requests from timing out.

Unfortunately eager loading pulls more editions out of the DB than required (we only need the latest edition, not all). However, it does save on extra round trips to the DB. This prevents Manuals Publisher maxing out the number of DocumentDB cursors, which is the problem we're trying to solve.

## Other solutions

* We should [add pagination to the manuals homepage](https://trello.com/c/amQiLOxa/1093-manuals-publisher-manuals-page-should-have-pagination), which solve this problem entirely.
* Ideally we'd do something like this, but it's not permitted:

    ```
     has_one :last_edition,
              class_name: 'ManualRecord::Edition',
              order: 'updated_at DESC'
    ```
    
    Perhaps swapping out MongoDB for Postgres might be a good idea.

### DIFF
--- a/app/models/manual.rb
+++ b/app/models/manual.rb
@@ -49,11 +49,19 @@ class Manual
   end
 
   def self.all(user, load_associations: true)
-    collection = user.manual_records
+    user.manual_records
+      .includes(:editions)
+      .all_by_updated_at
+      .lazy
+      .map do |manual_record|
+        edition = manual_record.editions.max_by(&:version_number)
 
-    collection.all_by_updated_at.lazy.map do |manual_record|
-      build_manual_for(manual_record, load_associations: load_associations)
-    end
+        build_manual_for(
+          manual_record,
+          edition: edition,
+          load_associations: load_associations,
+        )
+      end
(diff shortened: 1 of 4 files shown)


beccapearce (approved):
Awesome! Much speedier ⭐

PR #1833, opened 2022-03-03
Switch to using jasmine-browser-runner

BeckaL:
The Jasmine Ruby gems have been [deprecated](https://github.com/jasmine/jasmine-gem#discontinued) and are no longer supported. And whilst they still work with Rails 6 apps, they [aren't compatible with Rails 7](https://github.com/alphagov/govuk_publishing_components/issues/2621#issuecomment-1048007995) which is now becoming a blocker.

Note that this PR cannot be merged until Chromedriver has been installed on Jenkins - currently passing on CI as they happen to have run on a worker which has Chromedriver, but this isn't the case for all workers.


Rakefile
```diff
@@ -8,7 +8,7 @@ ManualsPublisher::Application.load_tasks
 task default: [
   "lint",
   "assets:precompile",
```
kevindew, line 10: Sorry I know you didn't add this line, but I think we should remove it. I'm not sure why someone added it before but our change will make this completely redundant. As part of the jasmine yarn task we run `assets:clobber` which would delete any assets this task compiled.

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,16 @@
...
+  "cssFiles": [
+    "application-*.css"
+  ],
+  "specDir": "spec/javascripts",
+  "specFiles": [
+    "**/*[sS]pec.js"
+  ],
+  "helpers": [],
+  "browser": "headlessChrome"
+}
```
kevindew, line 16: Looks like we're missing a new line character here - my pro tip would be to configure your text editor to always leave a new line on saving.

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,16 @@
+{
+  "srcDir": "public/assets",
+  "srcFiles": [
+    "govuk-admin-template-*.js",
+    "application-*.js"
+  ],
+  "cssFiles": [
+    "application-*.css"
+  ],
```
kevindew, line 9: These look to be the right files to test but I notice they don't match those of the previous configuration: https://github.com/alphagov/manuals-publisher/blob/c106827f4fbb7d9b46c442025ab8bad3ac9f8376/spec/javascripts/support/jasmine.yml#L13-L30

I think the previous version was configured a bit wrong though - perhaps someone trying to get themselves out of a bind. It'd be good if we could state in the commit message why we've configured it differently

kevindew (commented):
Looks great - just a couple of minor snags and some documentation to add. It'd also be good if the commit for jasmine swap over explained why we've needed to swap from Ruby Jasmine to Node Jasmine.

BeckaL:
@kevindew - have addressed your comments, do you mind having another quick look?

kevindew (approved):

PR #1883, opened 2022-06-07
Fixes previous edition ordering bug

1pretz1:
`previous_edition` was retuning the wrong edition as `.last` was ignoring
`limit(2)` and returning the last item in the non-limited ordered list,
which is the first version (desc ordering). Not sure if this was always
broken, or a result of a ruby / rails / mongoid upgrade.

Updates the code to return the actual previous edition by selecting the
second from the top.

Other uses of `limit` are working as expected within the app.


hannako (approved):
LGTM

PR #1894, opened 2022-06-29
Don't try and discard draft if it doesn't exist

1pretz1:
Fixes bug which returned the user an error when the section was not in a
draft state. This is possible either, if the user withdraws the section
via the UI, without editing the section first, due to it remaining in a
'published' state or when a section is withdrawn and redirected, the
state of the section is 'withdrawn'.

Both these scenarios are valid and so we shouldn't try and remove a
draft, unless there is one.

Trello:
https://trello.com/c/Xr6RU7UX/1363-fix-longstanding-bug-in-manuals-publisher


KludgeKML (approved):
LGTM

PR #1895, opened 2022-06-29
Add CSV to unpublish and redirect manuals

1pretz1:
### Dependancies 

**Task in https://github.com/alphagov/manuals-publisher/pull/1896 need to run first**

Adds temporary CSV to redirect and unpublish several manuals. This is needed as these manuals are supposed to be redirect but are in the wrong state, either in Manuals Publisher or downstream. We can now do this using rake tasks [1] which is much safer than attempting to do this as a one off task / from the console.

### Jenkins builds

- [x] [withdraw_and_redirect_manuals_to_multiple_paths:dry_run](https://deploy.blue.staging.govuk.digital/job/run-rake-task/218286/)
- [x] `withdraw_and_redirect_manuals_to_multiple_paths:real`, first run (https://deploy.blue.staging.govuk.digital/job/run-rake-task/218287/) and second (https://deploy.blue.staging.govuk.digital/job/run-rake-task/218289/). A space in a path in the CSV caused a validation error so had to run twice after pushing the fix.

[1]: https://github.com/alphagov/manuals-publisher/pull/1893

Trello:
https://trello.com/c/ZhJ2oj7N/1339-resolve-redirect-issues-for-manuals


KludgeKML (approved):
LGTM

PR #2003, opened 2022-12-21
Adding ability to add ASSETS_PREFIX ENV value.

nsabri1:
Same change that was applied to collections-publisher: https://github.com/alphagov/collections-publisher/commit/44cd9ccabeb51e4ef45947104de747352c193c8d

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ ARG builder_image=ghcr.io/alphagov/govuk-ruby-builder:3.1.2
 
 FROM $builder_image AS builder
 
+ENV ASSETS_PREFIX=/assets/manuals-publisher
+
 WORKDIR /app
 
 COPY Gemfile* .ruby-version /app/
@@ -16,7 +18,8 @@ RUN bundle exec rails assets:precompile && rm -fr /app/log
 
 FROM $base_image
 
-ENV GOVUK_APP_NAME=manuals-publisher
+ENV GOVUK_APP_NAME=manuals-publisher \
+    ASSETS_PREFIX=/assets/manuals-publisher
 
 WORKDIR /app
 
--- a/config/application.rb
+++ b/config/application.rb
@@ -37,5 +37,7 @@ module ManualsPublisher
     # to use CSS that has same function names as SCSS such as max.
     # https://github.com/alphagov/govuk-frontend/issues/1350
     config.assets.css_compressor = nil
+
+    config.assets.prefix = ENV.fetch("ASSETS_PREFIX", "/assets")
   end
 end


sengi (approved):
LGTM

kevindew:
Sorry to butt in.

I'm wondering if we gain anything by having this by env var configured. I think everything would still work in all environments if we just set `config.assets.prefix = "/assets/manuals-publisher"` which would save the complexity of this becoming part of environmental configuration.

I'd also somewhat wonder if it was slightly neater to keep the path as the Rails defaults "/assets" and use the web-server config to proxy that to your bucket location, but I'm guessing the horse might have bolted on that?

nsabri1:
> Sorry to butt in.
> 
> I'm wondering if we gain anything by having this by env var configured. I think everything would still work in all environments if we just set `config.assets.prefix = "/assets/manuals-publisher"` which would save the complexity of this becoming part of environmental configuration.
> 
> I'd also somewhat wonder if it was slightly neater to keep the path as the Rails defaults "/assets" and use the web-server config to proxy that to your bucket location, but I'm guessing the horse might have bolted on that?

Sorry as my Ruby knowledge is not great - changing the default to `config.assets.prefix = "/assets/manuals-publisher"` will mean a change to the way the current EC2 gets its assets, from:
`https://manuals-publisher.integration.publishing.service.gov.uk/assets/application-718198c0533914125bdeb775e5500631bad6333ddf463ac4c89af5eae51dcefa.js`
to
`https://manuals-publisher.integration.publishing.service.gov.uk/assets/colleections-publisher/application-718198c0533914125bdeb775e5500631bad6333ddf463ac4c89af5eae51dcefa.js`
personally io would like to avoid that with this change.

kevindew:
Yup would mean that the asset URLs change in EC2 but would save us having two sets of URLs. What would your concern be about changing them in EC2?

My concern is that by not changing it everywhere we actually make things more complicated by having something configurable and different between environments, which - based on past precedent - could get left around as a complication for years.

nsabri1:
> Yup would mean that the asset URLs change in EC2 but would save us having two sets of URLs. What would your concern be about changing them in EC2?
> 
> My concern is that by not changing it everywhere we actually make things more complicated by having something configurable and different between environments, which - based on past precedent - could get left around as a complication for years.

My concern is around changing the default behaviour which may impact in away i can't tell from initials tests and may not be covered by our test suite(Smokey) etc, baring in mind i am unsure how the current EC2 platform works with Assets.

You concern is totally valid one - if you are happy to test the change in ec2 alongside me  then i would be happy to try your suggestion.

kevindew:
> > Yup would mean that the asset URLs change in EC2 but would save us having two sets of URLs. What would your concern be about changing them in EC2?
> > My concern is that by not changing it everywhere we actually make things more complicated by having something configurable and different between environments, which - based on past precedent - could get left around as a complication for years.
> 
> My concern is around changing the default behaviour which may impact in away i can't tell from initials tests and may not be covered by our test suite(Smokey) etc, baring in mind i am unsure how the current EC2 platform works with Assets.
> 
> You concern is totally valid one - if you are happy to test the change in ec2 alongside me then i would be happy to try your suggestion.

Yup sure. I'm happy to help you test it.

sengi:
> I'm wondering if we gain anything by having this by env var configured. I think everything would still work in all environments if we just set config.assets.prefix = "/assets/manuals-publisher" which would save the complexity of this becoming part of environmental configuration.

That's a great suggestion! If we can just set `assets.prefix` without any side effects in Puppet land then yeah let's ditch the env var - sounds like a win to me 👍 (Now that you mention it, I vaguely remember we chatted about this ages ago but I must have forgotten in the meantime 😅)

> I'd also somewhat wonder if it was slightly neater to keep the path as the Rails defaults "/assets" and use the web-server config to proxy that to your bucket location, but I'm guessing the horse might have bolted on that?

Ah I think I can see where you're coming from on this one, but there are some compelling reasons for configuring this in Rails rather than further upstream in the request path.

- Rewriting these URLs upstream in order not to change the Rails default adds a little bit of unnecessary complexity to system vs. the least-surprise option of having the paths stay the same all the way through (`manuals-publisher.publishing.service.gov.uk/assets/manuals-publisher/this-is-fine.gif` goes to the same path on the backend and the same path on the app pod. Not a big deal by itself, though it does make it just that little bit harder to understand and faultfind the serving path. The real downside is that this fluffy gremlin would multiply when we come to put the publisher apps behind the CDN and serve assets via the CDN->S3 fast path instead of going via the origin LB and app nginx. So there's also some loss of generality/flexibility, which leads to accretion of more complexity over time.
- nginx config is notoriously hard to debug, test and reason about (because the language has more than its fair share of quirks and foot-cannons and there's little in the way of observability / debugging tools in the OSS version). It therefore pays to keep a tight rein on complexity there and avoid introducing any logic unless we're pretty convinced that it's the lesser of evils.
- This use case seems to me a good fit for Rails's `assets.prefix` and friends, whose purpose is presumably to provide the flexibility to support things like having multiple apps behind a CDN or other shared L7 routing components, or indeed any scenario where Rails isn't directly involved in serving assets.
- We don't have any to gain by *not* making use of Rails's `assets.prefix` setting, as far as I can tell. Am I missing something?

kevindew:
> That's a great suggestion! If we can just set `assets.prefix` without any side effects in Puppet land then yeah let's ditch the env var - sounds like a win to me 👍 (Now that you mention it, I vaguely remember we chatted about this ages ago but I must have forgotten in the meantime 😅)

Yup. Puppet currently configures nginx to set files in /assets to have long expiry times and gzip encoding. As this change is just a different specificity within that directory they'll continue to be served with the same header.


> Ah I think I can see where you're coming from on this one, but there are some compelling reasons for configuring this in Rails rather than further upstream in the request path.
> 
> * Rewriting these URLs upstream in order not to change the Rails default adds a little bit of unnecessary complexity to system vs. the least-surprise option of having the paths stay the same all the way through (where e.g. `manuals-publisher.publishing.service.gov.uk/assets/manuals-publisher/this-is-fine.gif` goes to the same path on the backend and the same path on the app pod). Not a big deal by itself, though it does make it just that little bit harder to understand and faultfind the serving path. The real downside is that this small fluffy gremlin would multiply when we come to put the publisher apps behind the CDN and serve assets via the CDN->S3 fast path instead of going via the origin LB and app nginx. So there's also some loss of generality/flexibility, which leads to accretion of more complexity over time.
> * nginx config is notoriously hard to debug, test and reason about (because the language has more than its fair share of quirks and foot-cannons and there's little in the way of observability / debugging tools in the OSS version). It therefore pays to keep a tight rein on complexity there and avoid introducing any logic unless we're pretty convinced that it's the lesser of evils.
> * This use case seems to me a good fit for Rails's `assets.prefix` and friends, whose purpose is presumably to provide the flexibility to support things like having multiple apps behind a CDN or other shared L7 routing components, or indeed any scenario where Rails isn't directly involved in serving assets.
> * As far as I can tell, we don't have anything to gain by _not_ making use of Rails's `assets.prefix` setting. (Do we? Am I missing something?)

Sure, that all sounds quite reasonable 👍 And can see how it's more 12 factor. 

My concerns were rather modest: felt like a lower level part of stack was dictating a higher level which felt like a potentially delicate coupling and, on the extremely modest, it just looks a bit weird that you have path that then repeats a chunk of the hostname, which feels like superfluous information.

nsabri1:
This can now be closed due to:
https://github.com/alphagov/govuk-puppet/pull/11945 
https://github.com/alphagov/manuals-publisher/pull/2004#pullrequestreview-1227877337

PR #2077, opened 2023-06-21
Add GDS system & layout for transition

CL-GDS:
**What**
Add design_system.html.erb for layout
Add select_with_search_helper.html.erb required by layout
Extend application.scss to include GDS components
Extend application_controller.rb to include preview_design_system & render_design_system functions
Extend manuals_controller.rb with get_layout function and implemented layout
Extend user.rb to include PermissionChecker for the release flags
**Why**
Enable porting of Manuals pages to GDS to improve Manuals Publisher functionality & mirror similar styles to Whitehall for User & Developer consistency. Adding design system flags to allow incremental functionality improvement's without visual page changes. This could also be done without the get_layout and flags by adjusting to `render "design_system"` but this would require all page views to be done at once
**Trello**
https://trello.com/c/M0crhdKs/314-move-pages-to-gov-design-system

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,6 +1,9 @@
 // govuk_admin_template
 @import "govuk_admin_template";
 
+// govuk_components
+@import "govuk_publishing_components/all_components";
+
 // govuk_frontend_toolkit
 @import "colours";
 @import "typography";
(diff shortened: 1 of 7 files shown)


CL-GDS:
Temp close til ticket is in priority

PR #2082, opened 2023-06-22
Add Gov Design system & layout for transition

CL-GDS:
Add GDS System & Layout for page transitions

## What

Add design_system.html.erb for layout
Add select_with_search_helper.html.erb required by layout Extend application.scss to include GDS components
Extend application_controller.rb to include preview_design_system & render_design_system functions 
Extend user.rb to include PermissionChecker for the release flags 

## Why

Enable porting of Manuals pages to GDS to improve Manuals Publisher functionality & mirror similar styles to Whitehall for User & Developer consistency. Adding design system flags to allow incremental functionality improvement's without visual page changes. This could also be done without the get_layout and flags by adjusting to render "design_system" but this would require all page views to be done at once 

## Trello

https://trello.com/c/M0crhdKs/314-move-pages-to-gov-design-system


app/assets/stylesheets/application.scss
```diff
@@ -1,6 +1,9 @@
 // govuk_admin_template
 @import "govuk_admin_template";
 
+// govuk_components
+@import "govuk_publishing_components/all_components";
```
kevindew, line 5: When using govuk_publishing_components it is conventional to use the components that you need rather than all as it reduces CSS size, could you switch to that? see: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/install-and-use.md#4-include-the-assets

app/views/layouts/design_system.html.erb
```diff
@@ -0,0 +1,54 @@
+<% environment = GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment %>
```
kevindew, line 1: Rather than adding a new layout we should use the component for admin layout: https://components.publishing.service.gov.uk/component-guide/layout_for_admin

lib/govuk_publishing_components/presenters/select_with_search_helper.rb
```diff
@@ -0,0 +1,84 @@
+##
```
kevindew, line 1: I don't know why we've got this code here nothing seems to be using it? I also don't understand why we'd add code under the GovukPublishingComponents namespace here 😕

app/lib/permission_checker.rb
```diff
@@ -1,7 +1,8 @@
 class PermissionChecker
   GDS_EDITOR_PERMISSION = "gds_editor".freeze
   EDITOR_PERMISSION = "editor".freeze
-
+  PREVIEW_DESIGN_SYSTEM = "Preview design system".freeze
+  PREVIEW_NEXT_RELEASE = "Preview next release".freeze
```
kevindew, line 5: Could you explain in the commit message that adds these what you intend to use them for? I have a vague understanding from working on Whitehall and without that wouldn't understand at all - so I imagine anyone other than me wouldn't understand. Even with my understanding I don't know how you intend to approach transitioning Manuals so would be good to have documented at what points we'd use either of these?

(I also somewhat imagine you'll only need one permission for Manuals? is the design system transition going to be shipped in multiple releases?)

app/controllers/application_controller.rb
```diff
@@ -14,6 +14,22 @@ class ApplicationController < ActionController::Base
...
+  end
+  helper_method :preview_design_system?
+
+  def render_design_system(design_system_view, legacy_view, locals)
+    if get_layout == "design_system"
+      render(design_system_view, locals:)
+    else
+      render legacy_view, locals:
+    end
+  end
```
kevindew, line 31: Is this all copied and pasted from Whitehall? I was quite confused by the need for a temporary forcing - do we actually need to temporarily force anything? I'm guessing Whitehall reasons for temporary forcing something and this apps might differ?

Also seems a bit strange to be adding this render_design_system code with a get_layout method when nothing calls the render_design_system method and get_layout is not defined anywhere (as an aside, can we avoid a `get_layout` method it's not very idiomatic Ruby to have a getter)

kevindew (commented):
I've added a few suggestions of our conventional approaches.

I'm a bit confused about this PR though as it seems to be some foundations that aren't used yet and it isn't clear to me what we need for the future yet. This could be resolved by breaking this up into more commits and documenting clearer intent.

I noticed you tagged @davidgisbey as a reviewer, but he has no context of this work. Could you keep me as the reviewer for the changes to Manuals Publisher until further notice? It would be good though if prior to my reviews your work could be reviewed by your team members.

CL-GDS:
Closing as replaced by PR #2090 as the need to notifiy users was more of a priority

PR #2099, opened 2023-07-07
Consolidation paste to govspeak squashed

mtaylorgds:
## What

Improve the Manuals Publisher UX by automatically converting HTML pasted into the body fields of manuals and sections into govspeak.

This PR is a slightly modified version of [PR-2083](https://github.com/alphagov/manuals-publisher/pull/2083) that was merged into [consolidation-main](https://github.com/alphagov/manuals-publisher/tree/consolidation-main). The changes from that PR:
- removing the autogovspeak from the summary fields (since they are just plain text fields)
- added a feature test for pasting into the section body
- rebased on latest main.

## Why

Provides a quality of life update to users.

## Trello

https://trello.com/c/VSFs9h05/306-auto-govspeak

## Pics

H1 HTML

Screenshot 2023-06-22 at 15 12 40

Before paste

Screenshot 2023-06-22 at 15 12 57

After paste converted to Govspeak

Screenshot 2023-06-22 at 15 13 53

### DIFF
--- a/features/step_definitions/manual_steps.rb
+++ b/features/step_definitions/manual_steps.rb
@@ -278,9 +278,9 @@ end
 When(/^I add another section and publish the manual later$/) do
   create_section(
     @manual.title,
-    section_title: "Another section so we can publish",
-    section_summary: "Another section so we can publish summary",
-    section_body: "Another section so we can publish body",
+    { section_title: "Another section so we can publish",
+      section_summary: "Another section so we can publish summary",
+      section_body: "Another section so we can publish body" },
   )
   go_to_manual_page(@manual.title)
   publish_manual
@@ -665,6 +665,24 @@ When(/^I start creating a new manual$/) do
   }
 
   create_manual(@manual_fields, save: false)
+end
+
+When(/^I start creating a section for the manual$/) do
+  create_section(@manual_fields.fetch(:title), {}, save: false)
+end
+
+When(/^I paste HTML into the manual body$/) do
+  fill_in_field("body", "")
+  page.execute_script(javascript_to_simulate_paste("manual_body", "<h2>Benefits of following this advice</h2>"))
+end
+
+When(/^I paste HTML into the section body$/) do
+  fill_in_field("section body", "")
+  page.execute_script(javascript_to_simulate_paste("section_body", "<h2>Benefits of following this advice</h2>"))
+end
+
+Then(/^the (manual|section) body field contains govspeak$/) do |content_type_name|
+  expect(page).to have_field("#{content_type_name}_body", with: "## Benefits of following this advice")
 end
 
 When(/^I preview the manual$/) do
--- a/features/support/form_helpers.rb
+++ b/features/support/form_helpers.rb
@@ -15,6 +15,18 @@ module FormHelpers
     end
   end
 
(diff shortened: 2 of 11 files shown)


features/support/form_helpers.rb
```diff
@@ -15,6 +15,18 @@ def fill_in_field(field_name, value)
     end
   end
 
+  def javascript_to_simulate_paste(element_id, html_to_paste)
```
kevindew, line 18:
```suggestion
  def javascript_to_simulate_paste(element_id, paste_content)
```

Minor: there isn't really any need for this function to reference html in the argument

features/step_definitions/manual_steps.rb
```diff
@@ -278,9 +278,9 @@
 When(/^I add another section and publish the manual later$/) do
   create_section(
     @manual.title,
-    section_title: "Another section so we can publish",
-    section_summary: "Another section so we can publish summary",
-    section_body: "Another section so we can publish body",
+    { section_title: "Another section so we can publish",
+      section_summary: "Another section so we can publish summary",
+      section_body: "Another section so we can publish body" },
```
kevindew, line 283: It's not clear to me why this was changed? does it fix a problem? if so it probably should be a separate commit.

yarn.lock
```diff
@@ -1894,6 +1894,11 @@ parseurl@~1.3.3:
   resolved "https://registry.yarnpkg.com/parseurl/-/parseurl-1.3.3.tgz#9da19e7bee8d12dff0513ed5b76957793bc2e8d4"
   integrity sha512-CiyeOxFT/JZyN5m0z9PfXw4SCBJ6Sygz1Dpl0wqjlhDEGGBP1GnsUVEL0p63hoG1fcj3fHynXi9NYO4nWOL+qQ==
 
+paste-html-to-govspeak@^0.3.0:
+  version "0.3.0"
+  resolved "https://registry.yarnpkg.com/paste-html-to-govspeak/-/paste-html-to-govspeak-0.3.0.tgz#9c4717690f3d6cd290c972c9bf106ead14adb177"
+  integrity sha512-K4jcJkZLQpvpukkiSzXZcXPbV80CDSKTI3zEs0/gVG+jx5H7v2XsdIOEuCZ56cRCCuuIH3rsrJcyoleGqPj6xw==
+
```
kevindew, line 1901: Nice one removing the churn of npmjs links

kevindew (commented):
Looks great - just a couple of minor things to resolve.

features/step_definitions/manual_steps.rb
danmooregds: It's necessary to fix error from refactoring to support manuals + sections work - a hint to ruby parser 

I believe the error was caused to adding an optional "save" parameter to create_section, along the lines of existing create_manual method, which means these existing parameters need to be passed as an explicit hash not gathered into a hash by default ruby parsing.

Other call sites don't get caught by this error as they reference the hash of fields by variable rather than automatic gathering in this one call.

kevindew: Ah yes, that makes sense, now you've introduced a keyword argument on the method it needs this to be distinct from them. It'd be useful having the information you've provided there with the commit.

kevindew (approved):

PR #2117, opened 2023-07-27
Add confirm publish warning page

mtaylorgds:
## What
Replaces the unclear alert popup when clicking the publish manual button with a new page that checks the user is aware that all draft sections will be published.

This new page uses the GDS design system, as part of the migration of manuals publisher to using that throughout.

Also fixes the styling of the flash messages that were appearing unstyled.

## Why

This change is to help make it clearer to users that it is not just their changes that will be published, but any other draft changes to sections that may have been made by others. An enhancement to this page will follow that will display the sections that this affects.

## Visuals
Clicking the publish button:
image

Would show this uninformative pop-up:
image

That the user would then click OK on to confirm publication.

With the changes in this PR, it will instead take the user to a confirmation page:
manuals-publisher dev gov uk_manuals_b655b49a-6ac6-4b66-bcf4-c3f0b4510900_confirm-publish

Where clicking publish will actually perform the publish operation (as clicking OK on the previous alert).

Once published, the success message, which was appearing as un-styled text:
image

Will now appear styled as it was intended to be:
image

## Trello
[Trello story](https://trello.com/c/BS9mtCv2)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -43,13 +43,13 @@ module ApplicationHelper
 
   def bootstrap_class_for(flash_type)
     case flash_type
-    when :success
+    when "success"
       "alert-success" # Green
-    when :error
+    when "error"
       "alert-danger" # Red
-    when :alert
+    when "alert"
       "alert-warning" # Yellow
-    when :notice
+    when "notice"
       "alert-info" # Blue
     else
       flash_type.to_s
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -84,6 +84,10 @@ module ApplicationHelper
     "#{Plek.external_url_for('draft-origin')}/#{manual.slug}"
   end
 
+  def allow_publish?(manual, slug_unique)
+    manual.draft? && manual.sections.any? && current_user_can_publish? && slug_unique
+  end
+
   def publish_text(manual, slug_unique)
     if manual.state == "published"
       text = "<p>There are no changes to publish.</p>"
--- a/app/views/manuals/confirm_publish.html.erb
+++ b/app/views/manuals/confirm_publish.html.erb
@@ -1,0 +1,39 @@
+<% content_for :page_title, manual.title %>
+<% content_for :title, manual.title %>
+
+<% content_for :title_margin_bottom, 6 %>
+
+<%= render "govuk_publishing_components/components/breadcrumbs", {
+  collapse_on_mobile: true,
+  breadcrumbs: [
+    {
+      title: "Your manuals",
+      url: manuals_path
+    },
+    {
+      title: manual.title,
+      url: manual_path(manual)
+    },
+    {
+      title: "Publish"
+    },
+  ]
+} %>
+
+<%= render "govuk_publishing_components/components/title", {
+  title: "Publish #{manual.title}"
+} %>
+
+<p class="govuk-body govuk-!-margin-bottom-7">You are about to publish "<%= manual.title %>". All the sections which
+  are in draft status will be published.</p>
+
+<%= form_tag(publish_manual_path(manual), method: :post) do %>
+  <div class="govuk-button-group govuk-!-margin-bottom-6">
+    <%= render "govuk_publishing_components/components/button", {
+      text: "Publish",
+      name: "submit",
+    } %>
+
+    <%= link_to("Cancel", manual_path(manual), class: "govuk-link govuk-link--no-visited-state") %>
+  </div>
+<% end %>
--- a/features/publishing-a-manual.feature
+++ b/features/publishing-a-manual.feature
@@ -8,7 +8,9 @@ Feature: Publishing a manual
 
   Scenario: Publish a manual
     Given a draft manual exists with some sections
-    When I publish the manual
+    When I click the publish manual button
+    Then I am asked to confirm the publishing
+    When I confirm publishing the manual
     Then the manual and all its sections are published
     And I should see a link to the live manual
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,6 +1,8 @@
 @import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/component_support";
 
+@import "govuk_publishing_components/components/breadcrumbs";
+@import "govuk_publishing_components/components/button";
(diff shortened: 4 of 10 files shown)


features/publishing-a-manual.feature
```diff
@@ -8,7 +8,9 @@ Feature: Publishing a manual
 
   Scenario: Publish a manual
     Given a draft manual exists with some sections
-    When I publish the manual
+    When I click the publish manual button
+    Then I am shown a warning page
```
kevindew, line 12:
```suggestion
    Then I am asked to confirm the publishing
```

I think "warning page" isn't the right phrasing for this type of page as it is a confirm step. I also think we should be a bit more specific in the Cucumber step as this step isn't for a generic warning (or confirm) page it's strictly for the publishing confirmation

app/helpers/application_helper.rb
```diff
@@ -43,13 +43,13 @@ def publication_task_state(task)
...
+    when "success"
       "alert-success" # Green
-    when :error
+    when "error"
       "alert-danger" # Red
-    when :alert
+    when "alert"
       "alert-warning" # Yellow
-    when :notice
+    when "notice"
```
kevindew, line 52: Nice catch. 

Are we at risk though that any uses of symbol may fail, e.g here: https://github.com/alphagov/manuals-publisher/blob/e01ac22466d1863486a01c81b1bf0f443bf9ad4f/app/controllers/manuals_controller.rb#L25

app/helpers/application_helper.rb
```diff
@@ -84,6 +84,10 @@ def content_preview_url(manual)
     "#{Plek.external_url_for('draft-origin')}/#{manual.slug}"
   end
 
+  def allow_publish?(manual, slug_unique)
```
kevindew, line 87: Since we can derive slug_unique within this method (by calling `manual.slug_unique?(current_user)`) could the slug_unique argument be superfluous?

Though my subsequent comment may render this irrelevent.

app/views/manuals/confirm_publish.html.erb
```diff
@@ -29,10 +29,12 @@
...
-    <%= render "govuk_publishing_components/components/button", {
-      text: "Publish",
-      name: "submit",
-    } %>
+    <% if allow_publish?(manual, slug_unique) %>
+      <%= render "govuk_publishing_components/components/button", {
+        text: "Publish",
+        name: "submit",
+      } %>
+    <% end %>
```
kevindew, line 37: It seems an unusual pattern to show the user the confirmation page when they have nothing to confirm - it also seems a not very friendly user pattern to hide the button (our designers often have strong opinions on this as hidden buttons is not good for knowing what you _could_ do)

What' I'd suggest is that rather than having this page deal with this error scenario we should instead check for it in the controller and redirect them back to the manual page with a flash error saying why they can't publish.

kevindew (commented):
Nice job - I've made a few suggestions

app/helpers/application_helper.rb
mtaylorgds: I considered this, but wasn't sure about having the view call a method that is ultimately doing a DB query. The code is basically just moved as-is from the show view, which is another reason I did it this way.

app/views/manuals/confirm_publish.html.erb
mtaylorgds: The user shouldn't really get to this page without being able to publish (unless they've bookmarked it, or used their browser history to get back to it and the manual's state has changed in the meantime), so we thought a simple hide of the button would suffice in this case (even though it's a sub-optimal user-experience in that case), but we can implement the suggested logic in the controller.

kevindew: I'd do it in the controller - we have consistent patterns for that whereas this approach would not be an unusual pattern.

app/helpers/application_helper.rb
kevindew: Sure I can see your logic - we don't tend to treat methods as special if they call the DB and tend to have a bunch of methods hit the DB at rendering time (particularly with some of Rails' lazy loading). So in general, if we'd be concerned about performance we'd likely do memoising in the method to ensure queries aren't made multiple times.

app/helpers/application_helper.rb
mtaylorgds: Doesn't seem to be—the `class` of the `type` variable is `String`, even in the instance you mention.

I don't really understand why, but that's what my empirical testing shows…

app/helpers/application_helper.rb
mtaylorgds: Fair enough.

app/views/manuals/confirm_publish.html.erb
mtaylorgds: Will do that then.

app/helpers/application_helper.rb
kevindew: Oh that's a bit strange. Does that mean that this code didn't work before? since it couldn't have worked for string and symbol as far as I'm aware 🤔 

I'll just have a quick look

app/views/manuals/confirm_publish.html.erb
mtaylorgds: Is a generic flash error message about not being able to publish the manual good enough, or should we specify the reason (which will mean reworking the `allow_publish?` method to not just return a boolean result)?

mtaylorgds: OK, we've added a generic message, and addressed the other comments now.

app/helpers/application_helper.rb
mtaylorgds: We've changed this to use an optional parameter so as not to change how the existing show view was working (in case there was a good reason for that to be doing the check once and passing the value around).

For the `confirm_publish` controller method, it calls it without passing a second argument, and the helper will default to calling the method on the manual.

kevindew: Cool - that seems a neat compromise. I do think though the code ends up a bit heavy on the wtf factor having such an involved default argument so I think it'd be better to just passed in in both places. 

Sorry about the to and fro.

mtaylorgds: Yeah, that's fair. Have now reverted the parameter back to being required and passing the argument value in both places.

kevindew (approved):
Great, looks good. Happy for me to merge?

PR #2124, opened 2023-08-03
Indicate which sections are in draft

mtaylorgds:
## What

Update the manual's "show" view to indicate which sections are currently in draft. 

## Why

Enables manuals editors to see at a glance which sections are being edited (and thus will be updated when the manual is published), which is useful when multiple people are editing the same manual.

This will soon be enhanced to include the name of the person who last updated a section that is in draft status to further help coordination between manual editors.

## Visuals

image

## Trello

[Trello card](https://trello.com/c/cQVYqYIs)

### DIFF
--- a/app/models/section.rb
+++ b/app/models/section.rb
@@ -36,6 +36,7 @@ class Section
            :minor_update,
            :exported_at,
            :visually_expanded,
+           :state,
            to: :latest_edition
 
   attr_reader :uuid
--- a/app/views/manuals/show.html.erb
+++ b/app/views/manuals/show.html.erb
@@ -69,6 +69,9 @@
         <ul class="document-list">
          <% manual.sections.each do |section| %>
             <li class="document">
+              <% if section.state == "draft" %>
+                <span class="label label-primary">draft</span>
+              <% end %>
               <%= link_to(section.title, manual_section_path(manual, section), class: 'document-title') %>
               <ul class="metadata">
                 <li class="text-muted">Updated <%= time_ago_in_words(section.updated_at) %> ago</li>


kevindew:
I notice this is a draft, so is there more to come?

Otherwise, this looks good to me. I wondered about questioning if there is some testing we should do around this at all but concluded it doesn't need any as the view logic is so minor.

If I understand right, the state machine allows 3 possible states: draft, published and archive - do we need to consider a label for archive too (or, as might well be the case, are they not shown in the list?)

mtaylorgds:
> I notice this is a draft, so is there more to come?

I don't think it is a draft? It's not supposed to be and doesn't appear as one when I view it 🤔 
There is more to come for the story (to display the user who last made changes to a draft section), but I figured this PR provides some user-value already and so could be merged.

> Otherwise, this looks good to me. I wondered about questioning if there is some testing we should do around this at all but concluded it doesn't need any as the view logic is so minor.

Yeah, there's not really any rspec tests for views, and the cucumber tests seemed a little heavy-weight for something so simple.

> If I understand right, the state machine allows 3 possible states: draft, published and archive - do we need to consider a label for archive too (or, as might well be the case, are they not shown in the list?)

I'm not really sure where the archive state fits in—I don't really understand the state machine code in there yet (as an aside that state_machine gem hasn't been updated in 10 years…). Functionally you can withdraw a section, at which point it disappears from the list. I'll have to dig into this further, and see if there is a way for a section to appear in the list in an "archive" state.

kevindew:
Sorry you're totally right - I thought it was a draft PR but clearly I got confused multitasking as it is not.

kevindew:
Oh I think it's because the title says "draft" and I did a 2 + 2 = 5 🤦

kevindew (approved):

PR #2125, opened 2023-08-04
Display draft sections' most recent editor

mtaylorgds:
## What

Update the view for manuals to show, for draft sections, who last made changes to that section.

This will only affect sections that are edited after this change is merged, since we'll have no record of edits made before then.

## Why

This change improves the experience when there are multiple people editing a manual as it makes it possible to see what sections are being edited by other people, thus improving the ability to coordinate new releases of a manual.

### Before

image

### After

image

## Trello

[Trello story](https://trello.com/c/cQVYqYIs)

## Implementation notes

Once published, it isn't really necessary to know who last edited a section. We could delete the user's details at the point of publish, but that is an extra complexity that doesn't really seem necessary.

We store the user's name directly as an attribute on the section, rather than the their UUID, in order to keep the implementation simple and avoid having to 'join' the data together every time a section is loaded. Especially interested in feedback on this decision, and whether storing the UUID (and then having to do a 'join' for display purposes) would be considered worth the extra development time.


ollietreend (approved):
This looks good to me @mtaylorgds 👍🏻 

Thanks for creating such a clear and descriptive PR and set of commits.

Regarding your decision to store the user's name directly in the section, rather than their UUID and then 'joining' to get their name, I don't have a strong opinion either way. The approach you've taken does seem more fitting given this app is backed by a NoSQL (Mongo) database – so it's more common to denormalise data to support your read patterns. If this app were backed by a relational database (e.g. Postgres) then I'd likely be leaning in the other direction. The approach you've chosen does seem reasonable to me in this scenario 👍🏻

mtaylorgds:
> This looks good to me @mtaylorgds 👍🏻
 
Thanks Ollie—would you mind merging it for me when you get a chance please?

PR #2134, opened 2023-08-22
Add confirm discard page as an extra step when discarding a draft (un…

cavgovdig:
…published) manual

## What

When discarding a draft (unpublished) manual, replace the existing confirmation system alert popup with a separate confirmation page.

## Why

To improve user experience and move towards the design system.

## Visuals

Clicking the Discard draft manual button

image

Would raise this popup

image

That the user would then click OK on to confirm discard.

With the changes in this PR, it will instead take the user to a confirmation page:

image

Where clicking Discard draft will actually perform the discard operation (as clicking OK on the previous alert).

Once discarded, the success message, which was appearing as an info message:

image

Will now appear styled as a success message:

image

[Trello card](https://trello.com/c/1EYhkzIj)

### DIFF
--- a/features/deleting-a-manual.feature
+++ b/features/deleting-a-manual.feature
@@ -16,6 +16,7 @@ Feature: Rake task to delete a manual
     Given a draft manual exists without any sections
     And a draft section exists for the manual
     When I discard the draft manual
+    And I confirm draft deletion
     Then the manual and its sections are deleted
 
   Scenario: Deleting a published manual
@@ -23,3 +24,10 @@ Feature: Rake task to delete a manual
     When I run the deletion script
     Then the script raises an error
     And the manual and its sections still exist
+
+  Scenario: UI safeguarding against deleting a published manual
+    Given a published manual exists
+    And I am on the show page for that manual
+    Then I should not see the discard draft button
+    And I visit the confirm discard page directly
+    Then I am on the show page for that manual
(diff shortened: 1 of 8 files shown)


features/deleting-a-manual.feature
```diff
@@ -16,10 +16,18 @@ Feature: Rake task to delete a manual
...
     When I run the deletion script
     Then the script raises an error
     And the manual and its sections still exist
+
+  Scenario: UI safeguarding against deleting a published manual
+    Given a published manual exists
+    And I am on the show page for that manual
+    Then I should not see the discard draft button
+    And I visit the confirm discard page directly
+    Then I am on the show page for that manual
```
ryanb-gds, line 33: Feels like this should perhaps be two different scenarios? One testing the button visibility and another the direct access attempt? I might even go further and question whether a lower-level test might be appropriate here as feature tests are quite expensive (assuming they are running via a browser), so I tend to prefer them for happy path only tests. But that may not be practical here and obviously that's a very subjective point

mtaylorgds: I tend to agree with everything you've said, there just aren't much in the way of controller/view specs in the manuals publisher codebase. However, we have taken a stab at introducing some, and removed the cucumber test. Let us know your thoughts! 😄

ryanb-gds (approved):
Thanks for reviewing the tests, they look nicely balanced now to me

PR #2144, opened 2023-08-29
Move manual show page to design system

mtaylorgds:
## What

Migrate the manual 'show' view to the design system layout.

## Why

The entire Manuals Publisher app is being moved over to use the more modern design system and associated components. This change should also bring it more in line with Whitehall, reducing the usability friction for people who need to work across both apps.

### Before

manuals-publisher-show-view-legacy

### After

manuals-publisher-show-view-design-system

## Trello

[Trello card](https://trello.com/c/M0crhdKs)


app/helpers/application_helper.rb
```diff
@@ -15,6 +15,121 @@ def state(manual)
...
+    items
+  end
+
+  def manual_section_rows(manual)
+    manual.sections.map do |section|
+      row = {}
+
+      row[:key] = if section.state == "draft"
+                    tag.span("DRAFT", class: "govuk-tag govuk-tag--s govuk-tag--blue") <<
+                      tag.span(section.title, class: "govuk-!-static-margin-2")
```
ryanb-gds, line 120: This is a little bit hard to read, I would argue

app/views/layouts/design_system.html.erb
```diff
@@ -15,6 +15,32 @@
...
+              message: msg,
+            } %>
+          <% else %>
+            <%= msg %>
+          <% end %>
+        <% end %>
+      </div>
+    </div>
+  <% end %>
+
```
ryanb-gds, line 43: Would it make any sense to move this out in to a separate commit? It doesn't seem directly related to the show page

ryanb-gds (commented):
A bit hard for me to review really given I haven't done a lot of frontend work - I'm not really sure how the publishing components should be used. In terms of breaking it up a bit I can see why it's tricky, it's all quite coupled together. The only way I can see to do it is a two-step migration where you create new erb files and scss files for some of the page components/helpers (possibly in a commit per component), then change the manual page itself and remove the old components. That might be two separate PRs, depending on how you want to organise it, or just several commits (but it should be possible to keep tests green in each of them). That's a lot of extra effort to go to, but I suspect other reviewers might find that more digestible.

app/assets/stylesheets/views/manuals.scss
```diff
@@ -0,0 +1,40 @@
...
+.app-view-manuals__section:first-child {
+  padding-top: 0;
+}
+
+.app-view-manuals__sidebar-actions {
+  padding: govuk-spacing(3);
+  margin-bottom: govuk-spacing(6);
+  background-color: govuk-colour("light-grey");
+
+  .gem-c-button {
```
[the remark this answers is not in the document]
mtaylorgds: Not sure, but it doesn't look great without doing this (it's also just the spacing around the component, rather than changing its styling per se).

app/views/layouts/design_system.html.erb
mtaylorgds: It's tricky, because the show page is the first existing page moving over to the design system. The show page uses flash messages, so it is what's driving the need for this being added to the layout template.

ryanb-gds: If that's what rubocop wants, that's good enough for me 🙈

mtaylorgds:
Closing as have noticed an issue with the link checker on legacy forms.

PR #2151, opened 2023-09-06
Consolidation move manual show page to design system

mtaylorgds:
## What

Migrate the manual 'show' view to the design system layout.

This PR supersedes #2144.

## Why

The entire Manuals Publisher app is being moved over to use the more modern design system and associated components. This change should also bring it more in line with Whitehall, reducing the usability friction for people who need to work across both apps.

### Before

manuals-publisher-show-view-legacy

### After

manuals-publisher-show-view-design-system

## Trello

[Trello card](https://trello.com/c/M0crhdKs)


app/views/shared/_summary_card_component.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  rows ||=[]
+%>
+<%= tag.div class: "app-c-summary-card-component", id: id, data: data_attributes do %>
+  <div class="govuk-summary-card">
+    <div class="govuk-summary-card__title-wrapper">
+      <h2 class="govuk-summary-card__title"><%= title %></h2>
+      <ul class="govuk-summary-card__actions">
+        <% summary_card_actions.each do |action| %>
+          <li class="govuk-summary-card__action">
+            <%= link_to sanitize(action[:label] + tag.span(" #{title}", class: "govuk-visually-hidden")), action[:href], class: "govuk-link govuk-link--no-visited-state #{"gem-link--destructive govuk-!-font-weight-bold" if action[:destructive]}".strip %>
```
davidtrussler, line 14: Very small point: there's no need to add a space in the visually hidden text here (and elsewhere) since the `govuk-visually-hidden` CSS class does that. Not a big deal though. 

I do wonder how useful the hidden text is though for a unique action like "Edit first publication date". If we consider it from the perspective of a screen reader user it might even add confusion as the screen reader would announce "Edit first publication date about this manual". It might be worth considering removing the hidden text altogether depending on the contexts in which the component is used. 

This would be even more the case for the "Reorder sections" action, which would read as "Reorder sections sections". In that case, and if we keep the hidden text, could we just display "Reorder" and "Add"?

app/views/manuals/show.html.erb
```diff
@@ -1,117 +1,135 @@
...
-        <% if manual.originally_published_at.present? %>
-          <dl class="metadata-list">
-            <dt>Originally published</dt>
-            <dd><%= nice_time_format(manual.originally_published_at) %><% if manual.use_originally_published_at_for_public_timestamp? %><br/>This will be used as the public updated at timestamp on GOV.UK.<% end %></dd>
-          </dl>
-        <% end %>
-      </div>
+    <% if show_preview?(manual) %>
+      <section class="app-view-manuals__section">
+        <h2 class="govuk-heading-l">Preview</h2>
```
davidtrussler, line 54: As above, could we use the heading component here?

app/views/shared/_summary_card_component.html.erb
mtaylorgds: Fair points. This component was copied from Whitehall publisher, and I think I'd prefer to make the simpler change of updating the label text than having to add support for conditionally adding the visually hidden label.

I've spoken to our product owner and XD, and they're happy to update the section actions to "Reorder" and "Add", so we'll implement that change.

mtaylorgds:
@davidtrussler Thanks for your review, I've addressed your comments if you'd be so kind as to re-review 😄

app/views/shared/_summary_card_component.html.erb
davidtrussler: Yes, I was getting a bit pedantic here - sorry! There's a couple of issues that have come up on Whitehall in the same sort of area so we may make some updates on that some time. But that's all good for me! 👍

davidtrussler (approved):
Looks good! 👍

PR #2161, opened 2023-09-12
Add text indicating when edited original publication date will

cavgovdig:
## What
Add text indicating when edited original publication date will be used as public timestamp when manual is published.

## Why
Original functionality on old manual show page, accidentally omitted during migration to new design system.
safe_join() used to preserve html <br> tag instead of interpreting as string literal and rubocop doesn't permit concatenating with html tags.

## Visual
image

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -214,10 +214,16 @@ describe ApplicationHelper, type: :helper do
       expect(rows).not_to include(include(key: "Originally published"))
     end
 
-    it "returns a 'last published' row when the manual has publish tasks" do
-      manual = FactoryBot.build_stubbed(:manual)
-      allow_any_instance_of(ApplicationHelper).to receive(:publication_task_state).and_return("test publication task state")
-      manual.publish_tasks = [{ updated_at: Time.zone.now, state: "queued" }]
+    it "returns a 'last published' row with the most recent publication date when the manual has been published" do
+      manual = FactoryBot.build_stubbed(:manual, use_originally_published_at_for_public_timestamp: false)
+      most_recent_publish_task = { updated_at: Time.zone.now, state: "queued" }
+      manual.publish_tasks = [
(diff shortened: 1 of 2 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -227,14 +227,27 @@ def current_user_can_publish? = false
...
       manual.publish_tasks = []
 
       rows = manual_metadata_rows(manual)
 
       expect(rows).not_to include(include(key: "Last published"))
     end
+
+    it "indicates when originally published time will be used as public timestamp" do
+      manual = FactoryBot.build_stubbed(:manual, use_originally_published_at_for_public_timestamp: true)
+      allow_any_instance_of(ApplicationHelper).to receive(:publication_task_state).and_return("test publication task state")
```
ryanb-gds, line 241: If we're going to mock this (would it be possible not to using timecop or something?) do we want to check that it received the first item from `publish_tasks` as a parameter when `publication_task_state` was called?

ryanb-gds (commented):
Just want to check we're writing the best test we can be, under the circumstances. Appreciate it's a bit of an awkward edge case to handle, by the look of it

spec/helpers/application_helper_spec.rb
ryanb-gds: Oh I was just wondering if you'd mocked the `publication_task_state` helper method because it was time sensitive, in which case we could use the real function and use Timecop to freeze time so we could make a predictable assertion without mocking the method. But that's probably not the case. If that doesn't clarify things, ping me and we can jump on a call. It's not particularly important anyway

ryanb-gds (approved):
LGTM

spec/helpers/application_helper_spec.rb
mtaylorgds: Ah, no, it's more to avoid having to worry about the logic of what that method returns, to keep this test more "unity".

PR #2163, opened 2023-09-14
Migrate the edit first publication date page to the design system

cavgovdig:
## What
Update the page for editing a manual's first publication date to use the
 `design_system` layout, and be built using UI components from the
 design system.

The `datetime_fields` component is copied from Whitehall publisher
 (where it's called `_datetime_fields_with_govuk_date_component` to
 differentiate it there from a similar component that doesn't use the
 govuk `date_input` component).

## Why
Moving to the new design system to ensure a consistent user experience.

### Before 
image

### After
image

## Trello card

[Trello](https://trello.com/c/MmNstYaT)


ryanb-gds (commented):
Uh oh, my component has escaped 😱 

Just out of interest, have you tested putting an invalid date in here, e.g. 32/01/2023? My suspicion is it will blow up. I had to make a [little module](https://github.com/alphagov/whitehall/blob/main/app/models/date_validation.rb) to stop that happening on Whitehall

ryanb-gds (approved):
Approving this on the basis that it's not making things any worse, as it's already possible to cause a 500 error by leaving part of the date out

PR #2166, opened 2023-09-15
Migrate add section page to design system

mtaylorgds:
## What

Migrate the add section view to the GDS Design System.

The govspeak help for sections also has an extra section for "collapsible" sections, which only applies to sections and not manuals, so the `govspeak_help` partial has been updated to conditionally show this extra help.

## Why

This is part of the migration of all of the Manuals Publisher pages to the GDS Design System.

### Before

manuals-publisher dev gov uk_manuals_9c82538e-58d6-4934-b392-632217fb649a_sections_new (1)

### After

manuals-publisher dev gov uk_manuals_b544a5c1-a844-41d2-a89c-5cba98b1658a_sections_new

## Trello

[Trello card](https://trello.com/c/wujJastP).


app/views/sections/_form.html.erb
```diff
@@ -1,79 +1,116 @@
-<div class='row'>
-  <div class="col-md-8">
+<% content_for :head do %>
+  <%= javascript_include_tag "vendor/jquery-1.11.0.min" %>
+  <%= javascript_include_tag "ajax_setup" %>
+  <%= javascript_include_tag "markdown_preview" %>
```
ollietreend, line 4: What do these JavaScript files do? Are they needed in the Design System layout, or should they be in `_legacy_form.html.erb`?

While porting pages across to the Design System, we've generally tried to avoid including jQuery and legacy JavaScript. Instead, we've used govuk_publishing_components that implement the desired behaviour, or have used vanilla JS for implementation.

cavgovdig: at the moment we're still relying on the old govspeak preview, which requires these js tags. to upgrade to the new govspeak needs implementing a markdown preview endpoint, unless theres another way?

mtaylorgds: The jquery stuff has been removed from main, and this PR's changes rebased on top of that, so it no longer has any dependency on jquery on design system pages.

ollietreend (approved):

PR #2176, opened 2023-09-25
Migrate pre publish messages to design system

cavgovdig:
## What
There is a set of 'pre publish messages' dictated by the status of a manual and its sections. These appear within the sidebar containing the publish and/or discard draft buttons on the manuals show page. These have been readded to the page and migrated to the design system font. 

## Why
This is part of the migration of all of the Manuals Publisher pages to the GDS Design System.

### Before
image

### After
image

## Trello
[Trello card](https://trello.com/c/xvOR2mXC)

### DIFF
--- a/app/views/manuals/show.html.erb
+++ b/app/views/manuals/show.html.erb
@@ -101,6 +101,9 @@
   <div class="govuk-grid-column-one-third">
     <% if allow_publish?(manual, slug_unique) || !manual.has_ever_been_published? %>
       <div class="app-view-manuals__sidebar-actions">
+        <% publish_text(manual, slug_unique).each do |message| %>
+          <p class="govuk-body"> <%= message.html_safe %> </p>
+        <% end %>
         <ul class="govuk-list govuk-list--spaced">
           <% if allow_publish?(manual, slug_unique) %>
             <li>
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -378,4 +378,69 @@ describe ApplicationHelper, type: :helper do
       expect(rows).to include(include(key: "<span>test title</span>"))
     end
   end
+
+  describe "#publish_text" do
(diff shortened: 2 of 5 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -378,4 +378,58 @@ def current_user_can_publish? = false
       expect(rows).to include(include(key: "<span>test title</span>"))
     end
   end
+
+  describe "#publish_text" do
+    it "returns correct text if the manual is published" do
+      manual = FactoryBot.build_stubbed(:manual)
+      allow(manual).to receive(:state).and_return("published")
```
ryanb-gds, line 385: Could we override the attributes on the factory to set the state value, rather than mocking it? Or would that depend on the database?

app/views/manuals/show.html.erb
```diff
@@ -99,8 +99,11 @@
   </div>
 
   <div class="govuk-grid-column-one-third">
-    <% if allow_publish?(manual, slug_unique) || !manual.has_ever_been_published? %>
+    <% if allow_publish?(manual, slug_unique) || !manual.has_ever_been_published? || publish_text(manual, slug_unique) != "" %>
```
ryanb-gds, line 102: A minor gripe, but feels like this conditional is getting a bit lengthy, especially as we then check most of the conditions again inside it. I wonder if there's a way to model the conditions a little more effectively? We have [a component](https://github.com/alphagov/whitehall/blob/main/app/components/admin/editions/show/sidebar_actions_component.rb) for this in Whitehall, I know Manuals doesn't use components but maybe building the markup in the helper rather than the template would feel cleaner?

ryanb-gds (commented):
I think we can maybe look at the factory usage, unless there's a dependency on the database I'm not aware of?

ryanb-gds (approved):
Looks good. Handy that we were able to remove that check on whether any publish_text was present in the template

PR #2179, opened 2023-09-28
Replace the govspeak editing code

cavgovdig:
## What
Replace the `markdown_preview` script and associated gubbins with the `govspeak-editor` component from Whitehall (copied across and updated to work with Manuals Publisher).

To enable the govspeak editor component to work for both manuals and sections, the format of the data submitted has been updated to be non-specific (i.e. it now sends the markdown in a "body" field, rather than the previous "manual[body]"/"section[body]" fields), which requires a slight change to the controller. Since only edit manual has been moved to the design system, we still need a controller endpoint for the new manual preview to work, which is now called "legacy_preview" and will be removed once we migrate the "new manual" page to the design system.

## Why
This enables `jquery` to be removed as a dependency for pages that are using the design system layout template (and so to be removed completely once the migration to the GDS Design System is finished).

### Before
image

### After 
image

## Trello
[Trello card](https://trello.com/c/wujJastP)


ollietreend (approved):
Nice work! 🎉

PR #2201, opened 2023-10-10
Fix bug in which the edit manual page allowed saving a manual with an empty summary or title in the database

cavgovdig:
## What
Fixing a pre existing bug:
Previous behaviour: edit manual -> delete title / summary -> click 'save' -> manual database record updated -> errors raised -> edit manual page re-rendered with errors 
If this page is exited out of (using Cancel link), the manual record in the database will have empty title / summary. This can then cause knock on effects with breadcrumbs and other items rendered.

New behaviour: edit manual -> delete title / summary -> click 'save' -> manual validity checked, manual not saved -> errors raised -> edit manual page re-rendered with errors 
Manual with empty title / summary will not be saved to database.

## Why
Prevents invalid data being saved.

## Trello
[Trello card](https://trello.com/c/hDkz7XzD)

### DIFF
--- a/app/services/manual/update_service.rb
+++ b/app/services/manual/update_service.rb
@@ -12,9 +12,12 @@ class Manual::UpdateService
 
     manual.draft
     manual.assign_attributes(attributes)
-    manual.save!(user)
-    reloaded_manual = Manual.find(manual.id, user)
-    Adapters.publishing.save_draft(reloaded_manual)
+
+    if manual.valid?
+      manual.save!(user)
+      reloaded_manual = Manual.find(manual.id, user)
+      Adapters.publishing.save_draft(reloaded_manual)
+    end
 
     manual
   end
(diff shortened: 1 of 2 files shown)


app/services/manual/update_service.rb
```diff
@@ -12,9 +12,12 @@ def call
 
     manual.draft
     manual.assign_attributes(attributes)
-    manual.save!(user)
-    reloaded_manual = Manual.find(manual.id, user)
-    Adapters.publishing.save_draft(reloaded_manual)
+
+    if manual.valid?
```
ryanb-gds, line 16: This shouldn't really be necessary as `save!` should raise an exception if the model is not valid 🤔

ryanb-gds (commented):
Something doesn't seem quite right here

app/services/manual/update_service.rb
ryanb-gds: Having looked into this a bit further, we've agreed to stick with the current PR implementation, as this is how the `CreateService` solves the issue too. The `save!` method is not throwing an exception because it has been overridden. We don't really like this because the overridden method changes the contract provided by the default ActiveRecord `send!` method, but such is life without checked exceptions. If we were to change this it would need to be a separate PR that considered the whole application

ryanb-gds (approved):

PR #2205, opened 2023-10-16
Set default timezone to London in config

davidgisbey:
## Description

 We've had a bug come in that a that was published at 12.01am is saying that it was published the previous day on the updates section.

Looking at the config I can see that a default timezone hasn't been set. This means it's using UTC. This doesn't account for BST being an hour ahead during the summer time so when the updated_at is passed to Publishing API it's passing a time of an hour earlier than it was published.

We can resolve this by setting the default timezone to London.

As this is a pre-existing bug that's basically always existed, we run into a bit of an issue with data integrity of previous manuals. We can simply republish the most recent published edition of the manual, but there could be quite a large amount of editions which have the incorrect date that won't be updated.

It might be that tackling that in Publishing API itself is the right way to go.

## Trello card

https://trello.com/c/idVD4RQT/1794-update-history-shows-the-wrong-date

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -128,8 +128,7 @@ module ApplicationHelper
   end
 
   def publication_task_state(task)
-    zoned_time = time_with_local_zone(task.updated_at)
-    formatted_time = nice_time_format(zoned_time)
+    formatted_time = nice_time_format(task.updated_at)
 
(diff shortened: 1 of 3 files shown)


davidgisbey:
closing in place of https://github.com/alphagov/government-frontend/pull/2959

PR #2206, opened 2023-10-18
Do not publish a log for a removed section which was never published

ryanb-gds:
Previously a change log entry would be created if a section was withdrawn before it was ever published. This makes sure that no publication log entries are created if the section had never been published before.

I also tweaked the test setup a little for the publication logger test to rely less on shared test state and added a missing test for the removed section which has previously been published.

Trello: https://trello.com/c/sjGHrnOw


beccapearce (approved):
Looks good! 
And tests look great 🤩

PR #2217, opened 2023-10-30
Move to supported state machine gem

baisa:
[Trello card](https://trello.com/c/sUvy7JMe/1852-update-statemachine-to-statemachines-gem-for-manuals-publisher)


app/models/manual_publish_task.rb
```diff
@@ -1,4 +1,4 @@
-require "state_machine"
+require "state_machines"
```
ollietreend, line 1: Do we definitely need this `require` line?

The `SectionEdition` model seems to work okay without the explicit require.

config/initializers/state_machine_patch.rb
```diff
@@ -1,7 +1,7 @@
...
-    end
-  end
-end
+# module StateMachine
+#   module Integrations
+#     module ActiveModel
+#       # public :around_validation
+#     end
+#   end
+# end
```
ollietreend, line 7: Looks like this file should be safe to delete altogether 🙌🏻

ollietreend (commented):
Nice work @baisa and @beccapearce 🤩

dnkrj (approved):

PR #2263, opened 2023-12-20
Stop 404 when withdrawing a section on published manual

davidgisbey:
## Description 

When a manual is published all it's sections are published as well. Published sections retain a published state except in 2 circumstances.

1. they are edited at which point they come drafts
2. they are withdrawn at which point they're removed from the manual in the Manuals Publisher DB

If they are edited a draft section is pushed downstream to Publishing API and the Draft Content Store.

When this occurs and the section is withdrawn, we need to discard the draft Section from Publishing API.

There's some logic to handle this in the remove_service which checks to see if the section is a draft (i.e. has a draft edition in Publishing API) and if so it attempts to discard the draft.

However, there's an issue with this. Within the remove service `assign_attributes` is called on the section. If you go and have a look at t[he method defined in the Section model](https://github.com/alphagov/manuals-publisher/blob/f93fb57d54524ee1b973f242eaa1098ecd13d95f/app/models/section.rb#L81-L105), you can see that it actually creates a new draft
section when called on a published section. It then passes change note params to the new draft edition of the section.

This means that when `#draft?` is called on the section it will always return true. This causes an issue when the following is called:

    ```
    if section.draft?
      Adapters.publishing.discard_section(section)
    end
    ```

Publiishing API blows up as there is no draft section in the Publishing API database.

There are feature specs which explicitly cover that the section is removed when Published sections are withdrawn. The proposed fix of capturing the state prior so `assign_attributes` being called breaks these tests.

I'm pretty sure that the tests are actually testing the incorrect behaviour which is causing the bug though.

At first glance I thought it might be possible to not bother creating a new draft edition prior to removing the section from the manual, but it's actually required for capturing the change note the user inputs so that users can see why the section was removed.

When an edition is published downstream, any change notes are pulled from removed sections and added as PublicationLogs

These are then rendered on the frontend.

## Trello card

https://trello.com/c/GMy5HBPv/1830-fix-manual-section-withdrawal-error-caused-by-missing-draft-in-publishing-api

### DIFF
--- a/app/services/section/remove_service.rb
+++ b/app/services/section/remove_service.rb
@@ -20,6 +20,14 @@ class Section::RemoveService
       minor_update: attributes.fetch(:minor_update, "0"),
       change_note: attributes.fetch(:change_note, ""),
     }
+
+    # We need to capture the state of the section before assigning attributes.
+    # The Section#assign_attributes method always creates a new draft section if
+    # the latest edition is published.
+    # This causes Adapters.publishing.discard_section(section) to be called which
+    # blows up as there is no draft section in the Publishing API database.
+    draft_section = section.draft?
+
     section.assign_attributes(change_note_params)
 
     if section.valid?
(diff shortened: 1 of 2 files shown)


ryanb-gds (approved):
Good short term fix, thanks!

PR #2271, opened 2024-01-04
Fix Rails secrets deprecation warning

ryanb-gds:
Rails 7.1 deprecates the use of Rails.application.secrets in favour of Rails.application.credentials.

The credentials system introduces the burden of master encryption key administration at no benefit to us, because we manage our production secrets using environment variables instead of committing them to Git.

This commit loads the existing secret values and merges them into the credentials object. This approach was copied from commit 8937b172be530a5d91bd999f4538e5a722dcab19 on the GOV.UK account-api project.

### DIFF
--- a/app/controllers/link_checker_api_callback_controller.rb
+++ b/app/controllers/link_checker_api_callback_controller.rb
@@ -39,6 +39,6 @@ private
   end
 
   def webhook_secret_token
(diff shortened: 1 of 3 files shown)


davidgisbey (approved):
lgtm!

PR #2290, opened 2024-01-22
Include Publishing component stylesheets in build

ryanb-gds:
Without adding the Publishing component stylesheets in the build, the component guide stylesheet was not included, leading to a broken page at "/component-guide"


davidgisbey (approved):
lgtm!

PR #2355, opened 2024-04-05
Allow unpublsihing when a manual state is withdrawn

lauraghiorghisor-tw:
We make sure the publishing-api call succeeds before changing the state of the manual. We allow the unpublish call to execute when the manual is either published or (already) withdrawn.

We have experienced that we are unable to rerun the `withdraw_and_redirect_manual` rake task, should the unpublish api call fail. These changes ensure we can rerun.

[Trello card](https://trello.com/c/mrQPt8pw/2371-manuals-publisher-unpublish-17-manuals)

### DIFF
--- a/spec/lib/withdraw_and_redirect_manual_spec.rb
+++ b/spec/lib/withdraw_and_redirect_manual_spec.rb
@@ -33,27 +33,40 @@ RSpec.describe WithdrawAndRedirectManual do
     allow(publishing_adapter).to receive(:unpublish_and_redirect_manual_and_sections)
   end
 
-  it "withdraws the manual" do
+  it "withdraws the manual and unpublishes" do
     subject.execute
 
     reloaded_manual = Manual.find(manual.id, user)
     expect(reloaded_manual.withdrawn?).to eq(true)
-  end
-
-  it "calls the publishing adapter to unpublish the manual" do
-    subject.execute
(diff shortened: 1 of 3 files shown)


spec/lib/withdraw_and_redirect_manual_spec.rb
```diff
@@ -33,30 +33,47 @@
...
+
+    before do
+      manual.withdraw
+      manual.save!(user)
+    end
+
+    it "withdraws and unpublishes the manual again" do
+      subject.execute
+
+      reloaded_manual = Manual.find(manual.id, user)
```
ryanb-gds, line 67: The setup for this test feels a little bit awkward - could we create a separate context for each scenario (manual publisher/manual already withdrawn) rather than overriding the default context? It would make the tests easier to understand, I feel

ryanb-gds (approved):
Couple of suggestions, but ultimately would merge as is

spec/lib/withdraw_and_redirect_manual_spec.rb
lauraghiorghisor-tw: Kept the state let at the top as it feels somewhat "default" that it's set to published. The last test on dry-run also expects a default state. Removed the before so it reads better, and the let for our test as we were using the default anyway.

PR #2390, opened 2024-05-23
Introduce editing change history for gds-editors

lauraghiorghisor-tw:
Introducing a UI feature to allow users to delete change notes. These requests typically came through Zendesk and required manual intervention. This change would enable users to now do it themselves. 

The implementation follows the way `Whitehall` implements change notes, but adapted to the modelling in `manuals-publisher`, which had its own `PublicationLogs` implementation. 

The feature can only be used by `gds-editors` as it is a destructive action.

[Trello](https://trello.com/c/0o6xP2vR/2282-remove-change-notes-from-page-manuals-publisher)

Link showing in the side bar
Screenshot 2024-06-04 at 19 21 41

Change notes index page
Screenshot 2024-06-04 at 19 21 49

Confirmation page
Screenshot 2024-06-04 at 19 21 57

### DIFF
--- a/app/controllers/change_history_controller.rb
+++ b/app/controllers/change_history_controller.rb
@@ -1,0 +1,7 @@
+class ChangeHistoryController < ApplicationController
+  def index
+    manual_record = ManualRecord.find_by(manual_id: params[:manual_id])
+    @manual = Manual.build_manual_for(manual_record)
+    @publication_logs = @manual.publication_logs.reverse
+  end
+end
--- a/app/controllers/change_history_controller.rb
+++ b/app/controllers/change_history_controller.rb
@@ -6,6 +6,10 @@ class ChangeHistoryController < ApplicationController
     @publication_logs = @manual.publication_logs.reverse
   end
 
+  def confirm_destroy
+    @publication_log = PublicationLog.find(params[:id])
+  end
+
 private
 
   def find_manual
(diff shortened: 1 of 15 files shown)


app/controllers/change_history_controller.rb
```diff
@@ -0,0 +1,41 @@
+class ChangeHistoryController < ApplicationController
```
lauraghiorghisor-tw, line 1: Just to point out the naming unrailsiness, in case you've had additional thoughts since our chat.

app/controllers/change_history_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  before_action :find_manual
+  before_action :authorize_user_for_changing_history
+
+  def index
+    @publication_logs = @manual.publication_logs
```
lauraghiorghisor-tw, line 6: These are ordered by date, meaning the latest change notes do show at the bottom (matches Whitehall behaviour). I kind of disliked it first, but then it does make sense. Perhaps ordering in descending order would match the presentation on the FE better (which is latest at the top) and be better for the user?

lauraghiorghisor-tw:
There is some off flash behaviour,  in that the success or error messages show on top of the breadcrumbs. But that seems to pre-date my changes

app/views/manuals/show.html.erb
```diff
@@ -117,9 +117,11 @@
     <% end %>
 
     <% if manual.published? %>
```
lauraghiorghisor-tw, line 119: @ryanb-gds We discussed this at some point, I agree that there is no point to show the link if there is nothing published at all.

PR #2426, opened 2024-07-10
Remove orphaned sections

minhngocd:
There are data integrity issues we need to address. In the past, some manuals were withdrawn but the children sections were left published. More information on the Trello card below. 

Out of all these orphaned children sections that are still published when their parent manual has been withdrawn, some are still live, whereas some have been unpublished and redirected on Publishing API (but still has state published on the Manual Publisher)

For sections that are still live, we need to withdraw and redirect them. 
For sections that are already unpublished and redirected on Publishing API, we only need to update the state of the SectionEdition to `archived`.

These tasks should be **temporary** and should only need to run this once in production, and PR can be reverted afterwards


spec/lib/tasks/withdraw_and_redirect_section_rake_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    it "withdraws sections that exist and output an error line for ones that don't" do
+      FactoryBot.create(:manual_record, slug: "guidance/parent_path", state: "withdrawn")
+      FactoryBot.create(:section_edition, section_uuid: "1234", slug: "guidance/parent_path/existing_doc", state: "published")
+
+      expect { task.invoke("./spec/fixtures/bulk_test.csv") }.to output(/unable to withdraw guidance\/parent_path\/non_existing_doc/).to_stdout
+
+      assert_publishing_api_unpublish(
+        "1234",
+        type: "redirect",
+        alternative_path: "guidance/parent_path",
```
ChrisBAshton, line 27: Nitpick: doesn't Publishing API always have the `/` prefix? I.e. we should be passing `alternative_path: "/guidance/parent_path",`?

ChrisBAshton (approved):
Well done for getting to the bottom of this - I know a lot of prep work went into it! ⭐

spec/lib/tasks/withdraw_and_redirect_section_rake_spec.rb
minhngocd: updated to reflect this now 🙏

PR #2427, opened 2024-07-11
Add rake task for reslugging organisations

ChrisBAshton:
This org was recently renamed:
`ministry-of-housing-communities-and-local-government` =>
`ministry-of-housing-communities-and-local-government-2018-2021`

We want to update the slugs of relevant manuals as per
https://docs.publishing.service.gov.uk/manual/changing-organisation-slug.html#2-update-the-organisation-slug-in-manuals-publisher

However, the docs refer to creating a data migration, which is no longer supported in Manuals Publisher. This PR instead adds a rake task (and removes the stray data migration files). See commits for details.

I'll update the docs post-merge.

Trello: https://trello.com/c/G6OS639d/2731-reslug-ministry-of-housing-communities-local-government

---


ryanb-gds (approved):
I think you're right about not needing to republish, I think this is just to make sure users from the right organisation have access to the manuals

ryanb-gds (approved):
Probably good to have this as a rake task anyway as it won't be the last time

PR #2433, opened 2024-07-15
Fixing withdraw section rake task so it sends discard draft correctly…

minhngocd:
… to Publishing API

Previously, the rake task for removing sections had a conditional where it only sent discard draft flag to Publishing API only if the latest section edition was a draft and the rake task had the discard draft flag set to true.

However, when the rake task was run to remove orphaned sections, we seem to have sections where the latest edition is published (not draft) but somehow contain drafts on Publishing API - we were unable to remove them due to an error "Unable to unpublish edition with draft". Amending the publishing API call for this Withdraw Section task to always send discard_draft flag straight through rather than check for draft so we can unpublish these sections.

### DIFF
--- a/spec/lib/tasks/withdraw_and_redirect_section_rake_spec.rb
+++ b/spec/lib/tasks/withdraw_and_redirect_section_rake_spec.rb
@@ -25,6 +25,7 @@ describe "withdraw and redirect section rake tasks", type: :rake_task do
         "1234",
         type: "redirect",
         alternative_path: "/guidance/parent_path",
+        discard_drafts: true,
       )
     end
   end
--- a/app/adapters/organisations_adapter.rb
+++ b/app/adapters/organisations_adapter.rb
@@ -1,9 +1,6 @@
 class OrganisationsAdapter
-  def initialize
-    @cache = {}
-  end
-
-  def find(slug)
+  def self.find(slug)
+    @cache ||= {}
     @cache.fetch(slug) do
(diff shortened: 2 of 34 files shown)


spec/lib/tasks/withdraw_and_redirect_section_rake_spec.rb
```diff
@@ -25,6 +25,7 @@
         "1234",
         type: "redirect",
         alternative_path: "/guidance/parent_path",
+        discard_drafts: true,
```
ryanb-gds, line 28: It doesn't feel quite right that we can just change a line in the assertion here and not change any inputs or have to create a new test. Probably reflects more on there being a missing test case for the other side of the condition before though

ryanb-gds (commented):
The original behaviour is sort of correct if you assume that Manuals and Publishing API are in sync, which is an invalid assumption but does make me wonder if this will have any unanticipated side effects.

There's not some scenario where we're trying to republish the manual and we can't because Manuals thinks theres a draft section in place but Publishing API doesn't, is there?

minhngocd:
> Meant to approve this, as long as you're happy with both the comments I left before

I'm having a think about this. Will add the missing test for the withdraw service. 
As well as your question about whether we'll have the drafts be in different state. I'll add in another clause to delete any drafts on manuals publisher before sending over the unpublishing so that we make sure that drafts are blown away on both sides.

PR #2435, opened 2024-07-16
Manuals publisher refactoring

minhngocd:
Recommend review commit by commit

Refactoring Manuals Publisher as there are a number of bugs happening when withdrawing sections. 
It's probably best to do some refactoring so we are able to reason with and make things a little more consistent before we make any changes to fix the bugs. 

Two refactorings addressed in this PR:
- Removal of Adapters class. This class is a wrapper class which creates an instance of PublishingAdapter and OganisationsAdapter, and allows access to the instance globally. These two instances have been memo-ized since 2017 and don't really have any instance specific state. It should be safe to change them all to global static method calls. This removes the need to jump through 2 layers to get to an adapter, and reduces the need to do multi-level mocking.
- Inlining of `organisations_for` method on the PublishingAdapter. The method is a single line method that essentially calls out to the OrganisationsAdapter. If we inline this, we reduce the methods' coupling to a single method within PublishingAdapter and would allow us to start extracting methods to their own classes to further break down PublishingAdapter. 

As a side effect of the above refactoring, we had to restructure a number of test classes, the most significant of which was Section Removal Service spec which tested some quite complex scenarios, but based it all off of stubs for actual ActiveRecord models (sometimes the stubs were not stubbing realistic data. e.g. stub manual.valid? to return false, rather than setting up data to also sync in line with this). Ended up re-writing the entire class to use real data models for a realistic and hopefully more readable test. Added missing test cases where I can.

Some error handling were also not actually handling the errors when tests were written, so have also removed redundant error handling.

Further todo in next PR or planned refactoring:
We should look at breaking down PublishingAdapters into smaller services as they have some distinct flows that don't touch each other. 
We should also look at merging PublishingAdapter functionality back into Manual/Section Services. The original intent of extracting this as separate `Adapters` classes is documented in README as encapsulating calls to GDS API Adapters, but this is no longer the case, as they are also making data changes to ActiveRecord models (e.g. Section withdrawal). As the calls to Publishing API are already encapsulated within GDS API Adapters, maybe there is no longer a need for separate Adapters.

### DIFF
--- a/spec/services/manual/withdraw_service_spec.rb
+++ b/spec/services/manual/withdraw_service_spec.rb
@@ -1,0 +1,63 @@
+require "spec_helper"
+
+RSpec.describe Manual::WithdrawService do
+  let(:user) { User.gds_editor }
+  let(:state) { "published" }
+  let!(:manual_record) { FactoryBot.create(:manual_record, manual_id: "manual-id", state:) }
+
+  subject do
+    described_class.new(user:, manual_id: manual_record.manual_id)
+  end
+
+  it "raises error when Manual is not found" do
+    expect {
+      Manual::WithdrawService.new(user:, manual_id: "non-existant-id").call
+    }.to raise_error(Manual::NotFoundError, "Manual ID not found: non-existant-id")
+  end
+
+  it "withdraws published manuals" do
+    expect(PublishingAdapter).to receive(:unpublish).with(have_attributes(id: manual_record.manual_id)).once
+    subject.call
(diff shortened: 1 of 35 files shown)


spec/services/manual/withdraw_service_spec.rb
ChrisBAshton, line 1: Great work here 👍

ChrisBAshton (approved):
LGTM 👍

PR #2454, opened 2024-08-06
Allow section withdrawal to force discard draft on publishingn API

minhngocd:
Upon further investigation, there is no syncing between document state in Manuals Publisher and Publishing API. The sections are meant to be left at "published" state even when their manual parents have been withdrawn. The redirects that are breaking in production are introduced because they are redirects set on Router and not Publishing API unpublish/redirects.
This temporary rake task should not make any changes to the state, but just push the redirects.

Additionally added a check for safety when withdrawing sections to make sure we are fetching the correct UUID for a published section. If there are multiple published section editions with differing Section UUIDs, this should be a cause for concern and warrant further investigation. We error out instead of actioning the task.


minhngocd:
There is still a number of outstanding things that should be addressed in Manuals Publisher. But due to growing scope creep, they should probably be considered as separate Trello tickets and not be considered for this. 

These are: 
- Why is there SecureRandom uuid generated for contentID when we redirect Section for Publishing API put content? Is there any impact in sending the section uuid? 
- Why are the adapters not using presenters, but crafting the schema to send to publishing API within the adapter? Should follow Whitehall pattern of crafting the schema within Presenters?
- Refactor for rake tasks to use the services instead of creating their own. At the moment, we have 3 places where unpublishing, publishing, redirect etc. logic exist: Adapters, Services, and Libs. These should be consolidated, especially as some of them have diverged in logic. We need to draw out a flow and merge them where we can.

minhngocd:
After some investigations and discussions - we also deemed the withdraw logic and flow a little bit flawed (possibly due to modelling limitations) as it leaves strange states behind in the database (such as past sections still being in "published" state but that doesn't matter because it is no longer "latest". This _may_ also be the reason why these orphaned live sections are there in the first place, and may cause more in the future.

However, attempt to refactor / correct broke a few things so it's now been put outside the scope of this fix to do (much more effort required). 

For this fix, we just want to be able to discard drafts in Publishing API to force unpublish some live documents.

app/models/section.rb
```diff
@@ -126,6 +126,7 @@ def publish!
 
   def withdraw_and_mark_as_exported!(exported_at = Time.zone.now)
     latest_edition.exported_at = exported_at
+    latest_edition.save!
```
ryanb-gds, line 129: `latest_edition.archive` on the line below calls `save!` under the hood, so I think this is superfluous. Hopefully there aren't any section editions that are already withdrawn that don't have an `exported_at` attribute

ryanb-gds (commented):
I'm not actually sure the changes we've made are solving all of our problems. I've been thinking about where we check we're only trying to withdraw published sections.

A published section edition will have the same UUID as the latest draft. The only reason it wouldn't have the same UUID would be if it was in fact a new section which had been given the same title as an existing section (and therefore the same slug). 

Nothing in Manuals Publisher actually prevents this, but we do warn the user when they're about to publish the manual. However, if the user ignores this warning, there's nothing stopping us from having two section editions that are both published with the same slug and different UUIDs. In that case the `published` filter isn't actually going to help us get the UUID we want, potentially 🙈 

We might have to rewrite the withdraw rake task to accept a UUID, or something. Yuck.

Digression: It's properly wild that if you pass `published: true` to `Section.find` it won't load draft editions, so the latest edition will be published even though there is actually a later edition in the database. 🤯

app/models/section.rb
minhngocd: I changed one of the tests to stop using mocks and actually assert for exported_at and that test was starting to fail without this line added.

minhngocd:
> I'm not actually sure the changes we've made are solving all of our problems. I've been thinking about where we check we're only trying to withdraw published sections.
> 
> A published section edition will have the same UUID as the latest draft. The only reason it wouldn't have the same UUID would be if it was in fact a new section which had been given the same title as an existing section (and therefore the same slug).
> 
> Nothing in Manuals Publisher actually prevents this, but we do warn the user when they're about to publish the manual. However, if the user ignores this warning, there's nothing stopping us from having two section editions that are both published with the same slug and different UUIDs. In that case the `published` filter isn't actually going to help us get the UUID we want, potentially 🙈
> 
> We might have to rewrite the withdraw rake task to accept a UUID, or something. Yuck.
> 
> Digression: It's properly wild that if you pass `published: true` to `Section.find` it won't load draft editions, so the latest edition will be published even though there is actually a later edition in the database. 🤯

The way i've derived the list of sections that require unpublishing, though, is via slug, not UUID... 
- given a list of withdrawn manuals, fetch all sections that have published state. 
- curl through each of those slugs and check whether or not they have a redirect in place on Publishing API. 
- If they don't have a redirect, that means they are live and need to be unpublished
- if they have a redirect in place then we need to make sure the latest edition is really "withdrawn" but don't need to send Publishing API unpublish

minhngocd:
Upon further discussion, we've decided not to make any changes to the state of the section for this problem of having live orphaned sections. There seems to be no sync between manuals publisher and Publishing API, and changing of the state can cause possible side effects. Instead the rake task should just do a redirect.

lib/tasks/withdraw_and_redirect_section.rake
```diff
@@ -1,22 +1,29 @@
...
   section_count = CSV.read(args.fetch(:csv_path), headers: true).length
   puts "Withdrawing and unpublishing #{section_count} sections..."
 
   failures = {}
   CSV.foreach(args.fetch(:csv_path), headers: true) do |row|
     section_path = row["section_path"]
     base_path = section_path.split("/")[0..1].join("/")
-    redirect_section(section_path, "/#{base_path}", true)
+    section_edition = SectionEdition.where(slug: section_path).first
+    Publishing::RedirectAdapter.redirect_section(section_edition, to: "/#{base_path}")
```
ryanb-gds, line 26: Should this be withdrawing the sections after all, now that we know the manual page will show archived sections as long as the manual is in the withdrawn state?

ryanb-gds: Agreed that this will put the sections into the same state as 900ish other sections which have a redirect in place on Publishing API but are not in the `archived` state in the Manuals Publisher database. We can look at fixing them separately another time.

ryanb-gds (approved):
🚀

PR #2455, opened 2024-08-09
Rewrite documentation

ryanb-gds:
The documentation for Manuals Publisher hadn't been touched since 2017. Whilst it wasn't hugely out of date because the code hasn't changed much since then, it was confusing because it contained a lot of reference to the intentions of a specific project that the team responsible had taken on. It also didn't cover the relationship between models and sections and the edition workflow hugely effectively. I have decided to keep this documentation in its own archive directory with a new introduction, because it is useful and explains some of the history of Manuals Publisher quite nicely.

The new documentation includes diagrams of the model and section state machines and discusses how the manual interacts with the state of its sections, and vice-versa. Hopefully this will make our lives a bit easier next time we are seeking to work with manuals.


README.md
```diff
@@ -25,11 +25,16 @@ $ bundle exec rake
...
 
-* [History of the development of the application](docs/history.md)
-* [Current state of the application](docs/current-state.md)
-* [Information about Rake tasks](docs/rake-tasks.md)
-* [Possible next steps for development](docs/next-steps.md)
-* [Notes on investigation into fully migrating app with respect to Publishing API](docs/fully-migrated-spike.md)
+* [Manuals Publisher Data Model](docs/data-model.md)
+* [Manual and Section Edition Workflow](docs/edition-workflow.md)
+
+## 2017 Migration Project Archive
```
minhngocd, line 31: Maybe we can link to the start-here page from here?

minhngocd (approved):
Thanks for this @ryanb-gds !! 🙏  sorry i only just noticed 🙇‍♀️ 

Is it also worth calling out some of the state being out of sync that we've seen where sections in belonging to withdrawn manuals still have "published" state even though they have been unpublished on Publishing API? that is something we still need to correct and fix. 

also is it worth noting that the state of the section has no bearing over whether or not they appear under a manual in the publisher UI? e.g.  'archived' sections can still show up if it belongs to 'section_uuids' and 'published' sections can be removed if it's under 'removed_section_uuids'.

ryanb-gds:
> Is it also worth calling out some of the state being out of sync that we've seen where sections in belonging to withdrawn manuals still have "published" state even though they have been unpublished on Publishing API? that is something we still need to correct and fix.
> 
> also is it worth noting that the state of the section has no bearing over whether or not they appear under a manual in the publisher UI? e.g. 'archived' sections can still show up if it belongs to 'section_uuids' and 'published' sections can be removed if it's under 'removed_section_uuids'.

On the first point, I feel that's just a bug and we have a ticket on our Trello board about it I think, so I don't feel it should be part of the project docs.

Good shout about the UI though, I think that is definitely worth noting

PR #2458, opened 2024-08-14
Section Redirects should be publishing the redirect afterwards

minhngocd:
Testing out the rake task in integration and staging, the redirects were not applying until a publishing is made.

### DIFF
--- a/app/adapters/publishing/redirect_adapter.rb
+++ b/app/adapters/publishing/redirect_adapter.rb
@@ -2,8 +2,9 @@ require "securerandom"
 
 class Publishing::RedirectAdapter
   def self.redirect_section(section, to:)
+    redirect_content_id = SecureRandom.uuid
     Services.publishing_api.put_content(
-      SecureRandom.uuid,
+      redirect_content_id,
       document_type: "redirect",
       schema_name: "redirect",
       publishing_app: GdsApiConstants::PublishingApi::PUBLISHING_APP,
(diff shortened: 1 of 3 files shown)


ryanb-gds (commented):
Thinking about it, isn't that redirect adapter calling the wrong method completely? Why are we calling [put content](https://docs.publishing.service.gov.uk/repos/publishing-api/api.html#put-v2contentcontent_id) when we could be calling [unpublish](https://docs.publishing.service.gov.uk/repos/publishing-api/api.html#post-v2contentcontent_idunpublish)?

I suspect the put content with a redirect and publish probably amounts to about the same thing as unpublishing but it's very odd.

minhngocd:
> Thinking about it, isn't that redirect adapter calling the wrong method completely? Why are we calling [put content](https://docs.publishing.service.gov.uk/repos/publishing-api/api.html#put-v2contentcontent_id) when we could be calling [unpublish](https://docs.publishing.service.gov.uk/repos/publishing-api/api.html#post-v2contentcontent_idunpublish)?
> 
> I suspect the put content with a redirect and publish probably amounts to about the same thing as unpublishing but it's very odd.

I think because "put content redirect" doesn't rely on a content_id (hence the secure random guid - it just redirects a route - any route), and unpublish is unpublishing and redirecting a specific content with a content ID. 

Fair shout about why this is a redirect and not an unpublish - i wonder if it's because it's being used by the section reslugger - which might have differing content IDs? 

A similar task exists on whitehall to redirect a route
```
  desc "Publish redirect routes (eg /government/world)"
  task publish_redirect_routes: :environment do
```

ryanb-gds (approved):
Let's see if this does the trick

PR #2828, opened 2025-09-18
Add a confirmation step to relocate_manual rake task [WHIT-1876]

kashifatcha:
I have added a confirmation step, using the Thor gem, to the above mentioned rake task. This is to ensure that the person running this task explicitly opts in to relocating the manual from_slug to_slug.


TonyGDS (commented):
We're missing a test for this task. This is a good opportunity to add one and also verify the confirmation is working as expected.

lauraghiorghisor-tw (approved):
There is test coverage in the relocator, testing the rake task would just be testing Thor as it's quite simplistic. Think I am happy without tests.

Whilst it would be nice to surface some intricate business logic (like, we actually redirect manual sections as well, which may not be very intuitive), that feels like a whole can of worms.

PR #2830, opened 2025-09-24
Add confirmation step to reslug_organisation rake task [WHIT-1876]

kashifatcha:
I have added a confirmation step, using the Thor gem, to the above rake task. 

This is to ensure that the person running these rake tasks explicitly opts in to making these changes. I have also added a stub in the before :each block in reslug_organisation_spec to prevent the rake task from waiting for actual user input during the test.

### DIFF
--- a/lib/tasks/reslug_organisation.rake
+++ b/lib/tasks/reslug_organisation.rake
@@ -1,3 +1,9 @@
+require "thor"
+
+def shell
+  @shell ||= Thor::Shell::Basic.new
+end
+
 desc "Republish manuals"
 task :reslug_organisation, %i[old_slug new_slug] => :environment do |_, args|
   unless args[:old_slug] && args[:new_slug] && args.count == 2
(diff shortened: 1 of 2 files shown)


lib/tasks/reslug_organisation.rake
```diff
@@ -6,6 +6,10 @@ task :reslug_organisation, %i[old_slug new_slug] => :environment do |_, args|
 
   manual_records = ManualRecord.where(organisation_slug: args[:old_slug]).to_a
   puts "Updating the `organisation_slug` of #{manual_records.count} manual records from '#{args[:old_slug]}' to '#{args[:new_slug]}'"
+  unless Thor::Shell::Basic.new.yes?("Would you like to proceed with this? (yes/no)")
```
TonyGDS, line 9: We can avoid calling the fully qualified class name and also use Thor for printing out to the console. We tried this the other day in Whitehall:
https://github.com/alphagov/whitehall/blob/main/lib/tasks/data_hygiene.rake#L3
https://github.com/alphagov/whitehall/blob/main/lib/tasks/data_hygiene.rake#L113
Maybe we can adopt the same pattern?

Possibly we could include the operation in our confirmation like:
"Would you like to proceed with Updating the `organisation_slug` of X manual records from 'A' to 'B'"

spec/lib/tasks/reslug_organisation_spec.rb
```diff
@@ -8,6 +9,7 @@
   before :each do
     task.reenable # without this, calling `invoke` does nothing after first test
     allow($stdout).to receive(:puts) # supress Rake task output
+    allow_any_instance_of(Thor::Shell::Basic).to receive(:yes?).and_return(true)
```
TonyGDS, line 12: We should add a test to make sure the operation doesn't happen if the user declines, similar to:
https://github.com/alphagov/whitehall/blob/2c27908d47301e4b3d0a8baa2456d88a609856b0/test/unit/lib/tasks/data_hygiene/reassign_role_appointment_speeches.rb#L43

PR #2937, opened 2026-02-09
Add rescue for general KeyError case

KludgeKML:
- There are manuals in ManualsPublisher which appear published, but have no published editions, and therefore crash when trying to find the latest edition for a section during a bulk republish.

(This allows the job to complete, it still doesn't republish all the manuals but there are only three exceptions now).


ryanb-gds (approved):
Seems like a good idea. Sorry it took so long to get a review @KludgeKML

PR #3033, opened 2026-05-04
[WHIT-3381] [WHIT-1985] Bring back rake task for redirecting multiple sections

lauraghiorghisor-tw:
Revert "Remove rake task to withdraw and redirect manual and sections to different paths via a CSV file"

This reverts commit 6bd20c6666f95bb45ed6cac2c6027c4bc2aac5f7.

Since the reversal of commits c0c3c48f and 7efdc28e, we have brought back the `include_section` flag, due to user need. Despite the commit at the time of removal (6bd20c66) suggesting alternatives, we've recently had the need for an efficient bulk approach when dealing with multiple sections that need to be redirected to different resources (such as a recent request for redirects from manuals to mainstream for 'Assured periodic tenancies: a guide for landlords').

The concern at the time of removal was that a manual could be withdrawn without its sections. A confirmation step to the manual withdraw rake task now helps prevent this (commit 1749d288).

The existing rake task uses the csv approach, which can be a bit unwieldy, but it's handy to have it discoverable in the codebase. We did not find this rake task when doing that redirect earlier in May, so we managed with the existing tasks, but it was quite labour-intensive.

Note: I did not test this, it's an old task, but we might have to use it soon, so we can verify it then. It could benefit from some checks like ensuring ALL sections are dealt with, or at least prompting the user about any sections that haven't been included (they remain live with a funny title area since their parent is missing). There was a lot of manual counting and cross-checking when doing this in May with the section rake task.

Jira:
https://gov-uk.atlassian.net/browse/WHIT-3381
https://gov-uk.atlassian.net/browse/WHIT-1985

### DIFF
--- a/app/lib/withdraw_and_redirect_to_multiple_paths.rb
+++ b/app/lib/withdraw_and_redirect_to_multiple_paths.rb
@@ -1,0 +1,85 @@
+require "csv"
+
+class WithdrawAndRedirectToMultiplePaths
+  def initialize(csv_path:, discard_drafts: false, dry_run: false)
+    @csv_path = csv_path
+    @discard_drafts = discard_drafts
+    @user = User.gds_editor
+    @dry_run = dry_run
+  end
+
+  def execute
+    if dry_run
+      missing_paths = { sections: [], manuals: [] }
+    end
+
+    grouped_sections_by_manual.each do |manual_path, children|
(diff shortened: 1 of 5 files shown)


jamiestamp (approved):
LGTM!

